PR #86, opened 2025-04-10
Support for prepared statements over simple protocol

levkk:
### Features
- Support for `PREPARE`, `EXECUTE` and `DEALLOCATE` commands.
- Admin command to show prepared statements.


levkk:
Yes, that's how it works currently here too.

rocinant3:
Maybe i do something wrong?

tokio-postgres
```
    // client 1
    let (client, conn) = config.connect(NoTls).await?;

    let stmpt = client.prepare("SELECT * FROM orders").await?;
    client.query(&stmpt, &[]).await?; // ok

    // client 2
    let (client, conn) = config.connect(NoTls).await?;

    client.query(&stmpt, &[]).await?; // Error:
                                      // code: SqlState(E58000),
                                      // message: "prepared statement \"s0\" is missing from cache"
```

pgdog.toml includes prepared_statements = "full"

levkk:
Let me take a look.

levkk:
I think your code has a bug. Prepared statements are connection specific, i.e. two clients can't share the _same_ prepared statement. You need to prepare it twice, once for each client, and PgDog (and pgbouncer) will automatically dedup those at the pooler. But in your code, they need to be two different statements. This error is coming from the tokio-postgres I think.

rocinant3:
Double requests are something I would like to avoid, thanks anyway

Issue #85, opened 2025-04-09
Look into concurrent HashMaps like DashMap


levkk:
### Description

We use a couple `Arc<Mutex<HashMap<_, _>>>` in the hot path. We should explore faster hashmaps that are built for concurrent use cases.

e.g.:

1. https://github.com/xacrimon/dashmap

levkk:
Using DashMap for now, unless it gets too hot. No issues reported...yet.

PR #112, opened 2025-04-21
Use dashmap in a couple of low hanging places

AdamGS:
IDK if that's what you had in mind in #85. I think another two easy followups are:
1. Create specialized iterator type over dashmap to make sure users of caches can't mutate entries while iterating the content.
2. Switching the hashing function from the default `SipHash` to something faster but less DOS resistant like [foldhash](https://github.com/orlp/foldhash).


pgdog/src/frontend/router/parser/cache.rs
```diff
@@ -55,52 +66,45 @@ impl Cache {
...
 
     /// Reset cache.
     pub fn reset() {
-        let cache = Self::get();
-        let mut guard = cache.inner.lock();
-        guard.queries.clear();
-        guard.queries.shrink_to_fit();
-        guard.stats.hits = 0;
-        guard.stats.misses = 0;
+        // This is the only place we acquire the write lock, in order to synchronize clearing the cache
```
levkk: Why do we need an RwLock at all? Both `stats` and `queries` are synchronized, we can just clear the DashMap and reset the counters to zero. DashMap has its own RwLock, so we're double-locking for each access to the cache.

AdamGS: probably just out of an abundance of caution, making this operation atomic seemed nice but I guess its just fine if the stats are a few cycles behind the map.

levkk:
Thank you for the PR. I'm going to benchmark it to see if there is a significant difference between the two approaches. The query parser gets hit only in multi-shard configuration and when the client is using prepared statements (or the extended protocol). The peers/discovery module isn't used that heavily (yet). I wrote it before realizing most public clouds disable multicast on their network.

pgdog/src/frontend/comms.rs
```diff
@@ -55,21 +55,16 @@ impl Comms {
...
     }
 
     /// Get all connected clients.
-    pub fn clients(&self) -> HashMap<BackendKeyData, ConnectedClient> {
-        self.global.clients.lock().clone()
-    }
-
-    /// Number of connected clients.
-    pub fn clients_len(&self) -> usize {
-        self.global.clients.lock().len()
```
AdamGS, line 72: Seems like this function and `len` served the same purpose?

levkk: Oh, yeah. Good catch.

PR #119, opened 2025-04-24
Build arm64 and amd64 images

cskinfill:
Based on the example Github Actions workflow [here](https://docs.docker.com/build/ci/github-actions/multi-platform/).

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -25,6 +25,10 @@ jobs:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


levkk:
Let's give it a shot. Last time we got stuck in qemu for an hour and timed out, not sure why.

levkk:
https://github.com/pgdogdev/pgdog/actions/runs/14652626155/job/41121733721

levkk:
I'm thinking we need to pre-build the base image with all the dependencies. QEMU is slowing things down a lot I think and it just times out?

denysvitali:
FYI, you can use an `arm64` runner:
https://github.com/orgs/community/discussions/148648

denysvitali:
See #133

Issue #108, opened 2025-04-19
Make healthcheck query configurable


levkk:
### Description

Currently, the health check query is not configurable. We are using `;` to bypass the parser. Allow this to be configurable.

vitabaks:
It would be nice to be able to send checks to the [Paroni REST API](https://patroni.readthedocs.io/en/latest/rest_api.html#health-check-endpoints) to determine the role (primary, replica, sync replica, etc)

PR #127, opened 2025-05-02
Add new option `healthcheck_query` to configure the query during healthchecks

kvch:
This PR adds support for running custom queries during healthchecks. Once the PR is merged, I will put up another PR to add the new option to the documentation.

BTW awesome project! :dog: 

Related https://github.com/pgdogdev/pgdog/issues/108


levkk:
Thanks for the PR. I'm currently in the middle of optimizing that module (as you probably can see from the mild merge conflicts). I'll rebase your PR in a little while, just need to finish a few things. Speaking of which, when you clone the `healthcheck_query`, it happens while the pool lock is acquired. It's surprisingly slow to allocate memory for a new string and this can cause contention. I'm trying to figure out how to make this easier, but a quick fix could be to put that string behind an `Arc`.

kvch:
Thank you for the feedback! I put the query string behind an `Arc`. Let me know if there is anything else I can do to improve the code.

PR #133, opened 2025-05-05
Add ARM64 build

denysvitali:
Follow up of #119. Tested on my local fork and [it works](https://github.com/denysvitali/pgdog/actions/runs/14835375542/job/41645377337).


levkk:
Worked. Thanks!

PR #147, opened 2025-05-10
ci: add linux arm64 build

sandangel:
Use docker buildx for multi platform build

Reference build: https://github.com/sandangel/pgdog/actions/runs/14945641748/job/41990218584

Signed-off-by: San Nguyen <vinhsannguyen91@gmail.com>

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -45,14 +45,6 @@ jobs:
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
 
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v3
-
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3
-        with:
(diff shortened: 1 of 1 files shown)


levkk:
We tried this before a couple of times but the CI keeps breaking (timing out on the ARM build) and when using two separate runners, one would overwrite the tag for the other. I'll give your branch a try again to make sure it works before merging.

sandangel:
Actually my test is failing as well for multi platform build using QEMU. Will have to change to multi nodes build and then merge images https://docs.docker.com/build/ci/github-actions/multi-platform/

sandangel:
image

https://github.com/sandangel/pgdog/actions/runs/14951874217

The build success and now include multiple digests for different platforms.

levkk:
Cool! Let's try it.

denysvitali:
Awesome! Thank you!
I was about to provide a fix for this weird behavior but you were quicker than me!

PR #158, opened 2025-05-12
feat: side-effect SELECT

### LINKED ISSUE #134: `SELECT pg_advisory_lock`
It seems like [`pg_advisory_lock`](https://www.postgresql.org/docs/current/explicit-locking.html) is treated as a SELECT statement, thus the lock never gets acquired.  
  
Would it make sense to have a list of SQL statements that have side effects and that should therefore be whitelisted for being redirected to the write replicas?

denysvitali:
Fixes #134

### DIFF
--- a/pgdog/src/frontend/router/parser/query.rs
+++ b/pgdog/src/frontend/router/parser/query.rs
@@ -31,6 +31,24 @@ use pg_query::{
 use regex::Regex;
 use tracing::{debug, trace};
 
+static SIDE_EFFECT_FUNCTIONS: Lazy<HashSet<String>> = Lazy::new(|| {
+    let mut set = HashSet::new();
+    // Advisory lock functions
+    set.insert("pg_advisory_lock".to_lowercase());
+    set.insert("pg_advisory_xact_lock".to_lowercase());
+    set.insert("pg_advisory_lock_shared".to_lowercase());
+    set.insert("pg_advisory_xact_lock_shared".to_lowercase());
+    set.insert("pg_try_advisory_lock".to_lowercase());
+    set.insert("pg_try_advisory_xact_lock".to_lowercase());
+    set.insert("pg_try_advisory_lock_shared".to_lowercase());
+    set.insert("pg_try_advisory_xact_lock_shared".to_lowercase());
+    // Sequence functions
+    set.insert("nextval".to_lowercase());
+    set.insert("setval".to_lowercase());
+    // Add other functions with side-effects here
+    set
+});
+
 static REPLICATION_REGEX: Lazy<Regex> = Lazy::new(|| {
     Regex::new(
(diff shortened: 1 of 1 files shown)


pgdog/src/frontend/router/parser/query.rs
```diff
@@ -226,6 +244,13 @@ impl QueryParser {
         let mut command = match root.node {
             // SELECT statements.
             Some(NodeEnum::SelectStmt(ref stmt)) => {
+                // Check for side-effect functions in the target list
+                if Self::has_side_effect_function(stmt) {
```
levkk: This is called a bit too early. Right below is where we check the comment for a manually selected shard. You want to support use cases like:

```sql
/* pgdog_shard: 0 */ SELECT nextval('sequence_name');
```

pgdog/src/frontend/router/parser/query.rs
```diff
@@ -226,6 +244,13 @@ impl QueryParser {
...
+                    debug!("Query contains side-effect function, routing to primary.");
+                    return Ok(Command::Query(Route::write(None)));
```
levkk: This makes it a cross-shard query. I don't think we want that, necessarily, especially for advisory locks. You'd probably want to pick a shard that handles it, since advisory locks on different shards won't actually mutually exclude queries.

pgdog/src/frontend/router/parser/query.rs
```diff
@@ -701,6 +726,30 @@ impl QueryParser {
     fn delete(_stmt: &DeleteStmt) -> Result<Command, Error> {
         Ok(Command::Query(Route::write(None)))
     }
+
+    /// Check if a SELECT statement contains any functions with side effects.
+    fn has_side_effect_function(stmt: &SelectStmt) -> bool {
+        for target_item_wrapper_node in &stmt.target_list {
+            if let Some(NodeEnum::ResTarget(res_target)) = target_item_wrapper_node.node.as_ref() {
```
levkk: `ResTarget` is only one possible token here, you can also have `TypeCast` and I think some others, e.g.:

```
SELECT pg_advisory_lock('test')::bool
```

denysvitali: Fair point. Let me handle the other cases as well

denysvitali:
Unfortunately despite `Query contains side-effect function` is correctly printed when Immich (see https://github.com/immich-app/immich/discussions/18236) starts and tries to lock via `SELECT pg_advisory_lock($1)` - the application hangs.  
  
I think this  is due to the fact that the lock can't be acquired - but I wasn't able to find any additional information on the real reason this happens. From a code perspective, I don't see why my approach shouldn't work (tried both by sending to all instances and to just one shard).

Any tip is appreciated - I would really love to use pgdog so that I can load-balance Immich.

levkk:
Cool, I will take a look!

levkk:
Haven't forgotten about this, just trying to find the time. Sorry for the delay.

Issue #157, opened 2025-05-12
Cannot execute INSERT in a read-only transaction


sandangel:
```sh
 sqlalchemy.exc.DBAPIError: (sqlalchemy.dialects.postgresql.asyncpg.Error) <class 'asyncpg.exceptions.ReadOnlySQLTransactionError'>: cannot execute INSERT in a read-only transaction

```

I'm running stress test for my system with pgdog, it has this issue with some requests. Not always though.

```toml
# https://github.com/pgdogdev/pgdog/blob/main/pgdog.toml
[general]
host = "0.0.0.0"
port = 6432
shutdown_timeout = 5_000
openmetrics_port = 9930
query_timeout = 1_000
checkout_timeout = 1_000
connect_timeout = 1_000
idle_timeout = 30_000
prepared_statements = "extended"
passthrough_auth = "enabled_plain"
default_pool_size = 100
workers = 10
min_pool_size = 1
pooler_mode = "transaction"
load_balancing_strategy = "least_active_connections"
auth_type = "trust"

[admin]
user = "pgdog"
password = "pgdog"

[[databases]]
name = "postgres"
host = "localhost"
role = "primary"
port = 5432

[[databases]]
name = "foo"
host = "localhost"
role = "primary"
port = 5432

[[databases]]
name = "foo"
host = "localhost"
port = 5433
role = "replica"

[[databases]]
name = "bar"
host = "localhost"
role = "replica"
port = 5433

[tcp]
retries = 3
time = 1000
interval = 1000
user_timeout = 1000
```

levkk:
Could you enable trace logging:

```bash
export RUST_LOG=trace
```

And share the log around the time when the error happens? This error means pgdog sent a `INSERT` query to a replica. The way the load balancer works is when a transaction is first started, the first query that's executed inside the transaction decides where the entire transaction will go. For example:

```sql
BEGIN; -- No decision yet.
SELECT * FROM users; -- SELECT, the rest of the transaction goes to a replica.
```

Here's how PgDog currently decides where queries go:

| Query | Routing decision | Example |
|-|-|-|
| `SELECT` | replica | `SELECT * FROM users;`
| `SET` | replica | `SET intervalstyle TO 'iso_8601';`
| Everything else | primary |

What I suspect could be happening is SQLALchemy starts a transaction, uses `SET` to set some variables, and then runs an `INSERT` query, which gets routed to a replica because of `SET`. The trace log will tell us more.

[...]

PR #159, opened 2025-05-12
Send SET queries to primary

levkk:
### Description

#157 

When the first query inside a transaction is `SET`, send the transaction to the primary. PgDog needs to execute this query, but we don't have enough information on whether the transaction will read or write. To be safe, send it to the primary.

This PR also makes PgDog handle `SET` queries internally without assigning them to a server. Until a client runs an actual query.


sandangel:
This is affected by `INSERT ... ON CONFLICT DO UPDATE SET ...` query

levkk:
> This is affected by `INSERT ... ON CONFLICT DO UPDATE SET ...` query

Oh! Easy fix, let me add it here.

PR #197, opened 2025-05-29
feat(metrics): add Prometheus-compatible headers to metrics API

pinylin:
Previously the metrics endpoint not included headers compatible with Promtheus.
Now adds 'Content-Type: text/plain' and other headers needed 
for Prometheus scraping while maintaining backward

refer: [# protocol-headers](https://prometheus.io/docs/instrumenting/content_negotiation/#protocol-headers)

### DIFF
--- a/pgdog/src/stats/http_server.rs
+++ b/pgdog/src/stats/http_server.rs
@@ -28,7 +28,7 @@ async fn metrics(_: Request<hyper::body::Incoming>) -> Result<Response<Full<Byte
             "text/plain; version=0.0.4; charset=utf-8",
         )
         .body(Full::new(Bytes::from(metrics_data)))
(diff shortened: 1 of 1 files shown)


levkk:
Hey! Thanks for the PR and apologies for reviewing it late. Just small issue: could you return `Result` instead of using `expect`?  I know the odds of this panicking are 0, but I'm just trying to make sure we don't have anything that can crash PgDog in production unexpectedly. Thanks!

One more administrative thing to discuss as well is a CLA. PgDog is an AGPL-licensed project and also a company. If you'd like to contribute, you are more than welcome to do so. If not, I completely understand!

Thanks!

levkk:
Thank you!

PR #255, opened 2025-07-09
Support List Shards for UUID columns

niclaflamme:
# Changes
- List shards support `uuid` datatype
- Added a sharded_list_uuid table to the dev setup script.
- Added the `serde` feature on the `uuid` crate

# Misc Changes
- Split/grouped the `.../integrations/pgdog.toml` visually with comments.
- Added rollbacks on unrollbacked tests `../frontend/router/sharding/test/mod.rs`

### DIFF
--- a/pgdog/src/config/mod.rs
+++ b/pgdog/src/config/mod.rs
@@ -983,12 +983,19 @@ pub enum ShardedMappingKind {
 #[serde(untagged)]
 pub enum FlexibleType {
     Integer(i64),
+    Uuid(uuid::Uuid),
     String(String),
 }
 
 impl From<i64> for FlexibleType {
     fn from(value: i64) -> Self {
         Self::Integer(value)
+    }
+}
+
+impl From<uuid::Uuid> for FlexibleType {
+    fn from(value: uuid::Uuid) -> Self {
+        Self::Uuid(value)
     }
 }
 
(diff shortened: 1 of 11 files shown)


pgdog/src/config/mod.rs
```diff
@@ -983,6 +983,7 @@ pub enum ShardedMappingKind {
 #[serde(untagged)]
 pub enum FlexibleType {
     Integer(i64),
+    Uuid(uuid::Uuid),
```
niclaflamme, line 986: I didn't import `Uuid` to avoid confusion. Should I?

pgdog/src/frontend/router/sharding/value.rs
```diff
@@ -157,3 +171,23 @@ impl<'a> Value<'a> {
...
+
+        let value = Value {
+            data_type: DataType::Uuid,
+            data: Data::Binary(&raw_bytes),
+        };
+
+        assert_eq!(value.uuid()?, Some(expected_uuid));
+        Ok(())
+    }
+}
```
niclaflamme, line 193: is this testing anything or is this test theatre?

pgdog/src/config/mod.rs
levkk: LGTM.

pgdog/src/frontend/router/sharding/value.rs
levkk: Can't hurt! You can also fetch it from Postgres and decode it (there is a way to force binary encoding), but that may take a few more lines of code.

levkk (approved):
image

PR #263, opened 2025-07-12
DNS Cache

niclaflamme:


pgdog/src/backend/pool/dns_cache.rs
```diff
@@ -0,0 +1,388 @@
+use futures::future::join_all;
+use once_cell::sync::Lazy;
+use std::collections::{HashMap, HashSet};
+use std::net::IpAddr;
+use std::sync::{Arc, RwLock};
```
levkk: Use `parking_lot::RwLock`, it doesn't panic.

pgdog/src/backend/pool/dns_cache.rs
```diff
@@ -0,0 +1,388 @@
...
+// ------ Constant(s) ------------------------------------------------------------------------------
+
+const DEFAULT_TTL: Duration = Duration::from_secs(30);
+
+// -------------------------------------------------------------------------------------------------
+// ------ Singleton(s) -----------------------------------------------------------------------------
+
+static DNS_CACHE: Lazy<Arc<DnsCache>> = Lazy::new(|| {
+    Arc::new(DnsCache {
+        cache: Arc::new(RwLock::new(HashMap::new())),
```
levkk: If you'd  like, you could move this initialization logic into `DnsCache::new`.

pgdog/src/backend/pool/dns_cache.rs
```diff
@@ -0,0 +1,388 @@
...
+    pub async fn resolve(&self, hostname: &str) -> Result<IpAddr, std::io::Error> {
+        if let Some(ip) = self.get_cached_ip(hostname) {
+            return Ok(ip);
+        }
+
+        // Track hostname for future refreshes.
+        {
+            let mut hostnames = self.hostnames.write().unwrap();
+            if hostnames.insert(hostname.to_string()) {
+                info!("DNS Cache :: Added hostname: {}", hostname);
```
levkk: This feels like a `debug!` level log. Maybe something like this to keep existing format:

```rust
debug!("dns cache added hostname \"{}\"",  hostname);
```

pgdog/src/backend/pool/dns_cache.rs
```diff
@@ -0,0 +1,388 @@
...
+        let cache_ref = Arc::clone(self);
+        tokio::spawn(async move {
+            loop {
+                let interval = cache_ref.ttl();
+
+                select! {
+                    _ = sleep(interval) => {
+                        cache_ref.refresh_all_hostnames().await;
+                    }
+                    _ = cancel_clone.cancelled() => {
```
levkk: This feels like one of those infinite loops that don't need to  be cancelled. It'll just stop the the pooler is restarted.

pgdog/src/backend/pool/dns_cache.rs
```diff
@@ -0,0 +1,388 @@
+use futures::future::join_all;
+use once_cell::sync::Lazy;
+use std::collections::{HashMap, HashSet};
+use std::net::IpAddr;
+use std::sync::{Arc, RwLock};
+use std::time::{Duration, Instant};
+use tokio::net::lookup_host;
```
levkk: I don't think this one will bypass TTL set on the record. This uses the OS's DNS  resolution mechanism. You need to bypass that and talk directly to the DNS  servers. Check out this crate: https://docs.rs/trust-dns-resolver/0.23.2/trust_dns_resolver/


Edit: Apparently they moved to https://docs.rs/hickory-resolver/latest/hickory_resolver/

pgdog/src/backend/pool/address.rs
```diff
@@ -48,9 +50,9 @@ impl Address {
         }
     }
 
-    /// Get address for `TCPStream`.
-    pub fn addr(&self) -> String {
-        format!("{}:{}", self.host, self.port)
+    pub async fn addr(&self) -> Result<SocketAddr, std::io::Error> {
```
levkk: Check out the `backend::Error` enum. We use `thiserror` for returning errors most places, it might have an `std::io::Error` mapping already and you can continue to use  `?`.

pgdog/src/backend/pool/dns_cache.rs
```diff
@@ -0,0 +1,388 @@
...
+const DEFAULT_TTL: Duration = Duration::from_secs(30);
+
+// -------------------------------------------------------------------------------------------------
+// ------ Singleton(s) -----------------------------------------------------------------------------
+
+static DNS_CACHE: Lazy<Arc<DnsCache>> = Lazy::new(|| {
+    Arc::new(DnsCache {
+        cache: Arc::new(RwLock::new(HashMap::new())),
+        hostnames: Arc::new(RwLock::new(HashSet::new())),
+        ttl: Arc::new(RwLock::new(DEFAULT_TTL)),
```
levkk: You can fetch this from the config and it will always reflect the latest version of it:

```rust
crate::config::config().config.general.dns_ttl
```

pgdog/src/config/mod.rs
```diff
@@ -396,6 +396,9 @@ pub struct General {
     /// Disable cross-shard queries.
     #[serde(default)]
     pub cross_shard_disabled: bool,
+    /// How often to refresh DNS entries, in ms.
+    #[serde(default = "General::default_dns_ttl")]
+    pub dns_ttl: u64,
```
levkk: Make it optional. If not set, use OS DNS resolution, otherwise use our DNS cache.

pgdog/src/backend/pool/address.rs
```diff
@@ -48,9 +50,9 @@ impl Address {
...
+        let ip = DnsCache::global().resolve(&self.host).await?;
```
levkk: For this one, you might want to do something like this:

1. Check if DNS cache is enabled (setting is not `None`)
2. If `None` (disabled), just return host + port like before, letting `TcpStream` resolve it like before using the OS resolver
3. If enabled, use the `DnsCache`

pgdog/src/backend/pool/dns_cache.rs
```diff
@@ -0,0 +1,388 @@
...
+use tokio::net::lookup_host;
+use tokio::select;
+use tokio::time::sleep;
+use tokio_util::sync::CancellationToken;
+use tracing::{debug, error, info, warn};
+
+// -------------------------------------------------------------------------------------------------
+// ------ Constant(s) ------------------------------------------------------------------------------
+
+const DEFAULT_TTL: Duration = Duration::from_secs(30);
```
levkk: Default behavior should be disabled.

pgdog/src/backend/pool/dns_cache.rs
```diff
@@ -0,0 +1,388 @@
...
+    pub fn set_ttl(&self, ttl: Duration) {
+        let mut current_ttl = self.ttl.write().unwrap();
+        *current_ttl = ttl;
+    }
+
+    /// Start the background refresh loop.
+    /// Returns a CancellationToken that can be used to stop the loop (needed for testing).
+    /// In production, you can ignore the return value with `_`.
+    pub fn start_refresh_loop(self: &Arc<Self>) -> tokio_util::sync::CancellationToken {
+        let cancel_token = CancellationToken::new();
```
levkk: You're returning this but never using it, I don't think it's doing anything.

niclaflamme: it was used in tests to kill the loop, but instead i'll just let the test process die instead of adding complexity.

pgdog/src/backend/pool/dns_cache.rs
niclaflamme: https://docs.rs/hickory-resolver/latest/hickory_resolver/
project migrated to hickory-resolver for in-process dns but ✅

pgdog/src/backend/pool/dns_cache.rs
niclaflamme: it was just for test, but it's a better idea to not branch the logic and just let the test kill the loops when it ends.

pgdog/src/backend/pool/dns_cache.rs
```diff
@@ -0,0 +1,362 @@
...
+    /// Start the background refresh loop.
+    /// This spawns an infinite loop that runs until the process exits.
+    /// For testing, use short TTLs and allow the test runtime to handle shutdown.
+    pub fn start_refresh_loop(self: &Arc<Self>) {
+        let interval = Self::ttl();
+
+        let cache_ref = Arc::clone(self);
+        tokio::spawn(async move {
+            loop {
+                println!("-> I AM REFRESHING!!! {:?}", interval);
```
levkk: Don't  forget to remove all the println statements before  merging.

pgdog/src/backend/pool/dns_cache.rs
```diff
@@ -0,0 +1,362 @@
...
+
+        let ip = self.resolve_and_cache(hostname).await?;
+        Ok(ip)
+    }
+
+    /// Start the background refresh loop.
+    /// This spawns an infinite loop that runs until the process exits.
+    /// For testing, use short TTLs and allow the test runtime to handle shutdown.
+    pub fn start_refresh_loop(self: &Arc<Self>) {
+        let interval = Self::ttl();
```
levkk: Fetch this inside the loop. Otherwise, the refresh loop interval will not be changeable at runtime. We want this setting to be dynamically changeable, e,g.:

1. Change the config
2. Run `RELOAD` in the admin DB (or send `SIGHUP` to pgdog to reload config)

pgdog/src/backend/pool/dns_cache.rs
```diff
@@ -0,0 +1,362 @@
...
+
+        let cache_ref = Arc::clone(self);
```
levkk: We usually just call the `clone` method on the Arc:

```rust
let cache = self.clone();
```

pgdog/src/backend/pool/dns_cache.rs
```diff
@@ -0,0 +1,362 @@
...
+            .into_iter()
+            .map(|hostname| {
+                let cache_ref = Arc::clone(self);
+                tokio::spawn(async move {
+                    if let Err(e) = cache_ref.resolve_and_cache(&hostname).await {
+                        error!("failed to refresh DNS for {}: {}", hostname, e);
+                        println!("failed to refresh DNS for {}: {}", hostname, e);
+                    } else {
+                        println!("refreshed hostname: {}", hostname);
+                        debug!("refreshed hostname: {}", hostname);
```
levkk:
```suggestion
                        debug!("refreshed hostname \"{}\"", hostname);
```

To be consistent with other debug log.

pgdog/src/backend/pool/dns_cache.rs
```diff
@@ -0,0 +1,362 @@
...
+        let response = self
+            .resolver
+            .lookup_ip(hostname)
+            .await
+            .map_err(|e: ResolveError| Error::DnsLookupError(format!("{}: {}", hostname, e)))?;
+
+        let ip = response
+            .iter()
+            .next()
+            .ok_or_else(|| Error::DnsLookupError(format!("{}: no IPs found", hostname)))?;
```
levkk: The error enum expects the hostname here, not an error message. It would look pretty weird in the logs.

pgdog/src/config/mod.rs
```diff
@@ -396,6 +396,9 @@ pub struct General {
     /// Disable cross-shard queries.
     #[serde(default)]
     pub cross_shard_disabled: bool,
+    /// How often to refresh DNS entries, in ms.
+    #[serde(default)]
+    pub dns_ttl: Option<u64>,
```
levkk, line 401: This isn't used right?

levkk (approved):
LGTM

levkk (approved):

PR #264, opened 2025-07-12
Isolate SELECT Handling in QueryParser

niclaflamme:
### TLDR;
- Refactored `SELECT` query logic from split between `query() ->> match ast.node {}` branch and `select()` into **ONLY**  having the logic inside `select()`.
- Propagated to `explain()` for recursive calls.

### Key Changes:
- Moved `SELECT` routing, write detection, sharding from `query()` match block to `select()`.
- Updated `select()` params: `ast`, `cluster`, `shard`, `bind`, `sharding_schema`.
- Propagated changes to `explain()` function signature (added `ast`, `cluster`, `shard`) so it can call the new `select()` for inner `SELECTs`.
- Removed the logic duplication that was inside `EXPLAIN`

### Files:
- src/frontend/router/parser/query.rs:

### Why?
Avoids duplicating `SELECT` logic in `explain()`. `explain()` recursively invokes `select()` on its inner AST query; without centralizing in `select()`, `explain()` couldn't access the full logic.

### Testing/Impact:
- Passes existing tests.
- Did not add test because the goal was to match current behavior.

### DIFF
--- a/pgdog/src/frontend/router/parser/query.rs
+++ b/pgdog/src/frontend/router/parser/query.rs
@@ -302,7 +302,7 @@ impl QueryParser {
 
             // SELECT statements.
             Some(NodeEnum::SelectStmt(ref stmt)) => {
-                self.select(stmt, &ast, cluster, &mut shard, bind, &sharding_schema)
+                self.select(stmt, &ast, cluster, &shard, bind, &sharding_schema)
             }
             // COPY statements.
             Some(NodeEnum::CopyStmt(ref stmt)) => Self::copy(stmt, cluster),
@@ -342,7 +342,7 @@ impl QueryParser {
             }
 
             Some(NodeEnum::ExplainStmt(ref stmt)) => {
-                self.explain(stmt, &ast, cluster, &mut shard, bind, &sharding_schema)
+                self.explain(stmt, &ast, cluster, &shard, bind, &sharding_schema)
             }
 
             // All others are not handled.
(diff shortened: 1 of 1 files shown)


pgdog/src/frontend/router/parser/query.rs
```diff
@@ -603,17 +563,46 @@ impl QueryParser {
 
 impl QueryParser {
     fn select(
+        &mut self,
         stmt: &SelectStmt,
+        ast: &Arc<pg_query::ParseResult>,
+        cluster: &Cluster,
+        shard: &mut Shard,
```
levkk: Does it have to be mutable?

levkk:
Looks good, I would just remove the mutable reference there. To a reader this means you're going to mutate it, but when you don't, it gets confusing quick.

pgdog/src/frontend/router/parser/query.rs
niclaflamme: i'll repull the branch, but i believe this counts as a self.mutate:
```
  // Write overwrite because of conservative read/write split.
        if let Some(true) = self.write_override {
            writes.writes = true;
        }
```

I could remove this part of the logic from the select and run it higher upstream in the function (before the match statement)

niclaflamme: @levkk 

but yeah, if the `writes.writes` happens only in select, the select mutates self.

levkk: I was referring to the mutable reference  on `Shard`.

Issue #249, opened 2025-07-04
Server TLS Support


mscrivo:
Support the ability to connect to the physical shards over TLS. ie similar to pgBouncer's server_tls_* settings [here](https://www.pgbouncer.org/config.html#tls-settings).

I was not able to find anything in the documentation or the code about this, only client TLS settings. Unless I completely missed it?

I'm not sure anyone is going to be able to use this with AWS RDS in production without this. The `force_ssl` param defaults to true since postgres 15 and I doubt anyone is going to willing turn it off, as it will immediately fail any security audit.

levkk:
TLS for server connections is enabled by default. The server (Postgres) is in charge of this setting and will request PgDog connect with TLS (or not).

mscrivo:
> TLS for server connections is enabled by default. We don't have a setting to _force_ it (or disable connecting), so we can add that option, although TLS is enabled by default in RDS too, so it should just work as you'd expect.

But where does it get the CA cert bundle from? With pgbouncer we have to provide the CA cert bundle from: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html for it to work

levkk:
Sounds like you're using `verify-full`[^1], i.e. making the pooler verify the server's cert before connecting. I haven't built that yet, PgDog currently accepts any cert for TLS negotiation. If you deploy PgDog today, it'll connect just fine, but you'll need `verify-full` to pass your security audit I suspect.

We can add that feature to the roadmap as well.

[^1]: https://www.postgresql.org/docs/current/libpq-ssl.html#LIBQ-SSL-CERTIFICATES

jaggederest:
@mscrivo Added an option on the main branch just now, would very much appreciate you trying it out and seeing what you think. TL;DR Two new config options, `tls_verify` for one of (disable, prefer, verify_ca, verify_full), and `tls_server_ca_certificate` which takes a path to your cert bundle. We gave it a test on RDS but it would be great to have independent confirmation.

mscrivo:
@jaggederest thank you very much! Admittedly, I had to put this poc aside for now because the replication feature from an existing database into shards is not yet ready. Once that is ready, I will definitely give this a go again.

levkk:
[...]

PR #269, opened 2025-07-14
Add configuration value for tls verification and optional cert path

jaggederest:
#249

### DIFF
--- a/pgdog/src/backend/server.rs
+++ b/pgdog/src/backend/server.rs
@@ -88,6 +88,11 @@ impl Server {
 
         // Only attempt TLS if not in None mode
         if tls_mode != TlsVerifyMode::None {
+            info!(
+                "Requesting TLS connection to {} with verify mode: {:?}",
+                addr.host, tls_mode
+            );
+
             // Request TLS.
             stream.write_all(&Startup::tls().to_bytes()?).await?;
             stream.flush().await?;
@@ -97,6 +102,11 @@ impl Server {
             let ssl = SslReply::from_bytes(ssl.freeze())?;
 
             if ssl == SslReply::Yes {
+                info!(
+                    "Server {} supports TLS, initiating TLS handshake",
+                    addr.host
+                );
+
                 let connector = connector_with_verify_mode(
                     tls_mode,
                     cfg.config.general.tls_server_ca_certificate.as_ref(),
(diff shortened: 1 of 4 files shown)


pgdog/src/net/tls.rs
```diff
@@ -152,3 +139,221 @@ impl ServerCertVerifier for CertificateVerifyer {
...
+    fn verify_server_cert(
+        &self,
+        end_entity: &CertificateDer<'_>,
+        intermediates: &[CertificateDer<'_>],
+        _server_name: &rustls::pki_types::ServerName<'_>,
+        ocsp_response: &[u8],
+        now: rustls::pki_types::UnixTime,
+    ) -> Result<rustls::client::danger::ServerCertVerified, rustls::Error> {
+        // Use a dummy server name for verification - we only care about cert validity
+        let dummy_name = rustls::pki_types::ServerName::try_from("example.com").unwrap();
```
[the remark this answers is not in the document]
jaggederest: Seems valid, will change if that'd be better.

pgdog/src/backend/server.rs
```diff
@@ -83,24 +83,71 @@ impl Server {
...
+        let cfg = config();
+        let tls_mode = cfg.config.general.tls_verify;
 
-            let server_name = ServerName::try_from(addr.host.clone())?;
-
-            let cipher =
-                tokio_rustls::TlsStream::Client(connector.connect(server_name, plain).await?);
+        // Only attempt TLS if not in None mode
+        if tls_mode != TlsVerifyMode::None {
+            info!(
```
levkk: Most of these should probably go to `debug!` or be removed .

pgdog/src/backend/server.rs
```diff
@@ -83,24 +83,71 @@ impl Server {
...
+            stream.write_all(&Startup::tls().to_bytes()?).await?;
+            stream.flush().await?;
+
+            let mut ssl = BytesMut::new();
+            ssl.put_u8(stream.read_u8().await?);
+            let ssl = SslReply::from_bytes(ssl.freeze())?;
+
+            if ssl == SslReply::Yes {
+                info!(
+                    "Server {} supports TLS, initiating TLS handshake",
```
levkk: Let's keep the logs consistent, I think they mostly start with lowercase letters.

pgdog/src/config/mod.rs
```diff
@@ -667,6 +678,16 @@ pub enum LoadBalancingStrategy {
     LeastActiveConnections,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, Default, PartialEq, Copy)]
+#[serde(rename_all = "snake_case")]
+pub enum TlsVerifyMode {
```
levkk, line 683: We can probably name these the same as Postgres:
- `disabled`
- `prefer`
- `verify-ca`
- `verify-full`

As to avoid confusion. The default should be `prefer`, imo.

levkk (commented):
Looks good to me, a couple cosmetic changes only.

Issue #215, opened 2025-06-15
Does pgdog support replication lag aware routing


TheSuperAgent:
If I have 3 replicas and any one of them is lagging, then it should stop sending traffic, but once the replica is recovered, then it has to send the traffic back to that replica. 

I guess pgpool supports this feature, how about this feature on pgdog?

levkk:
We should definitely add this. I'll put it on the roadmap.

levkk:
Done in #788

PR #272, opened 2025-07-16
Time-based Shard Lag Monitoring

niclaflamme:
This PR adds async replica lag monitoring.

Each `Shard` now tracks its primary's WAL positions and estimates replica lag.
If a replica falls behind the threshold from `pgdog.toml`, it logs an error (no bans at the moment).

Partial fix for #215 -> needs to actually ban instead of logging, will do that soon.

---

### Why

* Routing to stale replicas breaks user expectations.
* We already had WAL LSN data—this wires it up.
* Monitoring belongs inside pgDog, not in external tools.

---

### What’s new

* **Config**: Adds `[replica_lag]` block with `check_interval` and `max_age` in ms.
* **Parsing**: Custom deserializer requires both fields or none.
* **Pool**: Adds `wal_flush_lsn()` and `wal_replay_lsn()` calls.
* **ShardMonitor**: Runs periodic tasks to track primary and estimate replica lag.
* **Error handling**: New error types for WAL query failures.
* **Cluster**: Routing helpers delegate to those.
* **Example config**: Updated `pgdog.toml` with sane defaults.
* **Admin**: Admin `SHOW POOLS;` display replica lag.

---

### How it works

* Primary WAL positions recorded at `check_interval / 2` (min 250ms).
* Replica lag estimated by bytes-behind converted to time.
* If lag > `max_age`, logs error. Otherwise logs debug.

---

### Compatibility

* Off by default unless `[replica_lag]` is set.
* Falls back to 1000ms interval / 25ms max lag if fields missing.
* No API breaks.

### Screenshots
image

### DIFF
--- a/pgdog/src/backend/pool/inner.rs
+++ b/pgdog/src/backend/pool/inner.rs
@@ -451,7 +451,12 @@ impl ReplicaLag {
                     return format!("{}s", total_secs);
                 }
 
-                format!("{}ms", d.as_millis())
+                let millis = d.as_millis();
+                if millis > 0 {
+                    return format!("{}ms", millis);
+                }
+
+                return "<1ms".to_string();
             }
             Self::Bytes(b) => format!("{}B", b),
             Self::Unknown => "unknown".to_string(),
--- a/pgdog/src/backend/pool/shard.rs
+++ b/pgdog/src/backend/pool/shard.rs
@@ -254,6 +254,8 @@ impl ShardMonitor {
             return;
         };
 
+        primary.set_replica_lag(ReplicaLag::NonApplicable);
+
         let lsn_metrics = match collect_lsn_metrics(primary, max_age).await {
             Some(m) => m,
             None => {
@@ -287,7 +289,7 @@ impl ShardMonitor {
         replica: &Pool,
         primary_lsn: u64,
         lsn_throughput: f64,
-        max_age: Duration,
+        _max_age: Duration, // used to make banning decisions when it's supported later
     ) {
         if replica.banned() {
             replica.set_replica_lag(ReplicaLag::Unknown);
@@ -309,23 +311,8 @@ impl ShardMonitor {
             let duration = Duration::from_secs_f64(bytes_behind as f64 / lsn_throughput);
             lag = ReplicaLag::Duration(duration);
         }
-
-        match lag {
-            ReplicaLag::Duration(d) if d > max_age => {
-                error!(
-                    "replica {} estimated lag {:?} exceeds {:?}; should ban replica",
-                    replica.id(),
-                    d,
-                    max_age
-                );
-            }
-            ReplicaLag::Duration(d) => {
-                debug!("replica {} lag {:?} (estimated)", replica.id(), d);
-            }
-            ReplicaLag::Bytes(b) => {
-                debug!("replica {} lag {} bytes", replica.id(), b);
-            }
-            _ => {}
+        if bytes_behind == 0 {
+            lag = ReplicaLag::Duration(Duration::ZERO);
         }
 
         replica.set_replica_lag(lag);


pgdog/src/backend/pool/shard.rs
```diff
@@ -209,6 +227,216 @@ impl ShardInner {
 // -------------------------------------------------------------------------------------------------
 // ----- Monitoring --------------------------------------------------------------------------------
 
+struct ShardMonitor {}
+
+impl ShardMonitor {
+    pub fn run(shard: &Shard) {
+        let history: Arc<RwLock<RecentLsnHistory>> = Arc::default();
+
+        println!("WHAT THE FUCK...");
```
levkk: LOL!

pgdog/src/backend/pool/shard.rs
```diff
@@ -52,96 +161,352 @@ impl Shard {
...
 
-    /// Create new identical connection pool.
-    pub fn duplicate(&self) -> Self {
+    fn duplicate(&self) -> Self {
         Self {
-            primary: self.primary.as_ref().map(|primary| primary.duplicate()),
+            primary: self.primary.as_ref().map(|p| p.duplicate()),
             replicas: self.replicas.duplicate(),
             rw_split: self.rw_split,
+            comms: ShardComms::default(),
```
niclaflamme: question for duplicate.
do we want to have a separate set of comms or the same one?

pgdog/src/backend/pool/shard.rs
```diff
@@ -131,17 +233,287 @@ impl Shard {
...
+// ----- ShardMonitoring :: LSN History ------------------------------------------------------------
+
+#[derive(Clone, Copy)]
+struct LsnEntry {
+    lsn: u64,
+    captured_at: Instant,
+}
+
+#[derive(Clone)]
+struct RecentLsnHistory {
```
levkk: You can use `VecDeque`. It'll allocate the capacity once with a single heap allocation and won't reduce it even if you `pop_front()` and then `push_back()`, e.g.:

```rust
let mut queue = VecDeque::with_capacity(25); // 1 allocation
queue.push_back(5); // No allocation.
queue.pop_front(); // No de-allocation.
queue.push_back(6); // No allocation.
// etc.
```

pgdog/src/backend/pool/shard.rs
```diff
@@ -1,49 +1,158 @@
...
         lb_strategy: LoadBalancingStrategy,
         rw_split: ReadWriteSplit,
+    ) -> Self {
+        Self {
+            inner: Arc::new(ShardInner::new(primary, replicas, lb_strategy, rw_split)),
+        }
+    }
+
+    /// Get a connection to the shard primary database.
+    pub async fn primary(&self, request: &Request) -> Result<Guard, Error> {
```
levkk, line 36: Consider implementing `std::ops::Deref`. Less boilerplate, everything is public anyway.

pgdog/src/backend/pool/shard.rs
levkk: Yeah that's fine. Good isolation of concerns.

pgdog/src/backend/pool/shard.rs
```diff
@@ -194,19 +85,44 @@ impl ShardInner {
...
+                comms: ShardComms::default(), // Create new comms instead of duplicating
+            }),
         }
     }
 
-    async fn cancel(&self, id: &BackendKeyData) -> Result<(), super::super::Error> {
+    /// Bring every pool online.
+    pub fn launch(&self) {
+        ShardMonitor::run(self);
+        self.pools().iter().for_each(|pool| pool.launch());
```
levkk, line 105: Since you're using pools inside the monitor, might want to launch them first. It's technically a race that will only happen on really fast hardware I believe.

pgdog/src/backend/pool/pool_impl.rs
```diff
@@ -416,4 +417,49 @@ impl Pool {
...
+
+        parse_pg_lsn(&lsn).ok_or(Error::ReplicaLsnQueryFailed)
+    }
+
+    pub fn set_replica_lag(&self, replica_lag: ReplicaLag) {
+        self.lock().replica_lag = replica_lag;
+    }
+}
+
+fn parse_pg_lsn(s: &str) -> Option<u64> {
```
levkk: This is cool. Would be awesome to add a reference to the spec.

pgdog/src/backend/pool/inner.rs
```diff
@@ -417,6 +419,71 @@ pub(super) struct CheckInResult {
...
+#[derive(Clone, Copy, Debug)]
+pub enum ReplicaLag {
+    NonApplicable,
+    Duration(std::time::Duration),
+    Bytes(u64),
+    Unknown,
+}
+
+impl ReplicaLag {
+    pub fn simple_display(&self) -> String {
```
levkk, line 434: It's cool to keep this for now but keep in mind that stats will be read by some kind of monitoring agent and it'll expect to have a consistent unit.

Since our lowest denomination is millisecond, we should probably just display it as a number and document that it's milliseconds.

levkk (approved):
image

PR #282, opened 2025-07-22
[WIP] Handle global sharding keys with mappings

niclaflamme:


pgdog/src/frontend/router/sharding/context_builder.rs
```diff
@@ -38,8 +41,53 @@ impl<'a> ContextBuilder<'a> {
         }
     }
 
-    /// Guess the data type.
-    pub fn from_str(value: &'a str) -> Result<Self, Error> {
+    pub fn shards(mut self, shards: usize) -> Self {
```
niclaflamme, line 44: this actually unchanged and unmoved.

`from_str` was moved in its own `impl` with its private helper method.

pgdog/src/backend/pool/cluster.rs
```diff
@@ -60,8 +63,26 @@ impl ShardingSchema {
     pub fn tables(&self) -> &ShardedTables {
         &self.tables
     }
+
+    pub fn has_global_sharding_key_with_mappings(&self) -> bool {
+        let mut global_sharding_rules_with_mappings = 0;
```
levkk: Checkout the `map`, `filter`, `count`, and other iterator functions. Looks prettier!

```rust
let count = self.tables.tables()
  .iter()
  .filter(|table| 
      table.name.is_none() && table.mapping.is_some()
).count();

count == self.tables.tables.len()
```

levkk: Why `name.is_none()` check?

pgdog/src/backend/pool/cluster.rs
```diff
@@ -60,8 +63,26 @@ impl ShardingSchema {
...
+            } else {
+                other_rules += 1;
+            }
+        }
+
+        global_sharding_rules_with_mappings == 1 && other_rules == 0
+    }
 }
 
+// -------------------------------------------------------------------------------------------------
```
levkk, line 87: Checkout the Rust style guide for comments: https://doc.rust-lang.org/book/ch03-04-comments.html

Could be cool to use it so we can generate documentation automatically using `rust doc`.

pgdog/src/backend/pool/cluster.rs
niclaflamme: i think we talked about this sync, so i'm mostly mentionnig it here:
1- we need a single global rule, until we support naming shardng keys in JSON. (comment 2).

2- i can use the iter methods, but i need 2 values:
the count of rules, the count global_mapping_rules.

```rust
// I rather not reduce with a tuple accumulator as it's less readable than a mutable local variable.
pub fn has_global_sharding_key_with_mappings(&self) -> bool {
    let (matching, total) = self.tables.tables()
        .iter()
        .fold((0, 0), |(matching, total), table| {
            let matching = if table.name.is_none() && table.mapping.is_some() {
                matching + 1
            } else {
                matching
            };
            (matching, total + 1)
        });
    
    matching == 1 && total == 1
}
```

```rust
// Could tehcnically write it this way, but it's the same level of readble.
pub fn has_global_sharding_key_with_mappings(&self) -> bool {
    let tables = self.tables.tables().collect::<Vec<_>>();
    let matching_tables = tables
        .iter()
        .filter(|table| table.name.is_none() && table.mapping.is_some())
        .count();
    let total_tables = tables.len();
    
    matching_tables == 1 && total_tables == 1
}
```

if this was a simple : `does this value exist` i'd write it as `.iter().filter().count() == 1`

levkk:
Consider these examples:

levkk:
Consider these examples:

### 2 tables, referring to the same column, same algorithm

```
[[sharded_tables]]
database = "prod"
name = "users"
column = "id"
data_type = "bigint"

[[sharded_tables]]
database = "prod"
column = "user_id"
data_type = "bigint"
```

### 3 tables, same algorithm

```toml
[[sharded_tables]]
database = "prod"
name = "sensor_readings"
column = "id"
data_type = "bigint"

[[sharded_tables]]
database = "prod"
name = "sensors"
column = "id"
data_type = "bigint"

[[sharded_mappings]]
database = "prod"
name = "sensor_readings"
column = "id"
values = [1, 2, 3]
shard = 0

[[sharded_mappings]]
database = "prod"
name = "sensors"
column = "id"
values = [1, 2, 3]
shard = 0
```

Both of these would indicate we're using the same sharding function and would benefit from `SET pgdog.sharding_key` working. I think the check should be: are all tables in the config using the same sharding function? If so, enable `SET pgdog.sharding_key`, else disable it and return an error.

niclaflamme:
```
[[sharded_tables]]
database = "prod"
name = "users"
column = "id"
data_type = "bigint"

[[sharded_tables]]
database = "prod"
column = "user_id"
data_type = "bigint"
```

This example doesn't really guarantee out "this is the same column".
As a human, i read this and sure, they look like go to the same place, but what if the column was `book_id`?
In any case, `id != user_id`. In this case, your schema has two sharding_keys.

The way i understand the above schema is:
- if the table is users and the id column is found -> shard(hash(user.id));
- if the query has a `user_id` -> shard(hash(user_id));

```sql
BEGIN;
SET pgdog.sharding_key = 12345;
-- what shard am i on right now? user_id, or user.id ?
```

# Tuple uniqueness?

Unless the algo: "The tuple `(hashing_function, datatype, lists, ranges)` is unique across every rule."
It's confusing because it requires `(hashing_function, datatype, lists, ranges)` to match, but not `(column)`.


# With `table:column:key` mode
```sql
BEGIN;
SET pgdog.sharding_key = { table: 'users', column: 'id', key: 12345 };
--  i know exactly what shard this refers to.
```

In other words, all the unintuitive question marks go away as soon as you can give very precise sharding keys.
I can scan the schema and there is zero ambiguity.

# It's a convenience for when there are no edge cases.
I really see this as a featurette for when users have a very specific setup.
- global setting
- every table

```sql
BEGIN;
SET pgdog.sharding_key = 12345;
--  i know exactly what shard this refers to.
```


# The idiot-proof API
If you give me a blank slate and ask me to design an API that's intuitive, i'm making it very strict.
Ideally there should be 3 very specific SET types.

```sql
SET pgdog.global_sharding_key = 12345;

SET pgdog.sharding_key = { table: 'users', column: 'id', key: 12345 };
SET pgdog.sharding_key = { table: '*'    , column: 'id', key: 12345 };     -- table = None
SET pgdog.sharding_key = {                 column: 'id', key: 12345 };     -- table = None

SET pgdog.shard = 0;
```

Zero confusion.
You can equality check the `pgdog.toml` config to immediately know if it's valid.

Currently, the only time I'd feel safe with a simple sharding key is if i have a global schema.
Which makes sense if you want a `kv` store for an app that doesn't `JOIN` anything ever.

Assume: `sharding_key = { table: 'users', column: 'id', key: 12345 }` is available to me... 
Why would I ever use `sharding_key = 12345` and hope for the best.

```sql

-- my insert should go to the same shard as user 12345... 🤷 ? **hope and prayers activated**
BEGIN;
SET sharding_key = 12345;
INSERT INTO addresses(street, state) VALUES ($1, $2);
COMMIT;

-- my insert will 100% go to the same shard as user 12345.
BEGIN;
SET sharding_key = { table: 'users', column: 'id', key: 12345 };
INSERT INTO addresses(street, state) VALUES ($1, $2);
COMMIT;

```

levkk:
> This example doesn't really guarantee out "this is the same column".

It doesn't have to be. It just needs to land on the same shard, which will happen because:

1. It's the same data  type
2. It's configured on the same database, so the number of shards is the same

PR #283, opened 2025-07-22
[chore] Renamed {configs}.toml to {configs}.example.toml

niclaflamme:
# Changes
- Frees pgdog.toml and users.toml to be played with in dev without messing up with source control.
- Sometimes I don't want to use /integrations and want a simpler example.


levkk (approved):
Yes!

At some point we should add every single setting we support in here with a comment about what it's doing and its default value.

PR #285, opened 2025-07-24
Testing and candidate implementation for binary timestamp sorting

### LINKED ISSUE #273: Cross-shard: sord by timestamp/tz
### Description

Add support for parsing Postgres timestamps (binary & text encoding) and add support for sorting by timestamp in a multi-shard (cross-shard) context.

```sql
CREATE TABLE users (
    id BIGSERIAL PRIMARY KEY,
    email VARCHAR,
    created_at TIMESTAMPTZ
);

-- Cross-shard order by
SELECT * FROM users ORDER BY created_at DESC
```

jaggederest:
Closes #273 

Not entirely sure about the logic for the actual binary parsing, my rustfu is limited, but I've validated the tests carefully, so if the implementation has issues the tests are at least a reasonable guideline

### DIFF
--- a/pgdog/src/net/error.rs
+++ b/pgdog/src/net/error.rs
@@ -73,6 +73,9 @@ pub enum Error {
     #[error("wrong size binary ({0}) for type")]
     WrongSizeBinary(usize),
 
+    #[error("invalid timestamp components")]
+    InvalidTimestamp,
+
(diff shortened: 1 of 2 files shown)


levkk (approved):
Really great stuff, thank you! Definitely turned out to be more complex than just parsing an integer...!

PR #317, opened 2025-08-08
WIP : LogicalTranasactions in client

niclaflamme:


integration/load_balancer/run.sh
```diff
@@ -10,7 +10,7 @@ export PGHOST=127.0.0.1
 export PGDATABASE=postgres
 export PGPASSWORD=postgres
 
-docker-compose up -d
+docker compose up -d
```
niclaflamme, line 13: `docker-compose` is deprecated.
`docker compose` has replaced it.

`docker-compose` does not work on my recent version of `docker` desktop.
`docker compose` works on the version of docker in our CI.

win-win?
I suppose it will work for a every engineer's machine too. Can rollback this change if that is not the case

pgdog/src/frontend/client/inner.rs
```diff
@@ -58,7 +59,7 @@ impl Inner {
         buffer: &mut Buffer,
         prepared_statements: &mut PreparedStatements,
         params: &Parameters,
-        in_transaction: bool,
+        logical_transaction: &LogicalTransaction,
```
niclaflamme, line 58: this is the root change, followed the compiler from here basically.

pgdog/src/frontend/client/mod.rs
```diff
@@ -627,13 +641,24 @@ impl Client {
         // ReadyForQuery (B)
         if code == 'Z' {
             inner.stats.query();
-            // In transaction if buffered BEGIN from client
-            // or server is telling us we are.
-            self.in_transaction = message.in_transaction() || inner.start_transaction.is_some();
```
levkk, line 632: This is so strange to read, I can't remember why we're checking `inner.start_transaction.is_some()` here because I don't think it's even possible: we send it over to the server before `server_message` handler ever runs.

I think it may be a left over from when we were processing client/server messages independently from each other. Now we process one request at a time: https://github.com/levkk/pgdog/blob/33f1ec5b3cd002cfd7396d7f5af69675475f61cd/pgdog/src/frontend/client/mod.rs#L610-L619

I reckon it should be safe to remove the `inner.start_transaction.is_some();` check entirely.

pgdog/src/frontend/client/mod.rs
```diff
@@ -627,13 +641,24 @@ impl Client {
...
-            inner.stats.idle(self.in_transaction);
+            // 1) Does the backend server say we're in a transaction?
+            let should_be_tx = message.in_transaction() || inner.start_transaction.is_some();
+
+            // 2) Is the frontend client in a logical transaction?
+            let in_transaction = self.in_transaction();
+
+            // 3) Reconcile against our LogicalTransaction
+            if should_be_tx && !in_transaction {
```
levkk: I'd probably clarify that the query parser isn't always active and therefore we don't catch `BEGIN` statements always. So sometimes, the only way to know we're inside a transaction is because the server told us.

levkk (approved):
Cool!

pgdog/src/frontend/client/mod.rs
niclaflamme: I think you're right, I also think it might be cheap to check twice. I'll check 👀

pgdog/src/frontend/client/mod.rs
niclaflamme: still looking but it seems like:

- `inner.start_transaction` was storing a `[Q] "BEGIN"`.clone() in its Option<BufferedQuery>
- then relaying it in the body of `client_messages()` like so:
  ```rust
  
     // We don't start a transaction on the servers until
     // a client is actually executing something.
     //
     // This prevents us holding open connections to multiple servers
     if self.request_buffer.executable() {
            if let Some(query) = inner.start_transaction.take() {
                inner.backend.execute(&query).await?;
            }
      }

  ```
- then later, inside `self.start_transaction()` we were sending a
- `messages.push(ReadyForQuery::in_transaction(true).message()?.backend());`

TLRDR:
- It seems like we were sending `BEGIN` twice to the backend and swallowing the warning from the extraenous one.
- Once in the body of `client_resquest()` once in the body of `start_transaction`
- Removing and letting CI / integrations tests run to make sure i'm not missing something.

pgdog/src/frontend/client/mod.rs
```diff
@@ -444,24 +447,21 @@ impl Client {
                 Some(Command::StartTransaction(query)) => {
                     if let BufferedQuery::Query(_) = query {
                         self.start_transaction().await?;
-                        inner.start_transaction = Some(query.clone());
```
levkk, line 447: This is important to record. `BEGIN` can have multiple arguments that affect what kind of transaction we're starting: https://www.postgresql.org/docs/current/sql-begin.html

pgdog/src/backend/pool/connection/binding.rs
```diff
@@ -237,6 +237,13 @@ impl Binding {
         Ok(())
     }
 
+    /// Execute a BEGIN on all servers
+    /// TODO: Block mutli-shard BEGINs as transaction should not occur on multiple shards
+    pub async fn begin(&mut self) -> Result<(), Error> {
+        let query = Query::new("BEGIN");
```
levkk, line 244: See other comment. Client can pass options to `BEGIN`, changing transaction type.

PR #327, opened 2025-08-12
[Resharding] Update sequences

levkk:


vitabaks:
@levkk maybe this will help you, this is how I usually transfer sequences:
```shell
# Copy sequences from source database
psql -h <source_dbhost> \
-U postgres \
-d <source_dbname>  \
-tAXc "
  select format(
    'select setval(%L, %s);',
    format('%I.%I', schemaname, sequencename),
    coalesce(last_value, 1) + 1000000
  )
  from pg_sequences
" > pg_sequences_increase.sql

# Apply new sequence values on target database
psql -U postgres -d <target_dbname> -f pg_sequences_increase.sql
```

Please note that here we copy the current value of all the sequences and add 1 million to them, usually this is enough, but in any case it is individual for each cluster because it depends on the load.

levkk:
I'm using the primary schema dump as the source of truth and separating statements into 3 phases:

- pre logical replication sync (tables, defaults, primary key index)
- post logical replication sync (foreign keys, secondary indexes)
- cutover sync (set sequence id to max(id) + 1)

PR #333, opened 2025-08-15
[Query engine] Refactor client into query engine

levkk:
### Description

- Refactor Client into QueryEngine in preparation for 2-phase commit and per-statement routing.
- Track bytes sent from intercepted  commands.
- Setup script drops databases to make sure they are clean.
- Skip simple protocol prepared statements test; they don't work and require per-statement query routing.


pgdog/src/frontend/client/query_engine/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+        match command {
+            Command::Shards(shards) => self.show_shards(context, *shards).await?,
+            Command::StartTransaction(begin) => {
+                self.start_transaction(context, begin.clone()).await?
+            }
+            Command::CommitTransaction => {
+                if self.backend.connected() {
+                    self.query(context, &route).await?
+                } else {
+                    self.end_transaction(context, false).await?
```
niclaflamme, line 127: If all these are self.query (event on commit and rollback) should it be self.execute?

PR #342, opened 2025-08-18
Deteremine transaction type based on query. `BEGIN` vs `BEGIN READ ONLY`

niclaflamme:
## Changes

- Introduced `TransactionType` enum with `ReadOnly` and `ReadWrite` variants to classify transactions.
- Replaced `in_transaction` boolean with `transaction: Option<TransactionType>` in `Client` and `QueryEngineContext` structs for more precise state tracking.
- Updated transaction checks to infer state via `transaction.is_some()`, eliminating the need for a separate boolean.
- Added `detect_transaction_type` function in `begin.rs` to parse `BEGIN` queries (using `pg_query`) and detect options like `READ ONLY` or `READ WRITE`.
- Made minimal adjustments to tests and related code (e.g., in `execute.rs`, `parse.rs`) to maintain compatibility without breaking existing functionality.

### DIFF
--- a/pgdog/src/frontend/client/query_engine/start_transaction.rs
+++ b/pgdog/src/frontend/client/query_engine/start_transaction.rs
@@ -1,4 +1,9 @@
-use crate::net::{CommandComplete, Protocol, ReadyForQuery};
+use pg_query::protobuf::{a_const, node, Node, TransactionStmtKind};
+
+use crate::{
+    frontend::client::TransactionType,
+    net::{CommandComplete, Protocol, ReadyForQuery},
+};
 
 use super::*;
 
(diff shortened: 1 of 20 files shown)


pgdog/src/frontend/client/query_engine/start_transaction.rs
```diff
@@ -9,20 +14,152 @@ impl QueryEngine {
...
+        "",
+        "INVALID",
+    ];
+
+    let read_write_queries = vec![
+        "BEGIN",
+        "BEGIN;",
+        "begin",
+        "bEgIn",
+        "BEGIN WORK",
```
levkk, line 118: 40 years of syntax changes, so many ways to say the same thing

PR #352, opened 2025-08-21
Transaction on ClientRequest

niclaflamme:
### Changes (wip)
- Renamed `Buffer` to `ClientRequest`.
- Renamed `Buffer.buffer` to `ClientRequest.messages`
- Added `ClientRequest.route`
- removed the `deref` on `Buffer` so that it reads `client_request.messages.iter()` instead of `client_request.iter()`

### En-passant changes:
- Added a condition to the `context.transaction` setting in server events, prevent a defined `ReadOnly` to be overwritten to a `ReadWrite`

```rust
let in_transaction = message.in_transaction() || self.begin_stmt.is_some();
if !in_transaction {
    context.transaction = None;
}

// Query parser is disabled, so the server is responsible for telling
// us we started a transaction. Do not override an existing transaction state.
if in_transaction && context.transaction.is_none() {
    context.transaction = Some(TransactionType::ReadWrite);
}
```


pgdog/src/frontend/client/query_engine/query.rs
```diff
@@ -98,9 +98,11 @@ impl QueryEngine {
...
             if !in_transaction {
                 context.transaction = None;
-            } else if context.transaction.is_none() {
-                // Query parser is disabled, so the server is responsible for telling us
-                // we started a transaction.
+            }
+
+            // Query parser is disabled, so the server is responsible for telling
+            // us we started a transaction. Do not override an existing transaction state.
+            if in_transaction && context.transaction.is_none() {
```
levkk: That's what the `else if` does basically:

`if !in_transaction` evaluates to `false` (that means it's `true`) and triggers the execution of `else if` which then checks `context.transaction.is_none()`.

niclaflamme: actually, 100% correct.
i'll put it back, no idea why i read it wrong.

PR #353, opened 2025-08-21
Remove a route clone()

niclaflamme:


niclaflamme:
> Cloning the route isn't so bad. If we have a large object inside of it, we can put it into an `Arc`.

It's not the memory so much as it's that we're carrying a bunch of copies of the truth, rather than referrer to a single boolean.

PR #365, opened 2025-08-27
Numeric rough draft, includes binary protocol

jaggederest:
Still missing:

- [ ] NaN support
- [ ] Test and code cleanup - working but not terse
- [ ] Infinity and -Infinity coverage in tests and code


levkk:
The failing integration test is coming from the `SHOW STATS` admin command, which is a bit odd, since the multi-shard connection which performs the sorting shouldn't be used.

pgdog/src/frontend/router/sharding/vector.rs
```diff
@@ -15,7 +16,8 @@ impl Distance<'_> {
                 assert_eq!(p.len(), q.len());
                 p.iter()
                     .zip(q.iter())
-                    .map(|(p, q)| (**q - **p).powi(2))
+                    // TODO: vectors should use f64?
```
levkk: pgvector uses f32

pgdog/src/net/messages/data_row.rs
```diff
@@ -188,7 +189,8 @@ impl DataRow {
     // Get float at index with text/binary encoding.
     pub fn get_float(&self, index: usize, text: bool) -> Option<f64> {
         self.get::<Numeric>(index, if text { Format::Text } else { Format::Binary })
-            .map(|numeric| *numeric.deref())
+            // TODO: unused? loses precision?
```
levkk: Not sure, this is f32 -> f64, should be fine?

pgdog/src/net/messages/data_types/double.rs
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    fn encode(&self, encoding: Format) -> Result<Bytes, Error> {
+        match encoding {
+            Format::Text => {
+                let s = if self.0.is_nan() {
+                    "NaN".to_string()
+                } else if self.0.is_infinite() {
+                    if self.0.is_sign_positive() {
+                        "Infinity".to_string()
```
levkk, line 78: Is it ok you're using lowercase for encoding and uppercase for decoding?

pgdog/src/net/messages/data_types/numeric.rs
```diff
@@ -70,44 +81,319 @@ impl Add for Numeric {
...
+                    Err(e) => {
+                        // Check for special PostgreSQL values
+                        match s.to_uppercase().as_str() {
+                            "NAN" => {
+                                // TODO: Add NaN support
+                                warn!("NaN values not supported (deferred feature)");
+                                Err(Error::UnexpectedPayload)
+                            }
+                            "INFINITY" | "+INFINITY" | "-INFINITY" => {
+                                warn!("Infinity values not supported");
```
levkk:
```suggestion
                                warn!("numeric infinity values not supported");
```

pgdog/src/net/messages/data_types/numeric.rs
```diff
@@ -70,44 +81,319 @@ impl Add for Numeric {
...
+                                Err(Error::UnexpectedPayload)
+                            }
+                            _ => {
+                                warn!("Failed to parse numeric: {}", e);
```
levkk:
```suggestion
                                warn!("failed to parse numeric: {}", e);
```

pgdog/src/net/messages/data_types/numeric.rs
```diff
@@ -70,44 +81,319 @@ impl Add for Numeric {
...
+                    0x4000 => true,  // Negative
+                    0xC000 => {
+                        // NaN - not supported yet
+                        // TODO: Add NaN support for binary format
+                        warn!("NaN values not supported in binary format (deferred feature)");
+                        return Err(Error::UnexpectedPayload);
+                    }
+                    _ => {
+                        // Invalid sign value
+                        warn!("Invalid numeric sign value: 0x{:04x}", sign);
```
levkk:
```suggestion
                        warn!("invalid numeric sign value: 0x{:04x}", sign);
```

pgdog/src/net/messages/data_types/numeric.rs
```diff
@@ -70,44 +81,319 @@ impl Add for Numeric {
...
+                            // Pad with zeros if needed
+                            for _ in fractional_str.len()..(dscale as usize) {
+                                result.push('0');
+                            }
+                        }
+                    }
+                }
+
+                let decimal = Decimal::from_str(&result).map_err(|e| {
+                    warn!("Failed to parse '{}' as Decimal: {}", result, e);
```
levkk:
```suggestion
                    warn!("failed to parse '{}' as Decimal: {}", result, e);
```

pgdog/src/net/messages/data_types/numeric.rs
```diff
@@ -70,44 +81,319 @@ impl Add for Numeric {
...
             }
 
-            Format::Binary => Ok(Self {
-                data: match bytes.len() {
-                    4 => bytes.get_f32() as f64,
-                    8 => bytes.get_f64(),
-                    n => return Err(Error::WrongSizeBinary(n)),
-                },
-            }),
+            Format::Binary => {
```
levkk, line 132: This looks like a pretty computationally expensive algorithm, but maybe that's what it takes to parse these?

pgdog/src/net/messages/data_types/double.rs
jaggederest: I believe the standard is to be case insensitive on ingest, but case-sensitive on emit. let me check the actual pg implementation though.

pgdog/src/frontend/router/sharding/vector.rs
jaggederest: Hmm. Should I change it to use f32 throughout? It's definitely using f64 right now.

levkk: Hmm. Maybe? It's been a minute since I touched this. I think...yes.

pgdog/src/net/messages/data_types/numeric.rs
jaggederest: Yeah I believe we should probably document that it's slow, but when you need exact math, you need exact math, so I suspect people in that situation will be okay-ish with the performance hit.

PR #370, opened 2025-08-28
Mirror metrics

jaggederest:
Add mirror metrics, including error rate and error type breakdown. Also fixes mirroring bug with multiple users and adds config options to their own section instead of below the [[database]]

Working on this, I think this is actually legitimately close to a CDC-style migration like old school Slony, but without all the hassle and triggers. I think if we implemented the behind the scenes copy to bring the mirror up to date we could easily offer this as a zero-touch <9.4-to-17 upgrade in a can with basically existing tools. I don't think there's *anything* like this on the market at all.

- [x] integration tests demonstrate the issue
- [x] proper structs instead of random grab bag of types
- [x] reasonably clean code
- [x] queue_length parameter
- [ ] check and ~possibly disallow~ ensure we have working mirroring + sharding
- [ ] validate that queue depth does something
- [ ] check mirroring exposure tests and/or write some that use less-than-100% exposure
- [ ] Add warning that local settings override global settings for mirroring
- [ ] Ensure local mirroring settings are actually used per above


pgdog/src/admin/show_mirror_stats.rs
```diff
@@ -0,0 +1,221 @@
...
+        "SHOW MIRROR_STATS_BY_DATABASE".into()
+    }
+
+    fn parse(_: &str) -> Result<Self, Error> {
+        Ok(Self)
+    }
+
+    async fn execute(&self) -> Result<Vec<Message>, Error> {
+        let fields = vec![
+            Field::text("database"),
```
levkk, line 125: The identifier for a "database" is `(database, user)`. We create a `Cluster` for each user/database pair. Both fields are available on the `Cluster`:

```rust
let database = cluster.name();
let user = cluster.user();
```

.gitignore
```diff
@@ -45,3 +45,4 @@ perf.data.old
 
 /pgdog.toml
 /users.toml
+CLAUDE.local.md
```
levkk, line 48: Interesting! Share it with the class! We got CLAUDE.md around the code somewhere.

pgdog/src/backend/databases.rs
```diff
@@ -174,7 +174,7 @@ impl ToUser for (&str, Option<&str>) {
 pub struct Databases {
     databases: HashMap<User, Cluster>,
     manual_queries: HashMap<String, ManualQuery>,
-    mirrors: HashMap<String, Vec<Cluster>>,
+    mirrors: HashMap<String, Vec<(String, f32, usize)>>, // Vec<(destination, exposure, queue_depth)>
```
levkk: You might want to `struct` that. Been  down this road before, even with different data types, things get confusing real quick.

.gitignore
jaggederest: Well, CLAUDE.local.md is explicitly intended for private matters, like how I want it to call me grand high poobah. CLAUDE.md is shared. I can commit my plan directory if you're interested in the full dope! (will do that now)

pgdog/src/config/mod.rs
```diff
@@ -222,6 +222,9 @@ pub struct Config {
     /// Replication config.
     #[serde(default)]
     pub replication: Replication,
+
+    #[serde(default)]
```
levkk, line 226: Let's remove the `mirror_` settings from `General`. I don't think they do anything anymore or do they  still have an effect?

pgdog/src/backend/pool/connection/mirror/mod.rs
```diff
@@ -71,6 +72,19 @@ impl Mirror {
     /// Handler for sending queries to the background task.
     ///
     pub fn spawn(cluster: &Cluster) -> Result<MirrorHandler, Error> {
+        let config = config();
+        Self::spawn_with_config(
+            cluster,
+            config.config.general.mirror_exposure,
```
levkk, line 78: It looks like we're using the general mirror settings. Are we then ignoring the ones set in the `[[mirrors]]` config?

jaggederest: Yep! Oversight on my part, great catch.

levkk:
> I also want to manually check out what happens when you try to mirror + shard, I suspect that for now we probably want to explicitly disallow that, and I want to test both the exposure percentages and the queue depth

We actually definitely want to support sharding via mirrored databases. It's one of the ways people can test sharding without actually sharding their primary DB.

pgdog/src/backend/pool/connection/mirror/mod.rs
levkk: I think we can keep the general settings and let the ones set on individual mirrors be override the general ones. Maybe make them optional on the `[[mirrors]]` config and only use them if they are set? Seems like a good backwards-compatible change.

jaggederest: I think that probably adds user confusion but I'm happy to do it if you prefer that. It would be a smaller change surface area, for sure. I'll add a warning about "using local mirroring exposure to override global mirroring exposure setting" and ensure that it actually uses the local one 👍

levkk: No need for the warning I think. This is how all our other settings work, for  example `databases.server_user` overrides `users.name`,  `users.pooler_mode` overrides `general.pooler_mode`, etc.

PR #377, opened 2025-08-29
Add testing and stats tweaks - prospective change that deserves independent review

jaggederest:
When pools are banned, the mirrored requests all showed as successful. The tests demonstrate it, but I wanted to review this separately so it could have an independent go/nogo.

I'm pretty sure the tests are useful, they're just kind of edge case behavior.

### DIFF
--- a/integration/rust/tests/integration/test_mirror_with_sharding.rs
+++ b/integration/rust/tests/integration/test_mirror_with_sharding.rs
@@ -220,9 +220,9 @@ async fn test_mirroring_to_sharded_cluster() {
     let (_source_conn, mut shard0_conn, mut shard1_conn) =
         setup_databases().await.expect("Failed to setup databases");
 
-    // Connect through PgDog to source_db
+    // Connect through PgDog to pgdog database
     let (client, connection) =
-        tokio_postgres::connect("postgres://pgdog:pgdog@127.0.0.1:6432/source_db", NoTls)
+        tokio_postgres::connect("postgres://pgdog:pgdog@127.0.0.1:6432/pgdog", NoTls)
             .await
             .expect("Failed to connect through PgDog");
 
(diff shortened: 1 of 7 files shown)


integration/rust/tests/integration/test_mirror_with_sharding.rs
```diff
@@ -0,0 +1,725 @@
...
+"#;
+
+        std::fs::write(&self.config_file, config)?;
+        std::fs::write(&self.users_file, users_config)?;
+        Ok(())
+    }
+
+    fn start_pgdog(&self) -> PgdogGuard {
+        PgdogGuard {
+            child: Command::new("/Users/justin/Code/lev/pgdog/target/release/pgdog")
```
levkk: This won't work outside of your setup :) Danger of letting Claude loose!

integration/rust/tests/integration/test_mirror_with_sharding.rs
```diff
@@ -0,0 +1,725 @@
...
+            users_file: format!("/tmp/mirror_test_{}_users.toml", test_name),
+            exposure,
+        }
+    }
+
+    fn write_configs(&self) -> Result<(), std::io::Error> {
+        let config = format!(
+            r#"
+[general]
+openmetrics_port = 9090
```
levkk, line 40: You sure these run? These integration tests require pgdog to be running already, this would conflict on the port. I think if you want to test a custom config inside the integration test, add it to the integration config (`integration/pgdog.toml`) as a separate DB. We don't have a limit on how many databases that can go in there!

.claude/plans/mirror-stats-bug-2b.md
```diff
@@ -0,0 +1,250 @@
+# Mirror Stats Bug Fix Plan - Phase 2B
```
levkk: Second thought, maybe let's not commit these, I'm getting confused between what code is and what's temporary claude state during reviews!

jaggederest: Yes, I agree, I will expunge them. I also just basically pushed this because I wasn't sure if the approach was right, they need a cleanup pass as you've 100% observed

integration/rust/tests/integration/test_mirror_with_sharding.rs
jaggederest: Yes it'll be moved there, I just wanted a completely clean suite. We also need a way to test varying global settings but that's a less pressing problem.

levkk: True! I keep adding them here as they come up: https://github.com/levkk/pgdog/blob/26816db293b03aaf7875e855d6f76c1813282c9a/pgdog/src/admin/set.rs#L55

Works for testing general settings as long as the test suite runs serially (hence `--test-threads=1`). For more advanced stuff, we should be able to load a new config dynamically, e.g. `SET CONFIG <config encoded as base64 or something easy to pass to a query>`. Just for making testing easier, I can't imagine this would be used in prod.

PR #393, opened 2025-09-03
Log when dropping transaction

jaggederest:


pgdog/src/backend/pool/connection/mirror/handler.rs
```diff
@@ -96,7 +96,7 @@ impl MirrorHandler {
     /// Flush buffered requests to mirror.
     pub fn flush(&mut self) -> bool {
         if self.state == MirrorHandlerState::Dropping {
-            debug!("mirror transaction dropped");
+            warn!("mirror transaction dropped");
```
levkk: This is dropping when exposure is < 1.0. They are asking to know when it's dropped because the queue is full. They are using it for replication so their exposure is 1.0.

levkk (approved):
LGTM!

PR #403, opened 2025-09-03
Admin interface: Add mirror stats and mirror stats per cluster

jaggederest:


pgdog/src/admin/show_mirrors.rs
```diff
@@ -0,0 +1,127 @@
...
+            Field::numeric("dropped_count"),
+            Field::numeric("error_count"),
+        ];
+
+        let mut messages = vec![RowDescription::new(&fields).message()?];
+
+        // Iterate through all clusters and create a row for each
+        for (user, cluster) in databases().all() {
+            let stats = cluster.stats();
+            let stats = stats.lock();
```
levkk, line 35: Nit: drop the lock before building the message.

levkk (approved):
LGTM!

PR #405, opened 2025-09-03
New [[mirroring]] config and related code changes

jaggederest:


pgdog/src/backend/databases.rs
```diff
@@ -472,18 +454,59 @@ pub fn from_config(config: &ConfigAndUsers) -> Databases {
...
+    };
+
+    // Validate mirroring configurations and collect valid ones
+    let mut valid_mirrors = std::collections::HashSet::new();
+
+    for mirror_config in &config.config.mirroring {
+        let source_users = get_database_users(&mirror_config.source_db);
+        let dest_users = get_database_users(&mirror_config.destination_db);
+
+        if !source_users.is_empty() && !dest_users.is_empty() && source_users == dest_users {
```
levkk, line 485: We shouldn't insert mirrors with no users: their pools wont be activated.

pgdog/src/backend/databases.rs
```diff
@@ -472,18 +454,59 @@ pub fn from_config(config: &ConfigAndUsers) -> Databases {
...
+            valid_mirrors.insert((
+                mirror_config.source_db.clone(),
+                mirror_config.destination_db.clone(),
+            ));
+        } else {
+            error!(
+                "Mirroring disabled for {} -> {}: invalid configuration (source exists: {}, destination exists: {}, users match: {})",
```
levkk:
```suggestion
                "mirroring disabled from \"{}\" into \"{}\": users don't match",
```

levkk:
Couple of small suggestions, good to go otherwise.

pgdog/src/config/mod.rs
```diff
@@ -346,6 +350,40 @@ impl Config {
     pub fn multi_tenant(&self) -> &Option<MultiTenant> {
         &self.multi_tenant
     }
+
+    /// Get mirroring configuration for a specific source/destination pair.
+    pub fn get_mirroring_config(
```
levkk, line 355: This does the search on every client connection. You would be better off storing this directly on the cluster, like we did with mirrors before and retrieving a pre-built list. You can think of `Cluster`, `Pool`, `Databases`, etc. as immutable configuration objects: generated from config on load time and store various settings, incl. mirroring. When we reload the config, old versions of them are destroyed and replaced with new ones with new config values.

This contrasts with `Client` which is long-lived and survives config reloads, that's why we fetch its config from the global config structure on each new query request.

pgdog/src/backend/databases.rs
jaggederest: I believe this checks that both source_users and dest_users are not empty, unless I'm misunderstanding what you're saying?

levkk: You're right, I think it was late when I read this :)

PR #412, opened 2025-09-04
Add queue_length as a metric, including prom/admin

jaggederest:


pgdog/src/backend/pool/connection/mirror/handler.rs
```diff
@@ -151,6 +152,18 @@ impl MirrorHandler {
...
+    pub fn increment_queue_length(&self) {
+        let mut stats = self.stats.lock();
+        stats.counts.queue_length += 1;
+    }
+
+    /// Decrement the queue length.
+    pub fn decrement_queue_length(&self) {
+        let mut stats = self.stats.lock();
+        stats.counts.queue_length = stats.counts.queue_length.saturating_sub(1);
+    }
```
levkk, line 166: Smart! No overflows!

pgdog/src/backend/pool/connection/mirror/mod.rs
```diff
@@ -113,6 +113,11 @@ impl Mirror {
                 select! {
                     req = rx.recv() => {
                         if let Some(mut req) = req {
+                            // Decrement queue_length when we receive a message from the channel
+                            {
```
levkk, line 117: Don't you have a method for this?

jaggederest: Yes, but it's not available on that handler, only the main struct that we don't pass in. Handler has to use the long version, rip.

Issue #206, opened 2025-06-05
Support 2-phase commit


levkk:
### Description

Add support for 2-phase commit in Postgres to have consistent distributed transactions across shards.

Docs: https://www.postgresql.org/docs/current/two-phase.html

levkk:
This is done.

PR #419, opened 2025-09-08
Implement two-phase commit (2pc)

levkk:
### Changelog

- Implement two-phase commit. #206 

### Two-phase commit

Also known as **2pc** or **2-phase commit**. Postgres supports it out of the box by enabling `max_prepared_transactions`:

```sql
ALTER SYSTEM SET max_prepared_transactions TO 1000;
```

Once enabled in Postgres, enable 2pc in PgDog as well:

```toml
[general]
two_phase_commit = true
```

#### How it works

2pc splits the `COMMIT` command into 2 statements:

1. `PREPARE TRANSACTION '<unique name>'`
2. `COMMIT PREPARED '<unique name>'`

When the client sends `COMMIT` to us, we intercept it and run statement 1 on all shards. If that succeeds, we run statement 2 on all shards. If that succeeds as well, the transaction is finished and we send `CommandComplete` and `ReadyForQuery` to the client.

**Why is this necessary?**

Separating the `COMMIT` statement into two protects against single-shard failures. If statement 1 fails on any of the shards, PgDog automatically rolls back the transaction on _all_ shards by running `ROLLBACK PREPARED '<unique name>'`.

If statement 2 fails for any of the shards, PgDog will automatically retry it asynchronously until it succeeds. This ensures that the transaction is _eventually_ committed in full. This ensures that the database is _eventually consistent_, even if errors during this process occur.

During normal ops, the delay between statement 1 and statement 2 is very short. Other clients may see a partially updated database during that stage.

#### Automatic 2pc for cross-shard writes

2pc is automatically executed for single-statement write transactions. For example, if your client sends a cross-shard update, like:

```sql
UPDATE users SET admin = true WHERE created_at < NOW()
```

PgDog will transform this, under the hood, into 4 statements:

1. `BEGIN`
2. `UPDATE users SET admin = true WHERE created_at < NOW()`
3. `PREPARE TRANSACTION`
4. `COMMIT TRANSACTION`

This ensures that all writes use 2pc even if a transaction hasn't been explicitly started.

**Disable automatic 2pc**

If you don't need 2pc for single-statement transactions, either because you don't care about consistency or because you can safely retry writes in case of failure (idempotent statements), you can disable automatic 2pc in the config:

```toml
[general]
two_phase_commit_auto = false
```

### DIFF
--- a/pgdog/src/admin/parser.rs
+++ b/pgdog/src/admin/parser.rs
@@ -7,7 +7,8 @@ use super::{
     show_instance_id::ShowInstanceId, show_lists::ShowLists, show_mirrors::ShowMirrors,
     show_peers::ShowPeers, show_pools::ShowPools, show_prepared_statements::ShowPreparedStatements,
     show_query_cache::ShowQueryCache, show_servers::ShowServers, show_stats::ShowStats,
-    show_transactions::ShowTransactions, show_version::ShowVersion, shutdown::Shutdown, Command, Error,
+    show_transactions::ShowTransactions, show_version::ShowVersion, shutdown::Shutdown, Command,
+    Error,
 };
 
 use tracing::debug;


pgdog/src/backend/pool/connection/mod.rs
```diff
@@ -31,6 +31,8 @@ use std::{
 
 pub mod aggregate;
 pub mod binding;
+#[cfg(test)]
```
jaggederest, line 34: slick, do we want these to be at the end or is there a `cargo fmt` rule about it?

levkk: No rule for that one! Could be anywhere.

PR #433, opened 2025-09-10
Refactor Backend::Server into submodules

jaggederest:


jaggederest:
Hmm. I don't think the refactor should have fewer lines than the original... But let's see what integration has to say.

levkk:
Quick win here would be to move tests into their own file. Most other things in that module are tighly coupled. For those, you could do what we did for query engine and parser: move closely related functions into their own files and expose them with `pub(super)`.

PR #462, opened 2025-09-14
fix: correct aggregate function matching in parser

Prabhat1308:
found this bug while working to implement `avg()`


levkk (approved):
Good catch, thanks!

Issue #468, opened 2025-09-15
[Auth] GSSAPI


levkk:
### Description

Implement Kerberos/GSSAPI however Postgres does it.

levkk:
Probably won't add this anytime soon unless someone _really_ needs it.

PR #479, opened 2025-09-18
GSSAPI

jaggederest:
#468

Completely raw, but testing Works On My Machine(tm)

I suspect this will need a bunch of cleanup and rework but it's worth looking at to see the scale of the project.

### DIFF
--- a/pgdog/src/auth/gssapi/context.rs
+++ b/pgdog/src/auth/gssapi/context.rs
@@ -55,22 +55,32 @@ impl GssapiContext {
             .add(&GSS_MECH_KRB5)
             .map_err(|e| GssapiError::LibGssapi(format!("failed to add mechanism: {}", e)))?;
 
-        // Try to acquire credentials from cache first, then fall back to keytab
+        // Parse the principal name first so we can try to acquire for the specific principal
+        let principal_name = Name::new(principal.as_bytes(), Some(&GSS_NT_KRB5_PRINCIPAL))
+            .map_err(|e| GssapiError::InvalidPrincipal(format!("{}: {}", principal, e)))?;
+
+        // Try to acquire credentials for the specific principal
+        // This helps avoid "Principal in credential cache does not match desired name" errors
         let credential = match Cred::acquire(
-            None, // Try default principal from cache
+            Some(&principal_name), // Try specific principal first
             None,
             CredUsage::Initiate,
             Some(&desired_mechs),
         ) {
             Ok(cred) => cred,
             Err(cache_err) => {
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -13,7 +13,7 @@ jobs:
           override: true
       - name: Install CMake 3.31
         run: |
-          sudo apt update && sudo apt install mold -y
+          sudo apt update && sudo apt install -y mold clang pkg-config protobuf-compiler
```
levkk, line 16: We have protobuf! We are finally complete

pgdog/src/auth/gssapi/context.rs
```diff
@@ -0,0 +1,226 @@
...
+                // Set the keytab path for client use
+                std::env::set_var("KRB5_CLIENT_KTNAME", keytab.to_string_lossy().as_ref());
+
+                // Set a unique cache to avoid conflicts
+                let cache_file = format!(
+                    "/tmp/krb5cc_pgdog_context_{}_{}",
+                    principal.replace(['@', '.', '/'], "_"),
+                    std::process::id()
+                );
+                std::env::set_var("KRB5CCNAME", format!("FILE:{}", cache_file));
```
levkk, line 83: Doesn't look like access to this function is synchronized. `set_var` isn't thread-safe & if we have multiple creds (e.g., for multiple databases), this will race I think.

pgdog/src/auth/gssapi/context.rs
```diff
@@ -0,0 +1,226 @@
...
+
+                // Acquire credentials using the keytab
+                Cred::acquire(
```
levkk, line 86: This calls into gssapi right? How expensive is this call? We might want to offload this to `spawn_blocking` to avoid blocking Tokio.

pgdog/src/auth/gssapi/context.rs
jaggederest: Yes, there's a lot of issues like that here. Especially with the env variables but the KDC is kind of a synchronization point so we'll need to use ScopedEnv or something.

pgdog/src/auth/gssapi/context.rs
jaggederest: It's pretty heavy. There's a tension here, because Cred::acquire is blocking and times out if there's a problem (great api design), but shelling out to kinit or other command line tools is probably even more unreliable. in the v2 branch I'm trying to get either an async wrapper around it or a quicker failure through a proper timer in a different thread but it's tricky.

Issue #434, opened 2025-09-10
[Query engine] Support avg


levkk:
### Description

Add support for `avg()` aggregate function. This requires rewriting the query as follows:

1. Replace `avg` with `count(*)` and `avg()`
2. Execute query on all shards
3. Aggregate perform `sum(count() * avg()) / sum(count())` on result from each shard

PR #502, opened 2025-09-25
Add AVG aggregate function coverage, expression parser that rides on AST, RewriteEngine that produces ghost columns

jaggederest:
#434

We now have AVG support on simple + extended protocols.

New conceptual objects:

- ExpressionParser uses the AST to normalize and compare expressions to see if they're logically equivalent (and thus subject to rewriting or not).
- RewriteEngine takes an AST (might be doing its own parsing, but will fix) and changes it according to the logic in the Route from the frontend, emitting the AST/SQL and a RewritePlan which tells the QueryEngine what to do with the aggregate (drop columns that we've added, essentially, though we can do other rewrites later).
  I expect the RewriteEngine will want to have plugin hooks for additional rewrite rules. Also probably want to migrate any other query rewriting from the Frontend there.

Potential issues with cache expiring when structure changes, that kind of thing. Pretty raw, but conceptually the direction I was thinking of. Will follow up with more refinement.


pgdog/src/frontend/client/query_engine/query.rs
```diff
@@ -264,9 +264,8 @@ impl QueryEngine {
     async fn transaction_error_check(
         &mut self,
         context: &mut QueryEngineContext<'_>,
-        route: &Route,
     ) -> Result<bool, Error> {
-        if context.in_error() && !context.rollback && route.is_cross_shard() {
```
levkk, line 255: I don't remember why the cross-shard check is here. But I think this code is tested, so should be good.

Edit: Maybe it was because cross-shard transactions have their `BEGIN`/`COMMIT` statements intercepted, so that's the only  time we need to respond here?

jaggederest: Yeah it started failing the tests when that conditional was there, so I left it out. Honestly I am not 100% clear, so I'd have to write a test to falsify it if I thought we still needed it

pgdog/src/frontend/router/parser/rewrite_engine.rs
```diff
@@ -0,0 +1,271 @@
...
+#[derive(Default)]
+pub struct RewriteEngine;
+
+impl RewriteEngine {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Rewrite a SELECT query, adding helper aggregates when necessary.
+    pub fn rewrite_select(&self, sql: &str, aggregate: &Aggregate) -> RewriteOutput {
```
levkk, line 16: Shouldn't we pass the AST we already parsed here?

jaggederest: Yes, it was on my cleanup todo but I got distracted by stddev 👍 will do now.

jaggederest: Let's do it next PR, I think this is acceptable for now. the `stddev` branch handles this correctly at least.

pgdog/src/frontend/client/query_engine/query.rs
jaggederest: Checked out this code, it actually handles the edge case where before a couple queries could leak through if they weren't cross shard, even though the transaction was aborted.

PR #507, opened 2025-09-25
Add pure-sql comparison run to integration

jaggederest:
Ideally this runs each pure SQL scenario 6 times:

[Baseline, PGDog, PGDog + Sharding] * [Format::Text, Format::Binary]

There are separate setup-teardown files that will be run before/after each run (but not asserted against), and the return values for each case should be evaluated to be exactly equal in both row count, data types, and data values between all six arms.

I haven't figured out how to handle cases where we deliberately violate the postgresql spec, or how to mark tests as pending, we may need to do something like a SQL comment for those.

### DIFF
--- a/integration/run.sh
+++ b/integration/run.sh
@@ -5,4 +5,5 @@ pushd ${SCRIPT_DIR}
 bash python/run.sh
 bash ruby/run.sh
 bash java/run.sh
+bash sql/run.sh
 popd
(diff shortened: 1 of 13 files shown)


pgdog/src/frontend/listener.rs
```diff
@@ -169,6 +169,11 @@ impl Listener {
                     }
                 }
 
+                Startup::GssEnc => {
+                    // GSS encryption is not yet supported; reject and wait for a normal startup.
+                    stream.send_flush(&SslReply::No).await?;
```
levkk, line 174: Nice! I bet that's what was the blocker with that integration we were talking about the other day.

levkk (approved):
Sweet!

Issue #496, opened 2025-09-24
[Query engine] Pool banning improvements


levkk:
### Description

A couple improvements:

1. Don't trigger a ban iff only a single database is configured
2. Return the original error to the client instead of the opaque "pool banned" message

The error is returned on first error only. All subsequent queries don't use the broken connection pool.

baznikin:
Is this issue can help with our situation?

We started using pgdog on subset of services and today faced with strange bans (most likely due to increased traffic to service):

```
2025-09-26T14:12:57.562940Z  INFO new server connection [develop-postgresql:5432, core]
2025-09-26T14:12:59.599767Z  INFO client "core_owner_user" connected to database "core" [10.244.2.230:58160]
2025-09-26T14:12:59.599878Z  INFO client "core_owner_user" connected to database "core" [10.244.1.7:41326]
2025-09-26T14:12:59.600024Z  INFO client "core_owner_user" connected to database "core" [10.244.1.7:41328]
2025-09-26T14:12:59.601186Z  INFO client "core_owner_user" connected to database "core" [10.244.2.230:58164]
2025-09-26T14:12:59.601390Z  INFO client "core_owner_user" connected to database "core" [10.244.1.7:41338]
2025-09-26T14:12:59.601585Z  INFO client "core_owner_user" connected to database "core" [10.244.1.7:41350]
2025-09-26T14:13:04.602250Z ERROR checkout timeout [[10.244.1.7:41326]]
2025-09-26T14:13:04.602419Z ERROR pool is banned [[10.244.1.7:41296]]
2025-09-26T14:13:04.602414Z ERROR pool is banned [[10.244.1.7:41328]]
2025-09-26T14:13:04.602505Z ERROR pool is banned [[10.244.2.230:58160]]
2025-09-26T14:13:04.602522Z ERROR pool is banned [[10.244.1.7:42986]]
2025-09-26T14:13:04.602544Z ERROR pool is banned [[10.244.1.7:41350]]
2025-09-26T14:13:04.602565Z ERROR pool is banned [[10.244.1.7:41338]]
2025-09-26T14:13:04.602566Z ERROR pool is banned [[10.244.2.230:58164]]
2025-09-26T14:13:04.614568Z  INFO client "core_owner_user" disconnected from database "core" [10.244.1.7:41328]
2025-09-26T14:13:04.614568Z  INFO client "core_owner_user" disconnected from database "core" [10.244.1.7:41350]
[...]

Issue #483, opened 2025-09-19
PgDog instance going beyond "default_pool_size" connections


albert-toro:
Hello!

I believe this is an issue regarding the maximum number of connections an instance of PgDog can handle.

Referring to the [official documentation](https://docs.pgdog.dev/configuration/pgdog.toml/general/#default_pool_size), `default_pool_size` is "Default maximum number of server connections per database pool. The pooler will not open more than this many PostgreSQL database connections when serving clients."

Assuming this is true, I deployed a kubernetes pod of pgdog with the following configmap:

```
# PgDog configuration
[general]
host = "0.0.0.0"
port = ${var.pgdog_port}
openmetrics_port = ${var.pgdog_metrics_port}
default_pool_size = 620
min_pool_size = 1
pooler_mode = "session"
idle_timeout = 60000
query_timeout = 86400000
shutdown_timeout = 10000
load_balancing_strategy = "least_active_connections"
healthcheck_interval = 30000
idle_healthcheck_interval = 30000
checkout_timeout = 5000
connect_timeout = 5000
workers = 4
prepared_statements = "extended"
rollback_timeout = 5000 

[[databases]]
name = "${var.pgdog_database_name}"
host = "${var.pgdog_database_IP}"
port = 5432
role = "primary"

# TCP connection settings
[tcp]
keepalive = false
# retries = 10
# time = 10000
# interval = 1000
# user_timeout = 30000

[[users]]
name = "${var.pgdog_username}"
password = "${var.pgdog_password}"
database = "${var.pgdog_database_name}"
replication_mode = false
```

If I'm not wrong, this pod should only handle a maximum of 620 connections. However, after an intense workload, this single pod reached 2000 concurrent connections at once. Is this a bug or is there something I'm missing?

For further information, I'm using PostgreSQL 14 and the default `pgsql` driver.

Thank you!

albert-toro:
It was a fresh deploy in a different environment. My guess is that perhaps `pool_size` takes preference whether it's specified or not? And since it's not specified, PgDog interpreted it as unlimited.

But rest assured, as soon as I'm able to, I'll reproduce the high load scenario and fetch the information straight from the admin database!

albert-toro:
Hello! I ended up deploying PgDog in a new environment with the following settings:

```
default_pool_size = 5
min_pool_size = 1
```

```
admin=> SHOW POOLS;
[...]

PR #512, opened 2025-09-29
Tune banning

levkk:
#496 

### Description

Refactored banning to be triggered by failed health checks and checkout timeouts only. The banning logic is moved into the `Replicas` module and decoupled from connection pools themselves. The connection pools report their health, while the replicas load balancer decides to ban them (or not) based on that status, which is checked 3 times per second.

Any errors received from checking out connections from the  connection pool, like `CheckoutTimeout` are treated as bannable as well.

The connection pool doesn't stop its internal machinery when its banned. It continues to make connections to the instance and run health checks.

### Bug fixes

- Fixed idle connection leak when replenishing connection pool that was recently shut down. Possible fix for #483.

### Code changes

- `pool::Inner` doesn't have `ban` anymore and will replenish connections unless paused.
- `Pool` updates its health on healthcheck, check-in error or checkout timeout.
- `Pool::get` is now cancel-safe.
- `Inner::put` will give connection to a waiter if _any_ is still in the wait queue, even if some have dropped.

### Other features

- Added `client_login_timeout` setting allowing to limit how long clients are allowed to attempt to login into the pooler.
- Added `HEALTCHECK` command to trigger health check on a connection pool.
- Setting `ban_timeout` to 0 disables banning.


pgdog/src/backend/pool/healthcheck.rs
jaggederest, line 39:
```suggestion
    /// Perform a mandatory healthcheck.
```

A number of typos like this - might want to `git grep -i healtcheck`

integration/failover/pgdog.toml
jaggederest, line 1: I think, for future integration suites, this is a good policy - if each one has a separate pgdog.toml/users.toml we can run suites separately. shared config is messy.

integration/toxi/toxi_spec.rb
```diff
@@ -78,22 +77,89 @@ def warm_up
     # Connect (the pool is lazy)
     Sharded.where(id: 1).first
     errors = 0
+    ok = 0
```
jaggederest, line 80: Might want a `Concurrent::AtomicFixnum` here?

integration/failover/pgdog.toml
levkk: I agree! It is faster tho and you can see how one affects the other. I've caught quite a few bugs when one test left some "bad state" that another test triggered. Pros and cons.

pgdog/src/backend/pool/healthcheck.rs
levkk: Yeah, some day ... I'll learn how to spell health check. Is it one word? Or two?

pgdog/src/admin/healthcheck.rs
```diff
@@ -0,0 +1,43 @@
...
+}
+
+#[async_trait]
+impl Command for Healthcheck {
+    fn name(&self) -> String {
+        "HEALTHCHECK".into()
+    }
+
+    fn parse(sql: &str) -> Result<Self, Error> {
+        let parts = sql.split(" ").collect::<Vec<_>>();
```
jaggederest, line 16: might want a more liberal whitespace here, in case of newline etc

pgdog/src/backend/pool/replicas/ban.rs
```diff
@@ -0,0 +1,396 @@
...
+    inner: Arc<RwLock<BanInner>>,
+    pool: Pool,
+}
+
+impl Ban {
+    /// Create new ban handler.
+    pub(super) fn new(pool: &Pool) -> Self {
+        Self {
+            inner: Arc::new(RwLock::new(BanInner { ban: None })),
+            pool: pool.clone(),
```
jaggederest, line 19: (For my edification, not review:) Pools are immutable when cloned here, right? so we don't have any issue with the clone diverging from parent?

pgdog/src/admin/healthcheck.rs
levkk: Agreed. We should write our own parser for these, the right way (regex :D)

pgdog/src/backend/pool/replicas/ban.rs
levkk: They are `Sync`, which means access to its internals is safe. Cloning it, 99.99% of time is completely safe. There was an edge case, in the `duplicate` method, which duplicated the wrong instance of it, but that's a separate, confusing pattern, which I'd like to remove. Otherwise, the TLDR is, cloning it is safe and cheap.

Edit: furthermore, everything in `Shard`, incl. replicas, pools, etc. is all behind an `Arc`, so anything not behind a mutex or similar is immutable and we don't even call clone on any of it, because we only clone `Shard`.

integration/toxi/toxi_spec.rb
levkk: Don't need one, these are not parallel.

PR #516, opened 2025-09-29
Rewrite bug fixes - old cache key, more efficient update

### LINKED ISSUE #514: [Query engine] Query rewrite busts prepared statement cache
### Description

Rewriting queries in a cross-sharded context busts the prepared statement cache. The cache deduplicates statements by their SQL text. The code changes the cache key for a statement, so all _other_ clients that will send the same statement will result in creating independent prepared statements, per client.

https://github.com/levkk/pgdog/blob/01f8612daa6d9bcf16451925e26c895b15640eec/pgdog/src/frontend/prepared_statements/global_cache.rs#L224-L238

### LINKED ISSUE #515: [Query engine] Inefficient prepared statement cache update
### Description

When rewriting queries for cross-shard average calculation, the prepared statement global cache is accessed twice in a row, taking the rw lock twice in a row. This can be replaced by a single operation.

https://github.com/levkk/pgdog/blob/01f8612daa6d9bcf16451925e26c895b15640eec/pgdog/src/frontend/router/parser/query/select.rs#L102-L104

jaggederest:
closes #514 
closes #515

Adds a cache key reuse test, combines update_query method into update_and_set_rewrite_plan (is this actually more efficient?)

### DIFF
--- a/pgdog/src/frontend/prepared_statements/global_cache.rs
+++ b/pgdog/src/frontend/prepared_statements/global_cache.rs
@@ -419,7 +419,16 @@ mod test {
             names.push(name);
         }
 
-        assert_eq!(cache.close_unused(0), 0);
+        assert_eq!(cache.close_unused(0), 25);
+        assert!(cache.is_empty());
+
+        names.clear();
+        for stmt in 0..25 {
+            let parse = Parse::named("__sqlx_1", format!("SELECT {}", stmt));
+            let (new, name) = cache.insert(&parse);
+            assert!(new);
+            names.push(name);
+        }
 
         for name in &names[0..5] {
             assert!(!cache.close(name, 25)); // Won't close because


pgdog/src/frontend/prepared_statements/global_cache.rs
```diff
@@ -221,27 +219,28 @@ impl GlobalCache {
...
+    pub fn update_and_set_rewrite_plan(
+        &mut self,
+        name: &str,
+        sql: &str,
+        plan: RewritePlan,
+    ) -> bool {
         if let Some(statement) = self.names.get_mut(name) {
             let old_key = statement.cache_key();
             let cached = self.statements.remove(&old_key);
             statement.parse.set_query(sql);
```
levkk, line 234: Why not just mutate it in-place? Less confusing (incl. for our AI reviewer :))

pgdog/src/frontend/prepared_statements/global_cache.rs
```diff
@@ -221,27 +219,28 @@ impl GlobalCache {
...
-            let new_key = statement.cache_key();
+            if !plan.is_noop() {
+                statement.rewrite_plan = Some(plan);
+            }
             if let Some(entry) = cached {
-                self.statements.insert(new_key, entry);
+                self.statements.insert(old_key, entry);
```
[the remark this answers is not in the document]
jaggederest: That's... that's the point.

pgdog/src/frontend/client_request.rs
```diff
@@ -181,10 +181,7 @@ impl ClientRequest {
             if let ProtocolMessage::Parse(parse) = message {
                 parse.set_query(query);
                 let name = parse.name().to_owned();
-                let _ = prepared.update_query(&name, query);
-                if !plan.is_noop() {
-                    prepared.set_rewrite_plan(&name, plan.clone());
-                }
+                let _ = prepared.update_and_set_rewrite_plan(&name, query, plan.clone());
```
levkk:
```suggestion
                prepared.update_and_set_rewrite_plan(&name, query, plan.clone());
```

The `_` is typically used to ignore errors so people freak out when they see it :D

pgdog/src/frontend/prepared_statements/global_cache.rs
jaggederest: You're absolutely right! I should update the cached item in place instead of removing and readding it! (channeling Claude for a second there)

pgdog/src/frontend/prepared_statements/global_cache.rs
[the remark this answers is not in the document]
levkk: LOL

pgdog/src/frontend/client_request.rs
```diff
@@ -181,10 +181,7 @@ impl ClientRequest {
             if let ProtocolMessage::Parse(parse) = message {
                 parse.set_query(query);
                 let name = parse.name().to_owned();
```
levkk: Why are you cloning the name only to pass it by reference later? I think LLVM would typically optimize this but still kinda weird to read.

PR #520, opened 2025-09-30
Fix SCRAM authentication DoS vulnerabilities and add IP-based rate limiting

NikolayS:
This PR addresses SCRAM authentication DoS vulnerabilities by fixing async runtime blocking and adding IP-based rate limiting.

## Changes

**1. Fix SCRAM PBKDF2 Blocking Async Runtime**
- Moved PBKDF2 computation (4096 iterations, ~50-80ms) to blocking thread pool
- Prevents attackers from exhausting async worker threads with authentication spam

**2. Add Per-IP Authentication Rate Limiting** _(new feature)_
- Configurable rate limiting: 10 attempts per minute per IP by default
- Token bucket algorithm using the `governor` crate
- IPv6-aware: normalizes to /64 prefix to prevent bypass attacks
- Memory-bounded: hourly state reset prevents unbounded growth

## Configuration

```toml
# Authentication rate limit (attempts per minute per IP address)
# Default: 10
auth_rate_limit = 10
```

---

## Follow-up Work

(this would be a separate fix)

SCRAM verifier caching (pre-existing issue): The current SCRAM implementation (pre-existed before this PR) generates a new random salt on every authentication attempt, which violates the SCRAM-SHA-256 protocol design and wastes ~50-80ms of PBKDF2 computation per auth. Possible solution: pre-compute verifiers (salt, iterations, stored_key, server_key) at user creation time and store them in the config file, similar to how PostgreSQL stores them in `pg_shadow`.


pgdog/src/auth/scram/server.rs
```diff
@@ -50,10 +50,20 @@ use base64::prelude::*;
...
     fn get_password_for(&self, _user: &str) -> Option<PasswordInfo> {
-        // TODO: This is slow. We should move it to its own thread pool.
         let iterations = 4096;
         let salt = rand::thread_rng().gen::<[u8; 16]>().to_vec();
-        let hash = hash_password(&self.password, NonZeroU32::new(iterations).unwrap(), &salt);
+
+        // Move expensive PBKDF2 computation to blocking thread pool
+        // to avoid blocking the async runtime.
+        // Note: Using block_in_place() because AuthenticationProvider trait is synchronous.
+        let hash = tokio::task::block_in_place(|| {
```
levkk, line 58: That's ... better than what was there, but ideally we run SCRAM in a background thread overall. Requires a bigger refactor which we don't have to do now.

pgdog/src/frontend/client/mod.rs
```diff
@@ -150,6 +150,20 @@ impl Client {
             }
         };
 
+        if let Some(addr) = *stream.peer_addr() {
+            if !rate_limit::check(addr.ip()) {
```
levkk: This should be optional. Default should be unlimited. It's completely reasonable to receive thousands of these per second in a internal setting, especially during incidents when databases go down and all apps reconnect.

pgdog/src/frontend/client/mod.rs
```diff
@@ -150,6 +150,20 @@ impl Client {
...
+                error!(
+                    "Authentication rate limit exceeded for IP: {}, user: \"{}\", database: \"{}\"",
```
levkk, line 199: What's the unit? Per minute/second?

pgdog/src/frontend/client/mod.rs
NikolayS: Fixed. Rate limiting now defaults to unlimited (disabled). Only applies when auth_rate_limit is explicitly configured -- should handle econnect storm scenario

pgdog/src/frontend/client/mod.rs
NikolayS: Clarified - it's per minute per IP. Updated the config field documentation and example.pgdog.toml to make this explicit.

PR #522, opened 2025-09-30
swap to u16 on inbound and outbound paths

### LINKED ISSUE #493: [Query engine] Treat number of parameters in Bind as unsigned
### Description

So, according to [this](https://www.postgresql.org/docs/current/limits.html), Postgres supports 65k params in a query, while according to [that](https://www.postgresql.org/docs/current/protocol-message-formats.html#PROTOCOL-MESSAGE-FORMATS-BIND), it can only support 32k bound params. Is int16 actually uint16? Seems to be the case, and I think parsing it as so won't break anything, since clients aren't going to send negative number of parameters....right?

jaggederest:
closes #493

### DIFF
--- a/pgdog/src/net/messages/bind.rs
+++ b/pgdog/src/net/messages/bind.rs
@@ -269,7 +269,7 @@ impl FromBytes for Bind {
                 _ => Format::Binary,
             })
             .collect();
-        let num_params = bytes.get_i16();
+        let num_params = bytes.get_u16();
         let params = (0..num_params)
             .map(|_| {
                 let len = bytes.get_i32();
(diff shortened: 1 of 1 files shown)


levkk:
This one we should test against Postgres because their spec says its a signed 16 bit integer, but one of our users said they were able to prepare and execute a statement with like 50k params so we need to validate who's right.

jaggederest:
Added a test, looks like at least my local backend supports 65k parameters, we'll see what CI thinks. Will also convert the `num_codes` as suggested in the comment on the related ticket.

levkk:
Do we have an integration test that ensures we can process a query with 65k params? I suspect there could be more places where we use i16 for counting params.

levkk (approved):
Cool!

PR #523, opened 2025-09-30
TLS hot swap

### LINKED ISSUE #488: [TLS] Make cert settings changeable at runtime
### Description

Make `tls_*` settings changeable at runtime, re-creating TLS acceptors.

jaggederest:
closes #488


pgdog/src/net/tls.rs
```diff
@@ -12,91 +15,92 @@ use tokio_rustls::rustls::{
...
+    match (new_acceptor, tls_paths) {
+        (Some(acceptor), Some((cert, _))) => {
+            let acceptor = Arc::new(acceptor);
+            let previous = ACCEPTOR.swap(Some(acceptor));
 
-    connector()?;
+            if previous.is_none() {
+                info!(cert = %cert.display(), "🔑 TLS enabled");
+            } else {
+                info!(cert = %cert.display(), "🔁 TLS certificate reloaded");
```
levkk, line 121: Nice touch!

levkk:
Why is the connector no longer cached?

levkk:
The connector is a concern, since it was previously cached in a `OnceCell` and is now created for each server connection, which underneath does a sync IO call to the OS certificate store: https://github.com/pgdogdev/pgdog/blob/feat/tls_hotswap/pgdog/src/net/tls.rs#L177

jaggederest:
> The connector is a concern, since it was previously cached in a `OnceCell` and is now created for each server connection, which underneath does a sync IO call to the OS certificate store: https://github.com/pgdogdev/pgdog/blob/feat/tls_hotswap/pgdog/src/net/tls.rs#L177

Let me add a test for that too, good to keep track of.

PR #524, opened 2025-10-01
Fix: cancel backend sessions when shutdown timeout expires

### LINKED ISSUE #517: Cancel queries on shutdown
### Description

Cancel active running queries when shutting down. This helps to stop long-running queries when PgDog terminates.

Postgres doesn't stop running queries if a client disconnects. it only stops once it detects the socket is dead, which doesn't happen until it starts sending the response back which, with long queries, could take hours.

KR-bluejay:
## Related Issue  
closes #517  

## Description  
Ensure active queries are cancelled during shutdown.  
If the shutdown timeout elapses, PgDog now sends `CANCEL` requests to Postgres before tearing down connection pools, preventing long-running queries from lingering.  

## Code changes  
- **frontend/listen.rs**: Added query cancellation in `execute_shutdown` when shutdown timeout expires.  

## Testing  
Currently, there don’t seem to be existing tests covering shutdown + query cancellation.  
To properly test this, we would need to:  
1. Start a client running a long-running query.  
2. Trigger PgDog shutdown with a short timeout.  
3. After PgDog exits, verify that the query on Postgres was actually cancelled.  

Since this type of integration test isn’t present yet, I’m unsure of the best approach to add it here. Guidance would be appreciated.

### DIFF
--- a/pgdog/src/frontend/listener.rs
+++ b/pgdog/src/frontend/listener.rs
@@ -131,6 +131,14 @@ impl Listener {
                 "terminating {} client connections due to shutdown timeout",
                 comms.tracker().len()
             );
+
+            // Shutdown timeout elapsed; cancel any still-running queries before tearing pools down.
+            let cancel_futures = comms
+                .clients()
+                .into_keys()
+                .map(|id| async move { databases().cancel(&id).await });
+
+            let _ = futures::future::join_all(cancel_futures).await;
         }
 
         self.shutdown.notify_waiters();


pgdog/src/frontend/listener.rs
```diff
@@ -131,6 +131,31 @@ impl Listener {
                 "terminating {} client connections due to shutdown timeout",
                 comms.tracker().len()
             );
+
+            // Shutdown timeout elapsed; cancel any still-running queries before tearing pools down.
+            let cancel_futures = comms.clients().into_keys().map(|id| async move {
```
levkk: This is good. One note: let's make this code run conditionally on `shutdown_termination_timeout` being configured. Otherwise, this is good to go.

KR-bluejay: I had read #517 as “cancel queries by default on shutdown, and use shutdown_termination_timeout only for how long we wait.” 

If we guard the cancellation logic behind that option, we’d fall back to not cancelling at all. 

Was the intent to keep cancellation optional, or should we still do it by default and let the timeout act purely as a safeguard?

levkk: > If we guard the cancellation logic behind that option, we’d fall back to not cancelling at all.

That's right, we want this feature to be optional. If we allow query cancellation without a timeout, it can block pgdog from shutting down if the network is unreliable.

The timeout is a safeguard, and we can't use this feature without it safely.

KR-bluejay: Thanks for clarifying! 

I'll modify the code to only run cancellation when the timeout is set

KR-bluejay: Done. The cancellation now only runs when shutdown_termination_timeout is configured (updated in the latest commit).

Heads-up: once this lands, I’ll open an issue in the docs repo and follow up with the config update there.

levkk (approved):
Great! Thank you!

PR #527, opened 2025-10-01
Use parsed AST instead of double-parsing SELECT stmt

levkk:
### Description

- Don't parse `SELECT` statement twice when rewriting for ghost columns.


jaggederest (approved):
🚢

PR #533, opened 2025-10-03
Schema sync fixes and tests

levkk:
### Description

Use `pg_dump`'s entire output instead of per-table executions. This is much faster and also won't miss extensions and other data types we don't explicitly handle.


levkk:
Ignoring unrelated flaky toxi test.

Issue #541, opened 2025-10-06
[Query engine] Support sharding key updates


levkk:
### Description

When changing the sharding key, transform the query into a `DELETE` and an `INSERT` so the row is moved between shards.

scottjacobsen:
@jaggederest How do I configure this? I get an error when I add it to my config:

```toml
[general]
passthrough_auth = "enabled_plain"
rewrite_shard_key_updates = "ignore"
```

```
2025-11-04T17:47:14.355514Z  INFO 🐕 PgDog v0.1.13 [main@378c96e, rustc 1.91.0 (f8297e351 2025-10-28)]
Error: MissingField("unknown field `rewrite_shard_key_updates`, 
```

levkk:
This is done for single-row updates. For multi-row updates & rows with foreign key constraints, it's not. We'll add separate issues for those.

PR #555, opened 2025-10-09
Sharding key updates

jaggederest:
#541

This version just raises an error when sharding key updates are detected. Will add the actual rewriting shortly.

### DIFF
--- a/pgdog/src/frontend/client/query_engine/shard_key_rewrite.rs
+++ b/pgdog/src/frontend/client/query_engine/shard_key_rewrite.rs
@@ -190,6 +190,7 @@ impl QueryEngine {
         rows: usize,
         two_pc: bool,
     ) -> Result<(), Error> {
+        // Note the special case for 1 is due to not supporting multirow inserts right now
         let command = if rows == 1 {
             CommandComplete::from_str("UPDATE 1")
         } else {


pgdog/src/frontend/client/query_engine/shard_key_rewrite.rs
```diff
@@ -0,0 +1,530 @@
...
+        context: &mut QueryEngineContext<'_>,
+        plan: ShardKeyRewritePlan,
+    ) -> Result<(), Error> {
+        let cluster = self.backend.cluster()?.clone();
+        let use_two_pc = cluster.two_pc_enabled();
+
+        let source_shard = match plan.route().shard() {
+            Shard::Direct(value) => *value,
+            _ => {
+                return Err(Error::Router(router::Error::Parser(
```
levkk, line 39: Not sure if this is caught somewhere else, but you want to throw an error that says something like:

```
updating multiple rows not supported when updating the sharding key
```

This would be caught by a query like this:

```postgresql
UPDATE users SET id = $1 WHERE admin = true;
```

for example. It's nonsensical because it'll trigger a unique constraint violation but it'll be a cross-shard query that updates the sharding key nonetheless.

pgdog/src/frontend/client/query_engine/shard_key_rewrite.rs
```diff
@@ -0,0 +1,530 @@
...
+        let mut source = cluster
+            .primary(source_shard, &request)
+            .await
+            .map_err(|err| Error::Router(router::Error::Pool(err)))?;
+        let mut target = cluster
+            .primary(target_shard, &request)
+            .await
+            .map_err(|err| Error::Router(router::Error::Pool(err)))?;
+
+        source.execute("BEGIN").await?;
```
levkk, line 73: You want to check if you're not inside a transaction already first. If you are, you should probably abort.

pgdog/src/frontend/client/query_engine/shard_key_rewrite.rs
```diff
@@ -0,0 +1,530 @@
...
+                let _ = target.execute("ROLLBACK").await;
+                self.send_update_complete(context, 0).await
+            }
+            RewriteOutcome::Applied { deleted_rows } => {
+                if use_two_pc {
+                    let identifier = cluster.identifier();
+                    let transaction_name = self.two_pc.transaction().to_string();
+                    let guard_phase_one = self.two_pc.phase_one(&identifier).await?;
+
+                    let prepare_source = format!(
```
levkk: We have existing machinery to perform two-phase commit. Take a look inside `end_transaction.rs` sister file. It takes care of updating stats as well, and other cleanup.

pgdog/src/frontend/client/query_engine/shard_key_rewrite.rs
```diff
@@ -0,0 +1,530 @@
...
+            Err(err) => {
+                let _ = source.execute("ROLLBACK").await;
+                let _ = target.execute("ROLLBACK").await;
+                return Err(err);
+            }
+        };
+
+        match outcome {
+            RewriteOutcome::Noop => {
+                let _ = source.execute("ROLLBACK").await;
```
levkk: Consider re-using existing code which rollsback transactions. Take a look in `end_transaction.rs`

pgdog/src/frontend/client/query_engine/shard_key_rewrite.rs
```diff
@@ -0,0 +1,530 @@
...
+    Ok(SqlResult {
+        row_description,
+        data_row,
+        command_complete,
+    })
+}
+
+fn build_delete_sql(plan: &ShardKeyRewritePlan) -> Result<String, Error> {
+    let mut sql = format!("DELETE FROM {}", plan.table());
+    if let Some(where_clause) = plan.statement().where_clause.as_ref() {
```
levkk, line 338: I don't think an `UPDATE` without a where-clause should even be allowed to update sharding keys. Maybe this is just the artifact of parsing the AST, but we should definitely catch that case.

pgdog/src/frontend/client/query_engine/shard_key_rewrite.rs
```diff
@@ -0,0 +1,703 @@
...
+        }
+    }
+
+    async fn send_update_complete(
+        &mut self,
+        context: &mut QueryEngineContext<'_>,
+        rows: usize,
+        two_pc: bool,
+    ) -> Result<(), Error> {
+        let command = if rows == 1 {
```
levkk, line 194: Why the special handling of rows == 1 here?

jaggederest: We don't support inserts with >1 row so I special cased it to throw a fit when rows > 1, since if we `DELETED 10` for example, we can't `INSERT ...` in a sharded context basically. I will probably try to get multirow inserts to work next, I think that's a big blocker for sharding, but I think it needs a li'l refactor, all these special cases are getting complicated.

Issue #259, opened 2025-07-10
Route rows in a multi-row INSERT separately to different shards


niclaflamme:
### Description
PgDog is routing multi-row (multi-parameter) INSERT queries on sharded tables as `::All` when the sharding keys (e.g., `user_id`) across rows hash to different shards. This results in the write being duplicated across all shards, which is not desired. Instead, I expect the rows to be grouped by their target shards and routed as `::Multi(Vec<int>)`, with the query potentially rewritten/split to send only relevant rows to each shard group (similar to how COPY commands are handled for cross-shard ingestion).

### Context
- PgDog has three shard routing types: `::All`, `::Direct(int)`, `::Multi(Vec<int>)`.
- For a single-row INSERT on a sharded key (e.g., `user_id`), the query is correctly routed to the appropriate shard via hashing or the tenant list (`::Direct`).
- For multi-row INSERTs - like in the following example - where the `user_id` values hash to different shards, it's currently routing as `::All` (broadcasting the full query to all shards, causing duplication).
  ```sql
  -- Routes to All (::All)
  INSERT INTO users (user_id, name) VALUES
    (123, 'Alice'),
    (456, 'Bob'),
    (789, 'Charlie');
  ```

### Expected behavior
Group rows by target shard (e.g., rows for shard 0, rows for shard 1), then route as `::Multi` with separate INSERT queries sent to each relevant shard containing only its rows. This avoids duplication and ensures efficient writes.

If I have a multi-insert query, my expectation is that it would behave like COPY, splitting rows across shards without duplication.

#### Single Shard Example (Separate Single-Row INSERTs with ::Direct Routing)
Assuming hash-based sharding on `user_id` across 3 shards (for illustration; actual shard depends on your hashing logic or tenant list):

```sql
-- Routes to Shard 0 (::Direct(0))
INSERT INTO users (user_id, name) VALUES (123, 'Alice');

-- Routes to Shard 1 (::Direct(1))
INSERT INTO users (user_id, name) VALUES (456, 'Bob');

-- Routes to Shard 2 (::Direct(1))
INSERT INTO users (user_id, name) VALUES (789, 'Charlie');
```

#### Grouped Insert Example (Multi-Row INSERT with ::All Routing, Causing Duplication)
If the multi-row INSERT is routed as ::All (broadcast to all 3 shards), it duplicates the writes across shards. A subsequent `SELECT * FROM users ORDER BY user_id ASC;` (queried across all shards) would return tripled rows like this:

[...]

PR #571, opened 2025-10-21
Split sharded inserts to allow multirow insert into correct tables

jaggederest:
#259 

Currently multirow inserts are either erroring or silently being sent to all shards, which produces incorrect behavior.

### DIFF
--- a/pgdog/src/frontend/client/query_engine/insert_split.rs
+++ b/pgdog/src/frontend/client/query_engine/insert_split.rs
@@ -275,7 +275,7 @@ mod tests {
             },
         },
         net::{
-            messages::{Parse, Query, Sync},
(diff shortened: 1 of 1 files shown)


pgdog/src/backend/replication/logical/subscriber/stream.rs
```diff
@@ -247,8 +247,24 @@ impl StreamSubscriber {
             // we are able to replay changes we already applied safely.
             if let Some(upsert) = statements.upsert.insert() {
                 let upsert = Insert::new(upsert);
-                let val = upsert.shard(&self.sharding_schema, Some(&bind))?;
-                self.send(&val, &bind).await?;
+                let cfg = crate::config::config();
```
levkk, line 250: We won't ever generate multi-tuple insert statements here.

Issue #569, opened 2025-10-20
Add an option to enforce TLS from clients


zxaos:
> Sending plain text passwords over unencrypted connections isn't ideal, even if PgDog and Postgres are on the same local network. For this reason, passthrough_auth = "enabled" will only work if PgDog is configured to use [TLS encryption](https://docs.pgdog.dev/features/tls/).

This is good, but clients can still opt to not use TLS at all and expose the password. It would be helpful to be able to reject non-TLS connections from PgDog entirely.

levkk:
Awesome! Yeah, should be pretty easy. The config should be on our catch-all `General` struct, which you can then pull at connection startup by calling:

```rust
use crate::config::config;

let config = config();
let config_name = config.config.general.config_name;
```

I suspect you want to handle this inside `Client::login` where the config will already be available.

PR #587, opened 2025-10-29
add tls_client_required setting

zxaos:
First run at #569 

Things I'm unsure about:

- Do I need to do anything with for docs beyond editing `example.pgdog.toml`
- It feels like maybe `tls_client_required` should default to true if `passthrough_auth` is set since that's the behaviour we're trying to enforce, but that'd be backwards incompatible with current behaviour.

### DIFF
--- a/pgdog/src/frontend/client/mod.rs
+++ b/pgdog/src/frontend/client/mod.rs
@@ -20,7 +20,7 @@ use crate::net::messages::{
     Authentication, BackendKeyData, ErrorResponse, FromBytes, Message, Password, Protocol,
     ReadyForQuery, ToBytes,
 };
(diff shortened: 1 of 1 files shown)


pgdog/src/net/messages/error_response.rs
```diff
@@ -130,6 +130,18 @@ impl ErrorResponse {
         }
     }
 
+    pub fn tls_required() -> ErrorResponse {
+        Self {
+            severity: "FATAL".into(),
+            code: "08004".into(),
+            message: "Clients must connect with TLS".into(),
```
levkk:
```suggestion
            message: "only TLS connections are allowed".into(),
```

levkk:
Cool! No worries.

Issue #578, opened 2025-10-23
Support protocol version 3.2 and the NegotiateProtocolVersion message


JelteF:
With [PG18 the protocol version 3.2 was introduced](https://www.postgresql.org/about/news/postgresql-18-released-3142/) (the first minor protocol version bump ever in PG history). We intend to make minor protocol version bumps are frequent thing. Servers speaking the Postgres protocol should send a [NegotiateProtocolVersion](https://www.postgresql.org/docs/current/protocol-message-formats.html#PROTOCOL-MESSAGE-FORMATS-NEGOTIATEPROTOCOLVERSION) message with a version downgrade request to the client when they don't support the minor protocol version requested in the StartupMessage.

Currently libpq still defaults to requesting 3.0, because while postgres has supported NegotationProtocolVersion for many years, most proxies or other server implementations (like cockroach) do not yet support it. So hereby a nudge to implement it in pgdog, so that hopefully in some time we can change libpq its defaults and have the protocol evolve quicker.

Also actually supporting protocol version 3.2 is nice for security, since you can send a longer cancellation key.

JelteF:
The easiest way is to use the `psql` command shipped with PG18 and then use: `psql 'max_protocol_version=3.2'`

sarajmunjal:
Hey folks! I have a PR up for this [here](https://github.com/pgdogdev/pgdog/pull/926). I added unit tests and integration tests using `psql` v18. Would appreciate a review, thanks!

PR #593, opened 2025-10-31
feat(docker): upgrade psql to 18

mscrivo:
Instead of relying on the one packaged for the version of the ubuntu image being used. This is good for a few reasons:
* pins the version of psql so it can be controlled
* upgrades from 16 to 18 to allow connecting to newer versions than 16, and [unlocks others](https://github.com/pgdogdev/pgdog/issues/578)

psql 18 should be fully compatible with all previous versions.

Testing this locally and it seems to be working fine:

```sh
root@1768d35d96f3:/pgdog# psql --version
psql (PostgreSQL) 18.0 (Ubuntu 18.0-1.pgdg24.04+3)
root@1768d35d96f3:/pgdog#
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,7 +31,6 @@ RUN install -d /usr/share/postgresql-common/pgdg && \
     sh -c "echo 'deb [signed-by=/usr/share/postgresql-common/pgdg/apt.postgresql.org.asc] https://apt.postgresql.org/pub/repos/apt $VERSION_CODENAME-pgdg main' > /etc/apt/sources.list.d/pgdg.list"
 
 RUN apt update && apt install -y postgresql-client-${PSQL_VERSION}
-RUN apt remove -y curl
 
 COPY --from=builder /build/target/release/pgdog /usr/local/bin/pgdog
 


mscrivo:
@levkk FYI. not sure how you have your notifications setup.

Dockerfile
```diff
@@ -21,9 +21,17 @@ RUN source ~/.cargo/env && \
...
+    apt install -y curl ca-certificates ssl-cert && \
     update-ca-certificates
 
+RUN install -d /usr/share/postgresql-common/pgdg && \
+    curl -o /usr/share/postgresql-common/pgdg/apt.postgresql.org.asc --fail https://www.postgresql.org/media/keys/ACCC4CF8.asc && \
+    . /etc/os-release && \
+    sh -c "echo 'deb [signed-by=/usr/share/postgresql-common/pgdg/apt.postgresql.org.asc] https://apt.postgresql.org/pub/repos/apt $VERSION_CODENAME-pgdg main' > /etc/apt/sources.list.d/pgdg.list"
+
+RUN apt update && apt install -y postgresql-client-18
+RUN apt remove -y curl
```
levkk: I wouldn't mind curl around to be honest. You can use it to query prom stats.

mscrivo: great, me too! I'll leave it in.

levkk (approved):
Awesome, thank you.

Issue #604, opened 2025-11-12
Problem with interplay between simple and extended prepared statements when using `prepared_statement = 'full'`


mijoharas:
Hey!

So, I managed to create a reliable repro of an issue I was having with prepared statements.

I've reproduced the issue on both `main` and `0.1.13` (which was the version we were using so the first we tested).

Start `pgdog` with `prepared_statements = 'full'` (basically just defaults other than that, other than hooking up a test db):

```
RUST_LOG=trace ./target/debug/pgdog -c ./local.pgdog.toml -u ./local.users.toml
```
Open another `psql` separately to your test db, and prepare a query (you should be able to execute it):
```
(michael)@127.0.0.1:6432 12:39:25 [cleo_development]
# prepare testplanmike1(int) as select 1 from users where id = $1;
PREPARE
(michael)@127.0.0.1:6432 12:39:29 [cleo_development]
# execute testplanmike1(1);
 ?column?
----------
        1
(1 row)
```

Then open something else that uses the extended query protocol, and run some prepared statements (I use rails, simplest way I had to do that was something like:
```ruby
ApplicationRecord.descendants.take(30).each do |bar|
  bar.where(id: 1).to_a;
end
```

Now, for me, (seemingly reliably) I get an error executing the extended protocol query:
```
ActiveRecord::StatementInvalid: PG::InvalidSqlStatementName: ERROR:  prepared statement "a5" does not exist (ActiveRecord::StatementInvalid)
from /home/michael/.local/share/mise/installs/ruby/3.4.5/lib/ruby/gems/3.4.0/bundler/gems/rails-014cfe9c63a6/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb:917:in 'PG::Connection#exec_prepared'
Caused by PG::InvalidSqlStatementName: ERROR:  prepared statement "a5" does not exist

from /home/michael/.local/share/mise/installs/ruby/3.4.5/lib/ruby/gems/3.4.0/bundler/gems/rails-014cfe9c63a6/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb:917:in 'PG::Connection#exec_prepared'
```
And, going back to the original `psql` the prepared statement we created no longer works:
```
(michael)@127.0.0.1:6432 12:39:31 [cleo_development]
# execute testplanmike1(1);
ERROR:  prepared statement "__pgdog_1" does not exist
```

I don't see any of the extended protocol prepared statements when I look in `show prepared;` in the admin db.
[...]

PR #605, opened 2025-11-12
Create parse_queries_enabled configuration option.

mijoharas:
Hey @levkk.

I wanted an option to enable the queries to be parsed but not have `prepared_statements = 'full'`.

This works around https://github.com/pgdogdev/pgdog/issues/604.

I wasn't sure about the name. I considered `parse_queries_override` (I also considered it just being `parse_queries` and making it `on` / `off` / `auto`) but realised that the `off` would just be a bug if you wanted some features that relied on parsing queries. having only `on` / `auto` might have made sense (with auto being the default) but hey, that's two states so I went with a boolean, even if false isn't really false. What do you think?

This option allows you to enable the query parser, in cases when the default setting would be off.
Currently the default setting for when queries would be parsed is when we either:
* have a primary and a replica.
* have more than one shard.
* have prepared_statements = 'full'

However there are cases where you might not want to have any of these, but still want to have your queries be parsed. So that, for instance, you can have client state (as in your `SET` commands) properly tracked across different clients.

### DIFF
--- a/pgdog-config/src/general.rs
+++ b/pgdog-config/src/general.rs
@@ -92,6 +92,7 @@ pub struct General {
     #[serde(default)]
     pub prepared_statements: PreparedStatements,
     /// Parse Queries override.
+    #[serde(default = "General::query_parser_enabled")]
     pub query_parser_enabled: bool,
     /// Limit on the number of prepared statements in the server cache.
     #[serde(default = "General::prepared_statements_limit")]


pgdog-config/src/general.rs
```diff
@@ -91,6 +91,8 @@ pub struct General {
     /// Prepared statatements support.
     #[serde(default)]
     pub prepared_statements: PreparedStatements,
+    /// Parse Queries override.
+    pub query_parser_enabled: bool,
```
levkk, line 96:
```suggestion
    #[serde(default = "General::query_parser_enabled")]
    pub query_parser_enabled: bool,
```

mijoharas: I pulled and tested it and yep, that's what you need to leave this option off (and it defaults to false). Thanks!

PR #617, opened 2025-11-20
State::IdleInTransaction isn't updated until the end of a request

jaggederest:


pgdog/src/frontend/client/query_engine/mod.rs
```diff
@@ -112,6 +112,8 @@ impl QueryEngine {
         self.stats
             .received(context.client_request.total_message_len());
 
+        self.comms.stats(self.stats);
```
levkk: We have an `update_stats` function somewhere I think?

integration/ruby/rspec_helper.rb
```diff
@@ -31,8 +31,10 @@ def ensure_done
     expect(pool['cl_waiting']).to eq('0')
     expect(pool['out_of_sync']).to eq('0')
   end
+  current_client_id = conn.backend_pid
   clients = conn.exec 'SHOW CLIENTS'
   clients.each do |client|
+    next if client['id'].to_i == current_client_id
```
levkk, line 37: Do you mean next if not?

levkk (approved):
Lgtm, although I'm not sure why the ruby test change is needed

PR #625, opened 2025-11-24
Fix idle_in_transaction handling during queries

jaggederest:
Not my favorite way to handle it, open to better suggestions


jaggederest:
Superseded by #626

PR #645, opened 2025-12-03
Fix aggregate group by count

mijoharas:
Addresses https://github.com/pgdogdev/pgdog/issues/638.

I had a bit of spare time, and was curious about this.

So, it turns out the issue is an incorrectly parsed out `group_by` most of the time. The `group_by` expects an index into the `target_list`, so I made that work for the case where you have the exact `table_name.column_name` or same `column_name` in the `select` and the `group by` queries.

this feels a little bit janky to me, but :shrug: figured I'd raise for feedback, especially as I see a draft PR for a rewritten query engine.

```sql
(michael)@127.0.0.1:6432 16:54:06 [repro_sharded]
> select count(1), user_id from example group by example.user_id;
 count | user_id
-------+---------
     6 |       1
     3 |       2
     4 |       3
(3 rows)

(michael)@127.0.0.1:6432 16:54:07 [repro_sharded]
> select count(1), example.user_id from example group by example.user_id;
 count | user_id
-------+---------
     6 |       1
     3 |       2
     4 |       3
(3 rows)

(michael)@127.0.0.1:6432 16:54:14 [repro_sharded]
> select example.user_id, count(1), example.user_id from example group by example.user_id;
unexpected field count in "D" message
(michael)@127.0.0.1:6432 16:54:21 [repro_sharded]
> select example.user_id, count(1) from example group by example.user_id;
 user_id | count
---------+-------
       1 |     6
       2 |     3
       3 |     4
(3 rows)

(michael)@127.0.0.1:6432 16:54:28 [repro_sharded]
> select user_id, count(1) from example group by example.user_id;
 user_id | count
---------+-------
       3 |     4
       1 |     6
       2 |     3
(3 rows)

(michael)@127.0.0.1:6432 16:54:32 [repro_sharded]
> select example.user_id, count(1) from example group by user_id;
 user_id | count
---------+-------
         |    13
(1 row)

(michael)@127.0.0.1:6432 16:54:37 [repro_sharded]
> select user_id, count(1) from example group by user_id;
 user_id | count
---------+-------
       3 |     4
       1 |     6
       2 |     3
(3 rows)
```

See here for results, including some cases where it fails to work correctly. (where we duplicate the column in the select, and where we specify inconsistent `example.user_id` / `user_id` (NOTE: it only fails if we're more specific in the `select` than in the `group_by`.

### DIFF
--- a/pgdog/src/frontend/router/parser/aggregate.rs
+++ b/pgdog/src/frontend/router/parser/aggregate.rs
@@ -1,7 +1,6 @@
 use pg_query::protobuf::Integer;
 use pg_query::protobuf::{a_const::Val, Node, SelectStmt, String as PgQueryString};
 use pg_query::NodeEnum;
-use std::collections::HashSet;
 
 use crate::frontend::router::parser::{ExpressionRegistry, Function};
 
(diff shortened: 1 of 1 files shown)


mijoharas:
oh, and this'd probably want some tests if we went with something like this. :sweat_smile:.

PR #649, opened 2025-12-04
Add SHOW CONFIG FILE to admin db

jaggederest:


pgdog/src/admin/show_config_file.rs
```diff
@@ -0,0 +1,55 @@
...
+    fn parse(_sql: &str) -> Result<Self, Error> {
+        Ok(Self {})
+    }
+
+    async fn execute(&self) -> Result<Vec<Message>, Error> {
+        let config = config();
+        let snapshot = sanitize_for_toml(config.config.clone());
+        let toml = toml::to_string_pretty(&snapshot)?;
+
+        let mut messages = vec![RowDescription::new(&[Field::text("pgdog.toml")]).message()?];
```
levkk, line 24: Might as well show both of them, either as rows or columns.

Issue #565, opened 2025-10-17
Protocol out of sync with Ruby timeouts


levkk:
### Description

We could end up in a situation where the protocol gets out of sync if a Ruby `Timeout` is used (abruptly terminate thread).

```
2025-10-15T21:27:27.244000+00:00 [...] 2025-10-15T21:27:27.244556Z ERROR ProtocolOutOfSync got: C, extended buffer: ProtocolState { queue: [], names: [], simulated: [], extended: true, out_of_sync: false }
```

My spidey sense is telling me the issue is in `Guard` cleanup logic, where we drain the connection when client abruptly disconnects. Probably needs additional state checks.

levkk:
fixed in #887

PR #667, opened 2025-12-09
Drain connections when more messages are expected

jbielick:
## Purpose

#565

The `needs_drain()` condition prior to this commit checks the connection state is some kind of idle or otherwise "in sync". However, even when the connection state is Idle, there can be expected messages in the protocol queue. If a client disconnects abruptly at just the right time, needs_drain() would not return true, but the protocol queue is not empty and expected messages from the server are not drained.

In this scenario, the connection is checked back in in a tainted state and subsequent usage can produce unexpected results (ProtocolOutOfSync or off-by-1 results mixups).

## Fix

This updates `needs_drain()` to also account for
`prepared_statements.has_more_messages()` in addition to the "in sync" check, thus triggering a drain when there are more messages expected from the server (and tracked in the protocol queue).

### DIFF
--- a/pgdog/src/backend/pool/test/mod.rs
+++ b/pgdog/src/backend/pool/test/mod.rs
@@ -653,3 +653,65 @@ async fn test_lsn_monitor() {
 
     pool.shutdown();
 }
+
+/// If a client disconnects while the backend still has protocol messages
+/// queued for it, but the high-level stats state is Idle, the connection must
+/// still be treated as needing a drain. Otherwise the connection is checked
+/// back into the pool with a non-empty protocol queue and will be "tainted"
+/// for the next client.
+#[tokio::test]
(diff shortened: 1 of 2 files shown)


pgdog/src/backend/pool/test/mod.rs
```diff
@@ -653,3 +653,65 @@ async fn test_lsn_monitor() {
...
+    crate::logger();
+
+    let pool = pool();
+    let mut guard = pool.get(&Request::default()).await.unwrap();
+
+    guard
+        .send(
+            &vec![
+                ProtocolMessage::from(Query::new("SELECT 1")),
+                ProtocolMessage::from(Query::new("SELECT 1")),
```
jbielick, line 673: Holding off on this PR because stuffing the ClientRequest like this doesn't seem to be a realistic reproduction. Addressing some of the out-of-sync side effects in #668 instead.

PR #668, opened 2025-12-10
Mark protocol state in Error when out of sync detected

jbielick:
## Purpose

This addresses some side effects of #565 in that when a connection falls out of sync (due to currently unknown series of events), the connection is detected as out of sync, but is checked back into the pool without any corrective action.

## Approach

This updates server.rs to set its stats state to State::Error when we encounter a ProtocolOutOfSync error while reading from the server and decided whether to forward the message to the client. The result is that the connection will be disconnected when attempted to be checked back into the pool and the tainted connection will not be re-used.

### DIFF
--- a/pgdog/src/backend/protocol/state.rs
+++ b/pgdog/src/backend/protocol/state.rs
@@ -208,6 +208,11 @@ impl ProtocolState {
         &self.queue
     }
 
+    #[cfg(test)]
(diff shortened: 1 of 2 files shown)


levkk (approved):
Awesome thank you so much. Happy to merge this in if you'd be willing to accept the CLA. Cheers!

levkk:
Finally found the root cause for this. Fix in #887

Issue #560, opened 2025-10-13
[Query engine] Remove query comments


levkk:
### Description

Remove query comments before storing them in the query cache and the prepared statement cache. This generates a lot of unique statements and blows both of those caches up.

They are commonly used by tracing software, e.g. Sentry, Datadog, etc.

#### Example

```postgresql
/application:web,correlation_id:b33bdd84b4c99dc24cca4c781a063875,endpoint_id:graphql:EndPoint,db_config_name:main/
SELECT [...]
```

### Note

Re-attach comment for traces if query blows up.

PR #749, opened 2026-02-04
remove query comments #560

dev-lew:
This PR addresses #560

It modifies the comment function (now parse_comment), which will capture a role and shard and then return the original query string with all comments removed except for those that match 0 or more regexes.

If there is a cache miss, we do another cache lookup after removing comments except for those used for pgdog metadata (pgdog_shard, etc) in the hope that we can increase cache hit rate.

Two associated helper functions were added as well.

Also, some tests were modified to reflect the new functionality.


pgdog/src/frontend/router/parser/comment.rs
```diff
@@ -86,6 +87,76 @@ pub fn comment(
...
+        QueryParserEngine::PgQueryRaw => scan_raw(query),
+    }
+    .map_err(Error::PgQuery)?;
+
+    Ok(result
+        .tokens
+        .iter()
+        .any(|st| st.token == Token::CComment as i32 || st.token == Token::SqlComment as i32))
+}
+
```
dev-lew: It is possible that this function is over-generalized and may not need the except parameter.

pgdog/src/frontend/router/parser/cache/cache_impl.rs
```diff
@@ -118,6 +119,27 @@ impl Cache {
             }
         }
 
+        if comment::has_comments(query.query(), ctx.sharding_schema.query_parser_engine)? {
```
dev-lew: The check is here because we might as well short circuit if there are no comments. There is no need to tokenize the query and iterate through it again for no reason.

pgdog/src/frontend/router/parser/comment.rs
```diff
@@ -86,6 +87,76 @@ pub fn comment(
...
+    .map_err(Error::PgQuery)?;
+
+    let mut cursor = 0;
+    let mut out = String::with_capacity(query.len());
+
+    for st in &result.tokens {
+        let start = st.start as usize;
+        let end = st.end as usize;
+
+        out.push_str(&query[cursor..start]);
```
dev-lew, line 118: The code involving cursor keeps non token characters (between and at the end of all tokens) like spaces to preserve the original query. We don't want to do anything extra like normalize it or something.

pgdog/src/frontend/router/parser/comment.rs
```diff
@@ -86,6 +87,76 @@ pub fn comment(
     Ok((None, role))
 }
 
+pub fn has_comments(query: &str, engine: QueryParserEngine) -> Result<bool, Error> {
+    let result = match engine {
+        QueryParserEngine::PgQueryProtobuf => scan(query),
```
levkk: One small concern I have here is we are calling `scan` multiple times. It's actually a pretty expensive call, because we have to go all the way to the Postgres parser, so we try to minimize it as much as possible (i.e. that's why we have the cache). It would be great if we could do this all inside `fn comment`, somehow, calling `scan` only once. Maybe we need to move comment detection a little higher in the call stack?

dev-lew: Worst case we call scan 3 times, one for has_comments, one for remove_comments, and one for comment if we get to with_context call and create an Ast. A simple solution is to make has_comments return the `Option<Vec<ScanToken>>` so that remove_comments can use it, that would make the worst case 2 scan calls.

If we wanted to call scan a single time in comment, maybe we have a hashmap that maps query to tokens and populate it on every call to comment. Then, has_comments and remove_comments can operate on the tokens in the hashmap for a given query. If its not in the hashmap, we can actually skip the entire block and go right to the with_context call. This should reduce the worst case to a single scan.

The memory usage of that hashmap is unclear to me though, its probably unbounded as long as pgdog is running...

levkk: I think an easier way to do this is to move comment detection up here: https://github.com/pgdogdev/pgdog/blob/73819ecca2557fa4b9643904502c8c3256996372/pgdog/src/frontend/router/parser/cache/cache_impl.rs#L122

`comment` will return a `String` without the comments, iff the string had comments. Then you can check the cache again, and if it's a hit, return the cached `Ast`.

Else,  pass the comment in as an argument for `Ast` struct creation, so you don't have to `scan` twice.

pgdog/src/frontend/router/parser/comment.rs
```diff
@@ -181,11 +251,41 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_remove_comments_with_exceptions() {
+        let query = "SELECT /* comment */ * FROM table /* pgdog_shard: 4 comment */ WHERE id = 1";
+        let tokens = scan(query).unwrap().tokens;
+
+        let result = remove_comments(query, &tokens, Some(&[&SHARD])).unwrap();
+
+        assert_eq!(result, "pgdog_shard: 4SELECT  * FROM table  WHERE id = 1");
```
levkk: This is not correct, it should be:

```
/* pgdog_shard: 4 */ SELECT * FROM table WHERE id = 1;
```

pgdog/src/frontend/router/parser/comment.rs
```diff
@@ -181,11 +251,41 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_remove_comments_multiple_metadata() {
+        let query = "SELECT 1 /* pgdog_shard: 4 */ + 2 /* pgdog_role: primary */;";
+        let tokens = scan(query).unwrap().tokens;
+
+        let result = remove_comments(query, &tokens, Some(&[&ROLE, &SHARD])).unwrap();
+
+        assert_eq!(result, "pgdog_role: primarypgdog_shard: 4SELECT 1  + 2 ;");
```
levkk: Same as above. This query isn't valid SQL.

PR #780, opened 2026-02-19
Fix passthrough auth for users without a password 

### LINKED ISSUE #373: [Auth] Passthrough auth for users without a password is broken
### Description

Passthrough auth for users  in `users.toml` that don't have a password triggers weird behavior:

1. Connection pool is created and starts sending health checks to database
2. Health checks fail since PgDog can't connect to Postgres without a password

### Potential fix

1. Pause pool on creation and resume it when a user connects and provides a password

### Potential edge cases

1. When Postgres auth is set to  `trust`, we don't need  a password for the user,  so it's valid to not provide  one and the pool  should still be created. Maybe we don't resume it until the user connects anyway?

### Reproduce

**pgdog.toml**

```toml
[general]
passthrough_auth = "enabled_plain"

[[databases]]
name = "pgdog"
host = "127.0.0.1"
```

**users.toml**

```toml
[[users]]
name =  "pgdog"
database = "pgdog"
# Note: no password set.
# It can also be:
# password = "" # Empty string
# We don't differentiate between the two which is bad enough.
```

Start PgDog and notice the pool being banned on startup.

howenyap:
Closes #373 

# Logic change
- when building a new `Cluster` with `new_pool`, pause the cluster if passthrough auth is enabled and password is empty
- on login, resume cluster if passthrough password exists

# Testcases
| Test | is AuthType::Trust? | is password empty?  | cluster state |
|---|---:|---:|---:|
| `test_passthrough_empty_password_starts_paused` | `false` | `true` | `paused` |
| `test_passthrough_user_with_password_unpaused` | `false` | `false` | `not paused` |
| `test_passthrough_empty_password_trust_starts_unpaused` | `true` | `true` | `not paused` |

- `test_replace_empty_password_cluster_with_passthrough_password`
  -  start with empty password, cluster should be paused
  - add new connection pool with passthrough password, cluster should resume

- `test_pause_resume_toggles_all_pools`: `cluster.pause()` should pause every pool, while `cluster.resume()` should resume every pool`

# `cargo nextest run --profile dev`
image

# `bash integration/go/run.sh`
image

### DIFF
--- a/pgdog/src/backend/pool/connection/mirror/mod.rs
+++ b/pgdog/src/backend/pool/connection/mirror/mod.rs
@@ -251,14 +251,14 @@ mod test {
                 3,
                 "mirror buffer should have 3 requests"
             );
-            sleep(Duration::from_millis(50)).await;
+            sleep(Duration::from_millis(100)).await;
             // Nothing happens until we flush.
             assert!(
                 conn.execute("DROP TABLE pgdog.test_mirror").await.is_err(),
                 "table pgdog.test_mirror shouldn't exist yet"
             );
             assert!(mirror.flush(), "mirror didn't flush");
-            sleep(Duration::from_millis(50)).await;
+            sleep(Duration::from_millis(100)).await;
             assert!(
                 conn.execute("DROP TABLE pgdog.test_mirror").await.is_ok(),
                 "pgdog.test_mirror should exist"


howenyap:
thanks for reviewing @levkk ! made changes as requested

levkk:
Sorry for the delay here. I'm currently OOO, I will review this early next week.

PR #781, opened 2026-02-19
user pool_size takes precedence over database pool_size.

mijoharas:
User is more specific, and you have multiple users per database, so it's kinda surprising that the database pool size takes precedence.

Fix that.

This one caught us out, I'm assuming it's not intentional? if so, it's a bit confusing and maybe the docs should outline that? (I still think it should be changed though.)

### DIFF
--- a/pgdog/src/backend/pool/config.rs
+++ b/pgdog/src/backend/pool/config.rs
@@ -160,6 +160,7 @@ impl Default for Config {
 #[cfg(test)]
 mod test {
     use super::*;
+    use pgdog_config::PoolerMode;
 
     fn create_database(role: Role) -> Database {
         Database {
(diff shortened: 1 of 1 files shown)


levkk:
It was not intentional, and kind of became the norm since changing this is breaking. And that's fine. I'll put a note in the release notes, because this is the right way to handle it.

Could you quickly take a look at any other settings we set there and make sure user overrides database there as well? Then we can merge. I can update the docs (or if you're willing to make a PR there as well).

Thank you!

pgdog/src/backend/pool/config.rs
```diff
@@ -186,6 +182,23 @@ mod test {
         assert!(config.role_detection);
     }
 
+    #[test]
+    fn test_user_takes_precedence_over_database() {
```
levkk, line 187: Not a terrible idea to do the other configs too. Only the paranoid survive, something something

levkk:
Okay let's see if the tests catch anything. If we're all green, we know are tests don't cover enough! 😆

pgdog/src/backend/pool/config.rs
mijoharas: good call, let me do that.

mijoharas: ok had a meeting but just got back to it. Have pushed the extra test commit.

Issue #804, opened 2026-03-03
Make read_write_split a per database configuration option.


mijoharas:
Hey Lev,

We're doing some testing, and shifting things around. We have a db that is `primary_only` (just has one primary), one that is `replica_only` (just has one replica) and one that has both primaries, and replicas.

In our testing, we want to use the last option, but we only want queries to go to the replica. Now, because `read_write_split` is a general configuration option, we can't have it use `exclude_primary` (that makes our `primary_only` break.

I've got two ideas, and actually, the second one seems easier to implement:
* make `read_write_split` a database configuration option, so we can override it to `exclude_primary` just for the last db.
* create a new `read_write_split` option that's something like `exclude_primary_if_replicas_available` that will not exclude primary, if there is only one primary for the db.

Let me know what you think makes sense?

mijoharas:
cool, yeah that makes sense to me. 👌 Will get a PR in.

PR #808, opened 2026-03-04
Fix exclude primary with no replicas

mijoharas:
As above, fixes https://github.com/pgdogdev/pgdog/issues/804


pgdog/src/backend/pool/lb/test.rs
```diff
@@ -393,6 +393,42 @@ async fn test_read_write_split_include_primary() {
...
+        LoadBalancingStrategy::Random,
+        ReadWriteSplit::ExcludePrimary,
+    );
+    replicas.launch();
+
+    let request = Request::default();
+
+    // Try getting connections multiple times and verify both primary and replica can be used
+    let mut used_pool_ids = HashSet::new();
+    for _ in 0..20 {
```
mijoharas: let's get rid of this, copy/paste job. I'll simplify in a second. (can just do round robin and go twice).

PR #809, opened 2026-03-04
Use OUT_DIR for autogenerated files.

mijoharas:
addresses https://github.com/pgdogdev/pgdog/pull/808#issuecomment-3997647499


levkk (approved):
Nice!

Issue #709, opened 2026-01-12
[Query engine] Identify which tables are on which shard for \dt


levkk:
### Description

Intercept the queries sent by psql `\dt` and add a `shard` column to indicate which database they are on.

levkk:
After some thinking, I think this may not be needed, at least not yet. We'll add more advanced schema management, but probably won't do it by hacking `psql`.

PR #812, opened 2026-03-05
feat: Extend `\dt` psql command output with shard metadata (#709)

Adi-Goll:
This PR intercepts the psql `\dt` command, and add a `Shard` column to the output.

First a flag to `Route` that indicates if `\dt` is being executed so the Shard column is conditionally applied. 
A `shard_map` HashMap is also added to `Route` that stores table names with their corresponding shard.

This PR also introduces a `forward_with_shard()` function in `backend/pool/connection/binding.rs` that
exposes the shard_map property which provides the correct shards be streamed from the query engine.

Finally engine logic was added to populate the new column correctly and handle tables sharded
across multiple databases. The following output uses this config file for testing:

<details>
<summary><strong>pgdog.toml</strong></summary>

```toml
[general]
openmetrics_port = 9090
port = 6432
host = "127.0.0.1"
system_catalogs = "sharded"

[[databases]]
name = "mycluster"
database_name = "shard_db_0"
host = "127.0.0.1"
port = 5432
shard = 0

[[databases]]
name = "mycluster"
database_name = "shard_db_1"
host = "127.0.0.1"
port = 5432
shard = 1

[[databases]]
name = "mycluster"
database_name = "shard_db_2"
host = "127.0.0.1"
port = 5432
shard = 2

[[sharded_tables]]
name = "users"
database = "mycluster"
column = "user_id"
data_type = "bigint"
``` 

</details>
Before this PR: 

```
           List of tables
 Schema |   Name    | Type  | Owner
--------+-----------+-------+--------
 public | only_on_0 | table | ubuntu
 public | only_on_1 | table | ubuntu
 public | only_on_2 | table | ubuntu
 public | users     | table | ubuntu
 public | users     | table | ubuntu
 public | users     | table | ubuntu
```

Note, the same users table was created on multiple databases, so some deduplication logic was added during streaming, done through an added HashSet property in the QueryEngineContext.
After this PR:

```
Ex. output:
                List of tables
 Schema |   Name    | Type  | Owner  |  Shard
--------+-----------+-------+--------+---------
 public | only_on_0 | table | ubuntu | 0
 public | only_on_1 | table | ubuntu | 1
 public | only_on_2 | table | ubuntu | 2
 public | users     | table | ubuntu | 0, 1, 2
```

### DIFF
--- a/pgdog/src/frontend/client/query_engine/mod.rs
+++ b/pgdog/src/frontend/client/query_engine/mod.rs
@@ -10,6 +10,7 @@ use crate::{
     state::State,
 };
 
+use std::collections::HashSet;
 use tracing::debug;
 
 pub mod connect;
@@ -78,6 +79,7 @@ pub struct QueryEngine {
     notify_buffer: NotifyBuffer,
     pending_explain: Option<ExplainResponseState>,
     hooks: QueryEngineHooks,
+    seen_tables: HashSet<String>,
 }
 
 impl QueryEngine {
@@ -105,6 +107,7 @@ impl QueryEngine {
             pending_explain: None,
             begin_stmt: None,
             router: Router::default(),
+            seen_tables: HashSet::new(),
         })
     }
 
--- a/pgdog/src/frontend/client/query_engine/query.rs
+++ b/pgdog/src/frontend/client/query_engine/query.rs
@@ -7,7 +7,7 @@ use crate::{
         router::parser::{explain_trace::ExplainTrace, rewrite::statement::plan::RewriteResult},
     },
     net::{
-        DataRow, FromBytes, Message, Protocol, ProtocolMessage, Query, ReadyForQuery,
+        DataRow, Field, FromBytes, Message, Protocol, ProtocolMessage, Query, ReadyForQuery,
         RowDescription, ToBytes, TransactionState,
     },
     state::State,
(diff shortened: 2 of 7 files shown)


pgdog/src/frontend/client/query_engine/query.rs
```diff
@@ -152,6 +164,40 @@ impl QueryEngine {
...
+                if let Some(map) = shard_map {
+
+                    if self.seen_tables.contains(table_lookup) {
+                        return Ok(())
+                    }
+
+                    self.seen_tables.insert(table_lookup.to_string());
+
+                    let mut new_col = String::new();
+                    for (i, val) in map[table_lookup].iter().enumerate() {
```
Adi-Goll, line 186: I was thinking, maybe it would be better to have a more unique key for the HashMap, maybe instead of just the table name I should use the schema + table name? Thoughts?

pgdog/src/frontend/client/query_engine/mod.rs
```diff
@@ -105,6 +107,7 @@ impl QueryEngine {
             pending_explain: None,
             begin_stmt: None,
             router: Router::default(),
+            seen_tables: HashSet::new(),
```
Adi-Goll, line 110: Not sure if it's reasonable to add this for \dt because the original command doesn't do any aggregating or deduplication handling

Adi-Goll:
Work on this PR discontinued for now because it may be preferable to have more advanced schema management, achieved through simpler means as opposed to modifying the psql command. 

https://github.com/pgdogdev/pgdog/issues/709

PR #843, opened 2026-03-23
chore: fix clippy issues

meskill:
Clippy fixes for `--all-targets` and some of the fixes for the enterprise version


pgdog/src/backend/replication/logical/publisher/table.rs
```diff
@@ -237,11 +237,19 @@ impl Table {
             }
         }
 
-        copy_sub.copy_done().await?;
+        if let Err(err) = copy_sub.copy_done().await {
+            tracker.error(&err);
```
levkk: We track this in the caller of this function already?

meskill: yes, removed this addition.
I got confused with the enterpise version of this file, I'll remove this tracking in the ent version then

PR #858, opened 2026-03-30
feat: add 53-bit unique ID generator & database-generated sharded sequences

levkk:
### Smaller unique ID

_This feature is still experimental and subject to change_.

Add support for 53-bit unique ID. This makes them smaller and "javascript-safe" in case identifiers are exposed as integers to the app API.

This is configurable thourough `pgdog.toml`:

```toml
[general]
unique_id_function = "compact" # or "standard", which is default
```

Trade-offs: it only allows 64 pgdog nodes and 64,000 IDs generated _per second_ per node. This is considerably lower than the standard 64-bit unique ID. It should be used for omnisharded tables only with a low write frequency.

### Database-generated sharded sequences

Add the ability to generate cross-shard unique IDs using sequences. This produces smaller starting integers, but can only work with direct-to-shard `INSERT`s. For omnisharded tables, the unique ID function should continue to be used.

This is configurable through `pgdog.toml`:

```toml
[rewrite]
primary_key = "rewrite_omni" # Will only inject Unique ID to omnisharded table inserts, leaving others to be
                                                # generated by the database instead.
```

The sharded sequence is installed automatically on all tables with a `BIGINT` primary key when running `pgdog setup` or `SETUP SCHEMA` via the admin DB.

### Identity columns

`GENERATED ... WITH IDENTITY` is now removed entirely during schema sync. This constraint blocks our implementation of replication and doesn't work with sharded databases, incl. with our unique ID generator because it prevents the column value from being inserted by the query.

### Smaller features

- Add `RESET PREPARED` admin command which evicts all unused prepared statements from the global cache
- log any notice/warning received from Postgres when executing queries via `execute_checked`

### Bug fixes

- `unique_id_min` setting was being ignored by the unique ID generator. This didn't affect any deployments because unique IDs start very large already.
- Admin command `SETUP SCHEMA` was applied to all database/user pairs; it's now applied to the `schema_owner` only

### DIFF
--- a/pgdog/src/backend/schema/mod.rs
+++ b/pgdog/src/backend/schema/mod.rs
@@ -505,13 +505,6 @@ mod test {
     #[tokio::test]
     async fn test_identity_column() {
         let mut server = test_server().await;
-        let server_version_num = server
-            .fetch_all::<i32>("SELECT current_setting('server_version_num')::int")
-            .await
-            .unwrap()
-            .into_iter()
-            .next()
-            .unwrap();
         // Drop and recreate the schema so the test is repeatable.
         server
             .execute_checked("DROP SCHEMA IF EXISTS pgdog_schema_test CASCADE")
@@ -543,12 +536,8 @@ mod test {
                 .unwrap();
             assert_eq!(
                 identity.first().map(String::as_str),
-                Some(if server_version_num >= 170000 {
-                    "NO"
-                } else {
-                    "YES"
-                }),
-                "{parent}.id identity state should match install_server behavior for PostgreSQL {server_version_num}",
+                Some("NO"),
+                "{parent}.id should no longer be identity after install_server",
             );
 
             let default: Vec<String> = server
--- a/pgdog/src/backend/schema/setup.sql
+++ b/pgdog/src/backend/schema/setup.sql
@@ -273,9 +273,8 @@ BEGIN
     -- Make the sequence owned by the shadow table.
     EXECUTE format('ALTER SEQUENCE pgdog_shadow."%s" OWNED BY pgdog_shadow.%s.%s', shadow_seq_name, shadow_table_name, column_name);
 
-    -- Only PostgreSQL 17+ should have identity removed before installing
-    -- the custom default.
-    IF current_setting('server_version_num')::INTEGER >= 170000 AND EXISTS (
+    -- Drop identity constraint if one exists, since we're replacing it with a custom default.
+    IF EXISTS (
         SELECT 1
         FROM information_schema.columns c
         WHERE c.table_schema = install_sharded_sequence.schema_name


pgdog/src/unique_id.rs
```diff
@@ -122,25 +171,42 @@ impl UniqueId {
             return Err(Error::NodeIdTooLarge(node_id));
         }
 
+        // Compact (JS-safe) IDs only have 6 node bits.
+        if node_id > COMPACT_MAX_NODE_ID {
+            return Err(Error::CompactNodeIdTooLarge(node_id));
+        }
+
```
meskill: that's probably should be checked only if the function is compact. For standard it could cause unexpected error

pgdog/src/unique_id.rs
```diff
@@ -122,25 +171,42 @@ impl UniqueId {
...
         if min_id > MAX_OFFSET {
             return Err(Error::OffsetTooLarge(min_id));
         }
```
meskill: check is done only for `MAX_OFFSET` which is based on the standard generator, but it really depends on the actual layout.

meskill: might worth to move the checks into the id_type or state itself to avoid adding more conditions here and be ready for other extensions

pgdog/src/unique_id.rs
```diff
@@ -31,6 +32,14 @@ const TIMESTAMP_SHIFT: u8 = (SEQUENCE_BITS + NODE_BITS) as u8; // 22
 const MAX_OFFSET: u64 = i64::MAX as u64
```
meskill, line 32: btw, MAX_OFFSET if effectively 0. That makes `config.general.unique_id_function` unusable except for default 0 value. It seems unit tests are bypasses this since the offset tests use next_id directly that has no validation for `MAX_OFFSET`

pgdog/src/unique_id.rs
levkk: Good catch!

pgdog/src/unique_id.rs
levkk: I'm not sure what you mean. Could you elaborate? I think we test the min ID in unit tests by passing in as an argument.

meskill: here 
image

if we do the math the value is 0. And the tests are not catching this because the tests for `id_offset` are only calling `next_id` function that has no validation.

pgdog/src/unique_id.rs
meskill, line 147: a sidenote:
I wonder if we really need the sleep here to correct the clock drift. It seems like it's used to make sure the call to SystemTime::now() fits our expectation for monotonically increasing timestamp. But it's not strictly necessary for this calculation since it could be emulated by `now_ms().max(self.last_timestamp_ms)`. Yes, it won't be not exactly the system time, but we fulfill the requirements. 
Just because the sleep is not reliable and scheduler could skew the time in this case

pgdog/src/unique_id.rs
levkk: Nice catch. Let me see what we can do here...

pgdog/src/unique_id.rs
levkk: I think we're just trying to guarantee that the next call to `now()` returns the subsequent second or greater. This is used when the internal sequence is exhausted and we have to wait until next clock tick to guarantee that the sequence is monotonically increasing.

PR #876, opened 2026-04-07
fix: ProtocolOutOfSync error

meskill:
More details are in the https://github.com/pgdogdev/pgdog/pull/876/changes#diff-fa25f12c8c64f8778cd0849682b0310883829c9eed403e4cde728e08ff980f29

The pr fixes couple of issues that could potentially cause the unexpected ProtocolOutOfSync errors, but only the issue 1 looks like highly possible in the real environments. 

# Issue 1

The issue 1 is reproducible when `prepared_statements = "full"` and the error happens during `PREPARE` part of the query execution. After the `PREPARE` fail the previous code was dropping all the remaining expected codes from the stream and stopped. But the backend will actually return the errors for `EXECUTE` anyway and this error will appear as the response to some another query and maybe even for another client. 

The nature of the issue makes it harder to trace since the failed state of protocol is passed back to pool where another connection picks it up and breaks.

The integration test `integration/prepared_statements_full/protocol_out_of_sync_spec.rb` reproduces the issue with different configuration and it shows the errors in the https://github.com/pgdogdev/pgdog/actions/runs/24104237887/job/70323280283 

I'm not sure why the ci doesn't show all logs, but the server logs locally are 
image

The fix is mostly inside https://github.com/pgdogdev/pgdog/pull/876/changes#diff-8495b645d11def521aee36a05448da7fc100d632e6ad513d04fb7a47512bd08d

## Other

The other issues not that important, but I've tried to fix them anyway just to be sure. Also extend logs in case we'll still face the issue in the future.


meskill:
moved to https://github.com/pgdogdev/pgdog/pull/913

PR #886, opened 2026-04-10
fix: COPY sent via extended protocol with Sync can block server re-use in transaction mode

levkk:
The following sequence of messages can make a server be stuck in "unfinished state", blocking its re-use by other clients:

```
Parse("COPY t FROM STDIN")
Bind
Execute
Sync # This message is ignored by Postgres but not by pgdog
CopyData
[...]
CopyDone
Sync
```

The `Sync` sent after `Execute` is not necessary, but we record it and expect a `ReadyForQuery` from the server which only arrives after the second `Sync` is sent. An additional `ReadyForQuery` message remains in our internal state, and `server.done()` never returns true, causing the server to be tied to the client until it disconnects.

Related #885

### DIFF
--- a/pgdog/src/backend/protocol/state.rs
+++ b/pgdog/src/backend/protocol/state.rs
@@ -163,6 +163,12 @@ impl ProtocolState {
             ExecutionCode::ReadyForQuery => {
                 self.out_of_sync = false;
             }
+            ExecutionCode::Copy => {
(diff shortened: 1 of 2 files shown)


NikolayS:
is it a good idea to have a couple of tests where we know it's a problem – eg for tokio-postgres or postgres.js?

PR #887, opened 2026-04-10
fix: ProtocolOutOfSync error

levkk:
This was caused by the following conditions:

1. `query_parser = "on"`, so `SET` statements are intercepted
2. `ruby-pg` driver initialized with `application_name: ''` so it wouldn't run `SET application_name TO x`
3. A query executed via extended protocol
4. A bad parameter set manually via `SET`, e.g. `SET lock_timeout to sdf`
5. Any query executed afterwards

This was caused by us not resetting extended protocol state in our protocol state machine and allowing invalid `SET` statements to be executed in a batch.

### DIFF
--- a/integration/ruby/pg_spec.rb
+++ b/integration/ruby/pg_spec.rb
@@ -3,12 +3,21 @@
 require_relative 'rspec_helper'
 
 def connect(dbname = 'pgdog', user = 'pgdog')
-  PG.connect(dbname: dbname, user: user, password: 'pgdog', port: 6432, host: '127.0.0.1')
+  PG.connect(dbname: dbname, user: user, password: 'pgdog', port: 6432, host: '127.0.0.1', application_name: '')
 end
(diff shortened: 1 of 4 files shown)


integration/ruby/pg_spec.rb
```diff
@@ -3,14 +3,23 @@
...
 describe 'pg' do
   after do
     ensure_done
   end
 
+  it 'out of sync' do
+    conn = connect 'pgdog', 'pgdog'
+    conn.exec_params 'SELECT $1', [1]
+    conn.exec "SELECT 1"
+    conn.exec "SET lock_timeout TO sdfs"
```
frodsan, line 18: Any ideas why the error is raised later and not here? 🤔

levkk: We intercept SET statements and execute them only when a real query is sent.

PR #904, opened 2026-04-15
Implement text array format

jaggederest:
This adds roundtrip support for text and binary arrays containing a few sub-data-type formats, reusing the parser for other data types recursively inside the array parser. Has some SQL tests, tried to keep it relatively self-contained


levkk (approved):
very cool

PR #906, opened 2026-04-16
feat: server_auth with Microsoft Workload Identity 

MagicAbdel:
This PR adds support for Microsoft Workload Identity for server authentication on a per-user basis, following the existing pattern used for `rds_iam`.

**Changes**:
-  Implemented token acquisition logic for Azure Entra ID.
-  Ensured parity with the rds_iam implementation for 

This is my first contribution in Rust! 🦀
I’ve done my best to follow the existing patterns in the codebase, but I'm still learning the nuances of the language. Please be indulgent. If there are more idiomatic ways to handle the logic, async calls, or error handling, I am very happy to make changes based on your feedback!

Note: I also created a PR on the helm chart ([here](https://github.com/pgdogdev/helm/pull/85)) to allow  users to add `azure.workload.identity/use: "true"` label to the pod for injecting workload Identity.

Here are some logs I tested this solution in AKS with Workload Identity directly injected into the pod:
image

### DIFF
--- a/pgdog/src/backend/schema/sync/pg_dump.rs
+++ b/pgdog/src/backend/schema/sync/pg_dump.rs
@@ -1183,6 +1183,21 @@ mod test {
         assert_eq!(sslmode, Some(OsStr::new("require")));
     }
 
+        #[test]
+    fn test_build_pg_dump_command_sets_tls_for_azure_workload_identity() {
+        let mut addr = backend::pool::Address::new_test();
+        addr.server_auth = ServerAuth::AzureWorkloadIdentity;
+        let command = build_pg_dump_command("pg_dump", &addr, "token");
+
+        let sslmode = command
+            .as_std()
+            .get_envs()
+            .find(|(key, _)| *key == OsStr::new("PGSSLMODE"))
+            .and_then(|(_, value)| value);
+
+        assert_eq!(sslmode, Some(OsStr::new("require")));
+    }
+
     #[test]
     fn test_specific_dump() {
         let dump = r#"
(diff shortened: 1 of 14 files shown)


levkk (approved):
💯 👍 🙏 🚀

levkk:
`rand` has been an issue for a while. Every time a new crate is added that uses a different version than what we have, we get this conflict. Let me see if I can resolve it.

levkk:
Apply this diff when you get a second:

```diff
diff --git a/pgdog-config/src/util.rs b/pgdog-config/src/util.rs
index 64c10341..6e39d55d 100644
--- a/pgdog-config/src/util.rs
+++ b/pgdog-config/src/util.rs
@@ -1,6 +1,6 @@
 use std::time::Duration;
 
-use rand::{distr::Alphanumeric, Rng};
+use rand::{distr::Alphanumeric, RngExt};
 
 pub fn human_duration_optional(duration: Option<Duration>) -> String {
     if let Some(duration) = duration {
```

MagicAbdel:
Applied the diff and verified with cargo check. The rand conflict seems resolved now. Thanks for the help with that

pgdog/src/backend/schema/sync/pg_dump.rs
```diff
@@ -1183,6 +1183,21 @@ mod test {
         assert_eq!(sslmode, Some(OsStr::new("require")));
     }
 
+        #[test]
+    fn test_build_pg_dump_command_sets_tls_for_azure_workload_identity() {
```
levkk, line 1187: Make sure to run `cargo fmt` locally before pushing changes.

MagicAbdel: Apologies for that! It looks like my IDE didn't do the auto-format on save

levkk:
You should be able to run tests locally:

1. `bash integration/setup.sh`
2. `cd pgdog && cargo nextest run --test-threads=1 <name of your test function>`

levkk:
Kind of a bad day for our CI. I'll take a look at this later tonight / tomorrow to figure out what's going on. The code looks good, just getting the tests to pass now.

PR #913, opened 2026-04-19
Fix possible ProtocolOutOfSync when using prepared statements

meskill:
This is continue of https://github.com/pgdogdev/pgdog/pull/876

While one of the cases of unexpected ProtocolOutOfSync errors was fixed in the https://github.com/pgdogdev/pgdog/pull/887 the issue can still appear when using prepared_statements = "full"

This pr focuses on the refactoring the `pgdog/src/backend/protocol/state.rs` to provide fix for this case and also other possible issues that could also lead to issues in the future and therefore make the code less error-prone. Although right now they don't impose significant threat since it's not reproducible by the current pgdog architecture for processing commands. But I do think it's useful to clean this up to prevent any degradation in the future.

The details are described in the doc file https://github.com/pgdogdev/pgdog/pull/913/changes#diff-fa25f12c8c64f8778cd0849682b0310883829c9eed403e4cde728e08ff980f29 and tests for different cases were added.


meskill:
updated the doc to simplify it a little. The issue 1 in the doc should describe the main issue and the fix is mostly contained inside state.rs along the inline comments to explain the logic. The remaining issues and changes are mostly cosmetic and to simplify state.rs at some degree

PR #916, opened 2026-04-20
fix(copy_data): add retries for copy_data command

### LINKED ISSUE #894: [Resharding] COPY_DATA: reliability of the data sync stage
**PgDog version**
`v0.1.36` `veb5ab65`

**Description**
Not all tables were copied:

```
sudo docker logs pgdog-proxy 2>&1 | grep "data sync" | grep -c started
276
sudo docker logs pgdog-proxy 2>&1 | grep "data sync" | grep -c finished
63
```

The issue occurs because the target shard (port 5441) was restarted during the table copy process and became temporarily unavailable. Instead of retrying the table copy, PgDog continues processing the list of tables without waiting for the database to become available.

I also think we could implement a retry mechanism: if table copying is interrupted, the target shards could be truncated and the procedure retried. This would avoid having to start over and repeatedly copy large of data.

**Logs**
-  [data_sync_error.log](https://github.com/user-attachments/files/26705085/data_sync_error.log)

**Configuration**
.

### LINKED ISSUE #897: [Resharding] Retry COPY on transient errors
### Description

Retry parallel table copy if an error that's easily recoverable happens:

- can't connect to server (temporarily down, network issue, etc.)
- copy interrupted due to network error (server restarted mid copy); this is safe because copies are transactional.

The latter needs to handle two situations:

1. Destination shard is down: just retry
2. Origin is down: need to re-start the transaction and re-create the temporary replication slot

meskill:
fixes #897


pgdog/src/backend/replication/logical/publisher/parallel_sync.rs
```diff
@@ -106,22 +146,19 @@ impl ParallelSyncManager {
...
-        let replica = loop {
-            if let Some(replica) = replicas_iter.next() {
-                break replica;
-            } else {
-                replicas_iter = self.replicas.iter();
-            }
-        };
+        // cycle() is the idiomatic "rewind": it restarts the iterator from the
+        // beginning once exhausted, giving round-robin distribution across replicas.
+        let mut replicas_iter = self.replicas.iter().cycle();
```
meskill, line 152: @levkk could you please check this. Did I get the idea right about the original intention?

pgdog/src/backend/pool/cluster.rs
```diff
@@ -553,6 +563,16 @@ impl Cluster {
         self.resharding_parallel_copies
     }
 
+    /// Maximum retries for a per-table copy during resharding.
+    pub fn resharding_copy_retry_max_attempts(&self) -> usize {
+        self.resharding_copy_retry_max_attempts
+    }
+
+    /// Base delay between table copy retry attempts. Doubles each attempt, capped at 32×.
+    pub fn resharding_copy_retry_min_delay(&self) -> std::time::Duration {
```
levkk: nit: make it a `Duration` on creation in `Cluster::new` or in `ClusterConfig` above. You can then pass a reference the the value stored in `self`, e.g.:

```rust
pub(crate) fn resharding_copy_retry_min_delay(&self) -> &Duration {
    &self.resharding_copy_retry_min_delay
}
```

pgdog/src/backend/replication/logical/publisher/parallel_sync.rs
```diff
@@ -40,36 +40,71 @@ impl ParallelSync {
...
+                    warn!(
+                        "data sync for \"{}\".\"{}\" failed (attempt {}/{}): {err}, retrying after {}ms...",
+                        self.table.table.schema,
+                        self.table.table.name,
+                        attempt,
+                        max_retries,
+                        backoff.as_millis(),
+                    );
+
+                    tokio::time::sleep(backoff).await;
```
levkk: nit:

```rust
use tokio::time::sleep;
```

```suggestion
                    sleep(backoff).await;
```

pgdog/src/backend/replication/logical/publisher/parallel_sync.rs
```diff
@@ -40,36 +40,71 @@ impl ParallelSync {
...
+
+                    if let Err(trunc_err) = self.table.truncate_destination(&self.dest).await {
```
levkk: I don't think we should do this for two reasons:

1. `COPY` is atomic and transactional: if it fails, none of the rows will be saved in the table; it will be empty when we retry
2. `TRUNCATE` is a scary command to run for now. We should add a bunch more tests and conditions that prevent this from _accidentally_ being called on the source DB. For now, let's have the user truncate manually if this retry logic fails for some reason.

There is a chance for a race condition where the table copy completes and we get an error somewhere below, e.g., while running `COMMIT`, but the odds of that are slim. We should definitely account for this (and truncate), but after we implement a few "this is definitely the destination" checks. I'll write up a separate issue for this.

meskill: yes, dropped the truncate, left the comment for future

pgdog/src/backend/replication/logical/subscriber/copy.rs
```diff
@@ -174,12 +177,19 @@ impl CopySubscriber {
     }
 
     async fn flush(&mut self) -> Result<(usize, usize), Error> {
```
levkk, line 176: `send_one` is actually not free! The `ParallelConnection` gave us I think a 30-40% copy speed boost. Have you benchmarked this? Also curious what made you refactor this one.

meskill: moved to separate pr to verify and test it https://github.com/pgdogdev/pgdog/pull/920

pgdog/src/backend/pool/shard/monitor.rs
```diff
@@ -42,6 +42,9 @@ impl ShardMonitor {
 
 impl ShardMonitor {
     async fn spawn(&self) {
+        if self.shard.comms().lsn_check_interval == Duration::MAX {
```
levkk: We test this code somewhere in CI right? Just double checking. This is the replica/primary promoter, so we need to be sure its tested before tweaking it. At the very least, test it locally (with `role = "auto"`) making sure it still works. I think this change is fine, although probably a no-op since the code below won't take any action if data provided by `lsn_check_interval`-controlled loop isn't provided.

pgdog/src/backend/replication/logical/status.rs
```diff
@@ -79,6 +79,19 @@ impl TableCopy {
             state.sql = Arc::new(sql.to_owned());
         }
     }
+
+    /// Reset byte and row counters before retrying a failed table copy.
+    /// Prevents accumulated counts from a discarded attempt inflating totals
+    /// and throughput calculations across retries.
+    pub(crate) fn reset(&self) {
+        if let Some(mut state) = TableCopies::get().get_mut(self) {
```
levkk, line 87: Not a bad idea to track retries, but can add that as a follow-up.

levkk:
The mirror test is flaky. The rest looks good to me!

pgdog/src/backend/pool/shard/monitor.rs
meskill: removed this

levkk (approved):
Should be good to go!

PR #920, opened 2026-04-22
refactor: drop parallel_connections wrapper

meskill:


pgdog/src/backend/replication/logical/subscriber/copy.rs
```diff
@@ -174,12 +177,19 @@ impl CopySubscriber {
...
+
         let result = self.copy.shard(&self.buffer)?;
         self.buffer.clear();
 
         let rows = result.len();
         let bytes = result.iter().map(|row| row.len()).sum::<usize>();
+        self.bytes_sharded += bytes;
 
+        // Route each row to the right shard(s). send_one is a buffered write
+        // so this loop does no I/O — no concurrency needed here.
```
levkk, line 192: That's not true. `send_one` does a buffered write _only_ if there is space in the buffer. Once that space runs out, it will flush the data to the socket.

With copy, where it writes a lot of data, this happens very frequently. This part definitely needs to be parallelized.

meskill: you are right about `send_one`.

but at the end that's the same `send_one` in the listener that will block the listener loop when the io buffer is full. So  by using `ParallelConnection` we are defining additional buffer (channel) above that. And there is third buffer above it inside the `copy_data`.
The task offloading can help there to not block reading from the source while writing to the destination, but the computation cost should be the same (basically higher due to additional machinery).
My bet that's the size of the buffer is the important part since with enough size we can achieve parallel reading and writing. Where is this buffer is defined that's not important.

So, anyway, need to test and benchmark it.

levkk: Maybe! We'll certainly find out. The tokio machinery is basically free: it's just an in-memory queue with a mutex. The mutex has no contention on it, so it's just a couple assembly instructions. Compare that to the chance of waiting on IO (multiple milliseconds vs. 50ns) and parallel connections all of a sudden provide an exponential performance improvement.

Of course in reality that's not true. Network is bottlenecked by Postgres which itself is bottlenecked by its own CPU/disk performance, so the improvement will be modest, but the goal is to _always_ saturate the network and only slow down when TCP tells us to.

PR #926, opened 2026-04-23
feat: support pg protocol version negotiation for frontend clients using v3.2

sarajmunjal:
This PR adds frontend-side PostgreSQL protocol 3.2 support to PgDog - addressing https://github.com/pgdogdev/pgdog/issues/578.

PgDog now accepts 3.2 startup requests, negotiates newer 3.x minor versions down with NegotiateProtocolVersion, reports unsupported _pq_.* startup options in that response, and returns cancel keys / BackendKeyData in the right shape for the negotiated protocol. Backend connections to PostgreSQL are unchanged.

**What changed**

- added protocol version parsing/negotiation and NegotiateProtocolVersion
- updated startup parsing for 3.2, 3.3+ -> 3.2, and _pq_.* option reporting
- updated BackendKeyData / CancelRequest handling for protocol-dependent secret lengths
- generated client cancel keys based on the negotiated frontend protocol version
- added unit and integration coverage for negotiation and extended cancel-key behavior
- added an independent psql 18 smoke test pinned to exact 3.0 and 3.2

**Compatibility**

3.0 clients continue to work
3.2 clients are accepted directly
newer 3.x minor versions negotiate down to 3.2
non-3.x major versions remain unsupported

**Verification**

Ran locally:

```sh
cargo fmt --all
cargo test -p pgdog --lib --no-run
cargo test -p rust --test mod --no-run
cargo test -p rust --test mod integration::protocol_version:: -- --nocapture
bash integration/complex/protocol_version/run.sh
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,7 +110,15 @@ jobs:
           sudo -u postgres createuser --superuser --login $USER
           sudo -u postgres createdb $USER
           bash integration/setup.sh || (sudo service postgresql restart && bash integration/setup.sh)
-          sudo apt update && sudo apt install -y python3-virtualenv mold php-cli php-pgsql
+          sudo apt update && sudo apt install -y python3-virtualenv mold php-cli php-pgsql curl ca-certificates
+          # Use a versioned PostgreSQL 18 client so the protocol-version smoke
+          # test does not depend on whatever libpq the runner image ships.
+          sudo install -d /usr/share/postgresql-common/pgdg
+          sudo curl -o /usr/share/postgresql-common/pgdg/apt.postgresql.org.asc --fail https://www.postgresql.org/media/keys/ACCC4CF8.asc
+          . /etc/os-release
+          sudo sh -c "echo 'deb [signed-by=/usr/share/postgresql-common/pgdg/apt.postgresql.org.asc] https://apt.postgresql.org/pub/repos/apt $VERSION_CODENAME-pgdg main' > /etc/apt/sources.list.d/pgdg.list"
+          sudo apt update && sudo apt install -y postgresql-client-18
+          echo "PSQL_BIN=/usr/lib/postgresql/18/bin/psql" >> "$GITHUB_ENV"
           sudo gem install bundler
           sudo apt remove -y cmake
           sudo pip3 install cmake==3.31.6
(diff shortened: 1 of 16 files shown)


.github/workflows/ci.yml
```diff
@@ -109,8 +109,23 @@ jobs:
...
           sudo -u postgres createuser --superuser --login $USER
           sudo -u postgres createdb $USER
+          sudo -u postgres psql -c 'ALTER SYSTEM SET max_connections TO 1000;'
+          sudo -u postgres psql -c 'ALTER SYSTEM SET max_prepared_transactions TO 1000;'
+          sudo -u postgres psql -c 'ALTER SYSTEM SET wal_level TO logical;'
+          sudo -u postgres psql -c 'ALTER SYSTEM SET max_worker_processes TO 64';
+          sudo -u postgres psql -c 'ALTER SYSTEM SET max_wal_senders TO 32';
+          sudo -u postgres psql -c 'ALTER SYSTEM SET max_replication_slots TO 32';
+          sudo service postgresql restart
           bash integration/setup.sh || (sudo service postgresql restart && bash integration/setup.sh)
```
levkk, line 112: Might need to rebase on `main` here. The `ALTER SYSTEM` queries are part of `setup.sh` now.

integration/rust/tests/integration/protocol_version.rs
```diff
@@ -0,0 +1,142 @@
...
+    let nul = payload
+        .iter()
+        .position(|byte| *byte == 0)
+        .expect("NUL-terminated string");
+    let value = String::from_utf8(payload.split_to(nul).to_vec()).expect("valid utf-8");
+    payload.advance(1);
+    value
+}
+
+fn parse_negotiation(message: &Message) -> (i32, Vec<String>) {
```
levkk, line 56: We implement messages in `net/messages`. You can define a new struct there and implement the conversion. I think that's generally better for maintenance. Back in pgcat days we had one function per message and it got a bit messy, especially since pgdog will probably need to send the same messages to the server soon enough.

levkk: Oh nevermind, sorry, this is a test. Ignore what I said.

levkk (approved):
Amazing, thank you so much! Could you just push the rebase fix and this is good to go!

sarajmunjal:
@levkk rebased and pushed. I think rwx needs your approval to kick off?

levkk:
Will merge once tests pass (likely this weekend, worst case Monday).

PR #932, opened 2026-04-24
Parallelize CI

### LINKED ISSUE #931: [CI] Add base images
Add base images for:

1. Package build - no apt-get install for each docker build
2. CI - no apt-get install and no server configuration - that takes upwards of 5 minutes per run

levkk:
1. Fix #931. Add base images for pgdog package, so main builds are faster.
2. Parallelize all integration tests, re-use build binaries, so they run much faster (17min -> 6min). Flaky tests don't block others from running / finishing. Flaky tests can be independently retried.


.github/workflows/fmt.yml
```diff
@@ -0,0 +1,20 @@
...
+    runs-on: blacksmith-4vcpu-ubuntu-2404
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install CI deps
+        run: bash integration/ci/install-deps.sh
+      - name: Format
+        run: cargo fmt --all -- --check
+      - name: Clippy
+        working-directory: pgdog
+        run: cargo clippy
```
meskill: we can `cargo clippy --all-targets` to check the tests as well 🧪

.github/workflows/ci.yml
```diff
@@ -7,240 +7,82 @@ on:
...
+          - { name: java,          script: integration/java/run.sh }
+          - { name: mirror,        script: integration/mirror/run.sh }
+          - { name: sql,           script: integration/sql/run.sh }
+          - { name: toxi,          script: integration/toxi/run.sh }
+          - { name: rust,          script: integration/rust/run.sh }
+          - { name: python,        script: integration/python/run.sh }
+          - { name: complex,       script: integration/complex/run.sh }
+          - { name: dry-run,       script: integration/dry_run/run.sh }
+          - { name: copy-data,     script: integration/copy_data/run.sh }
+          - { name: load-balancer, script: integration/load_balancer/run.sh }
```
meskill, line 65: we can go even further and specify only name and infer the path from it

.github/workflows/ci.yml
meskill, line 1: rename the file to integration to make it explicit?

integration/ci/cache-key.sh
```diff
@@ -0,0 +1,47 @@
...
+
+dirs=(
+    pgdog/src
+    pgdog-config/src
+    pgdog-macros/src
+    pgdog-plugin/src
+    pgdog-postgres-types/src
+    pgdog-stats/src
+    pgdog-vector/src
+)
```
meskill, line 38: I bet we can use some grep/expansion magic here. But in that case we can touch unexpected things though won't break the cache when we add something new.

.github/workflows/ci.yml
```diff
@@ -7,240 +7,82 @@ on:
...
+        if: steps.pgdog-bin.outputs.cache-hit != 'true'
+        run: cargo build --release --bin pgdog
+
   integration:
     runs-on: blacksmith-4vcpu-ubuntu-2404
+    needs: [build]
     timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
```
meskill, line 49: oh, so cool, I wanted to do this as well, but I was staring in the lcov binary dances and the build sharing there

integration/ci/cache-key.sh
levkk: Yeah I was thinking the same thing. I think this is ok for now as long as `pgdog/src` is part of it, since it's very unlikely we'll add something somewhere else without modifications to the core code.

.github/workflows/ci.yml
levkk: Or rename tests to "ci" :)

.github/workflows/ci.yml
levkk: Ended up needing some additional options for some of the tests, so might keep this for now!

PR #934, opened 2026-04-25
feat: persist 2pc transaction state to disk

### LINKED ISSUE #911: [Sharding] Persist 2pc transaction state to disk
Persist the state of 2pc transactions to disk, if possible. When pgdog is rebooted / killed, it can restore the transaction state and rollback/commit.

Make this optional (auto detect on boot, warn if can't). We would need to run it with a PV in kube.

willothy:
Adds a WAL with recovery for 2pc transaction state for crash-safety.

Closes #911


levkk:
Sick! Will review tomorrow first thing. Don't worry about the rwx failure, we removed it from `main`. Tests now only run on GH actions (might need a rebase on `main` if you want to clear that error before merging).

levkk (approved):

PR #935, opened 2026-04-26
feat: Token caching for external identity providers

magicabdel:
### Summary
Introduces an in-memory token cache shared by `azure_workload_identity` and `rds_iam` authentication backends. Tokens are now fetched once and reused until expiry, instead of being fetched on every connection.

### Motivation
Token fetching from external identity providers can be slow — Azure Workload Identity in particular was measured at ~30s per token fetch. This was directly impacting pool startup time, as each connection attempt would block waiting for a fresh token.

### Changes
- Added a shared `token_cache` module with get/set helpers keyed by host, port, and user
- Refactored `azure_workload_identity` to extract `fetch_token()` returning `(String, SystemTime)`, using the `expires_on` field from the Azure SDK response as the cache TTL
- Refactored `rds_iam` to follow the same pattern, with a fixed 15-minute TTL (RDS IAM tokens are valid for 15 minutes but the AWS SDK does not return an expiry)
- Added cache hit/miss tests for both backends

### Impact
- Pool startup is significantly faster when multiple connections share the same identity — only the first connection pays the token fetch cost
- No behavioral change for single-connection or uncached scenarios
- TTL-based expiry ensures tokens are refreshed before they become invalid


levkk:
Nice! Quick question: do you think it would be possible to run the token acquisition as a background task instead? That way, the token is always fresh when accessed for creating server connections.

pgdog/src/backend/auth/azure_workload_identity.rs
```diff
@@ -8,28 +9,46 @@ pub async fn token(addr: &Address) -> Result<String, Error> {
...
-        .map(|token| token.token.secret().to_string())
         .map_err(|error| {
             Error::AzureWorkloadIdentityToken(format!(
                 "failed to get Azure AD token for {}@{}:{}: {}",
                 addr.user, addr.host, addr.port, error
             ))
-        })
+        })?;
+
+    let expires_at = std::time::SystemTime::from(access_token.expires_on);
```
levkk: Shouldn't we make it a little bit less to account for time skew + for good luck?

magicabdel:
Sorry for the delay, I finally had some time to circle back to your comments.

That was a great suggestion. I’ve added `EXPIRY_BUFFER` to `token_cache.rs` to trigger the refresh task 45 seconds before expiry. I extracted that logic into `get_or_fetch` and applied the same pattern to `rds_iam` to account for their 15-minute validity.

Let me know if this looks good!

pgdog/src/backend/auth/token_cache.rs
```diff
@@ -0,0 +1,191 @@
...
+{
+    let key = CacheKey::from(addr);
+
+    if let GetResult::Hit(token) = get(&key) {
+        return Ok(token);
+    }
+
+    let (token, expires_at) = fetcher(addr.clone()).await?;
+    set(key, token.clone(), expires_at);
+    spawn_refresh_task(addr.clone(), expires_at, fetcher);
```
levkk: There is a race condition here where multiple connections can spawn multiple `spawn_refresh_tasks`. You want to spawn one task at startup per configured address. A good place for this would be the connection pool (`src/backend/pool/monitor.rs`).

pgdog/src/backend/auth/token_cache.rs
```diff
@@ -0,0 +1,191 @@
...
+where
+    F: Fn(Address) -> Fut + Send + Sync + Copy + 'static,
+    Fut: Future<Output = Result<(String, SystemTime), Error>> + Send + 'static,
+{
+    let key = CacheKey::from(&addr);
+    if !claim_refresh(&key) {
+        return;
+    }
+
+    tokio::spawn(async move {
```
levkk: This loop needs to check for when this address becomes invalid, e.g. when the config is reloaded. Checkout the connection pool implementation, that might be a better place implementation for long-lived loops like this one.

magicabdel:
I tried my best to address your review comments. I learned a lot going through the codebase. Happy to make changes based on feedback.

levkk (approved):
This is great, thank you!

PR #936, opened 2026-04-27
feat: wait for primary to be detected

### LINKED ISSUE #927: Intermittent error: WARN new primary chosen
Hello. Initial data:

pgdog 0.1.34 x 6 pcs
postgresql 12 to 17

Constant switching, logs show errors like:
```
WARN database role changed in shard 0
WARN new primary chosen:
```

Rare errors in application logs:
`cannot execute INSERT in a read-only transaction`

PostgreSQL clusters are different, located in different data centers, including some with no load.
I tried setting _query_parser = "on"`_

It got even worse, errors started appearing like:
`consuming input failed: no primary server closed the connection unexpectedly This probably means the server terminated abnormally before or while processing the request.`

There are no network connectivity issues. Again, the pgdog and postgresql versions vary widely; below is just a small example. What could be the cause?

Config example:

pgdog.toml
```
[general]
host = "0.0.0.0"
port = 6432
workers = 8
default_pool_size = 32
min_pool_size = 1
pooler_mode = "transaction"
load_balancing_strategy = "least_active_connections"
read_write_strategy = "conservative"
query_parser = "off"
query_log = "/var/log/pgdog/query.log"
log_connections = false
log_disconnections = false
passthrough_auth = "enabled_plain"
connect_attempts = 3
connect_attempt_delay = 1
unique_id_min = 1
openmetrics_namespace = "pgdog_"
openmetrics_port = 9090
lsn_check_interval = 5000
lsn_check_timeout = 6000
lsn_check_delay = 5 

[[databases]]
name = "apps"
database_name = "apps"
host = "10.10.200.36"
port = 5432
pool_size = 10
role = "auto"

[[databases]]
name = "apps"
database_name = "apps"
host = "10.22.1.91"
port = 5432
pool_size = 10
role = "auto"
```

users.toml
```
[[users]]
name = "apps_ro"
password = "123"
database = "apps"

[[users]]
name = "apps_rw"
password = "xxx"
database = "apps"
```

levkk:
fix #927 

```toml
[general]
read_write_strategy = "conservative_primary_required"
```

This will pause all queries until a primary is detected by `lsn_check_interval`. This will prevent this error:

```
 cannot execute INSERT in a read-only transaction
```

since PgDog won't allow any queries through until a primary is detected, so a write won't be sent to a replica.

This is not default behavior because primary databases are optional, i.e. pgdog can proxy a set of replicas only, so a primary will never be detected.

### DIFF
--- a/integration/role_detector/pgdog.toml
+++ b/integration/role_detector/pgdog.toml
@@ -2,6 +2,7 @@
 lsn_check_delay = 0
 lsn_check_interval = 5_000
 idle_timeout = 1_000
+read_write_strategy = "conservative_primary_required"
 
 [[databases]]
 name = "postgres"
--- a/pgdog-config/src/database.rs
+++ b/pgdog-config/src/database.rs
@@ -17,8 +17,21 @@ pub enum ReadWriteStrategy {
     /// Transactions are writes, standalone `SELECT` are reads (default).
     #[default]
     Conservative,
+    /// Same as conservative, except primary is required before accepting queries.
+    ConservativePrimaryRequired,
     /// Use first statement inside a transaction for determining query route.
     Aggressive,
+}
+
+impl ReadWriteStrategy {
+    pub fn is_conservative(&self) -> bool {
+        matches!(self, Self::Conservative | Self::ConservativePrimaryRequired)
+    }
+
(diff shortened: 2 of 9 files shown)


levkk:
#941

PR #940, opened 2026-04-27
fix(replication): proper handling for toasted data

meskill:
Implement handling for unchanged toasted data in replication for update events. Tests added and without fix the unchanged-toasted column is replicated as an empty string.


levkk (approved):
Really awesome find!

PR #944, opened 2026-04-28
ci: fix build fail on toml file changes

meskill:


.github/workflows/ci.yml
```diff
@@ -13,22 +13,23 @@ jobs:
...
+      # build, before rust-cache's post-job cleanup prunes workspace binaries
+      # from target/{debug,release}.
       - name: Compute cache key
         id: cache-key
         run: echo "key=pgdog-bin-${{ runner.os }}-$(bash integration/ci/cache-key.sh)" >> "$GITHUB_OUTPUT"
-      - name: Cache pgdog binaries
+      - name: Restore pgdog binaries
         id: pgdog-bin
-        uses: useblacksmith/cache@v5
+        uses: actions/cache/restore@v4
```
levkk: This one hits the azure storage API which is super slow. Their entire stack is honestly not great. Blacksmith I think uses either R2 or S3 which actually works and is super fast, so I try to stick to their actions, if possible of course. If this works, it works, but double check the latency.

.github/workflows/ci.yml
```diff
@@ -81,11 +90,11 @@ jobs:
       # cargo reconciles fingerprints during plugin builds.
       - name: Restore Rust cache for plugin builds
         if: matrix.needs_rust_cache
-        uses: useblacksmith/rust-cache@v3
+        uses: Swatinem/rust-cache@v2
```
levkk, line 93: Same story here. Blacksmith actions are much much faster typically.

.github/workflows/ci.yml
meskill: this should be not relevant anymore - https://www.blacksmith.sh/blog/cache

levkk (approved):

PR #952, opened 2026-05-01
feat(replication): add support for replica identity full

### LINKED ISSUE #930: Add support for REPLICA IDENTIFY FULL
Currently the pgdog doesn't support REPLICA IDENTIFY FULL configuration. During https://github.com/pgdogdev/pgdog/issues/914 the proper validation upfront of copy has been added to catch this early before making the copy since otherwise it was expecting to fail during the replication stage.

But, maybe we can improve this on pgdog side:
1. Modify table validation to accept IDENTIFY FULL
2. Add custom statement generation for queries to support full identity (ignore toasted, not distinct for queries etc.)
3. Verify the use cases where it can fail and add helpful logs where possible

meskill:
Fixes #930


pgdog/src/backend/replication/logical/publisher/publisher_impl.rs
```diff
@@ -328,9 +328,7 @@ impl Publisher {
             .collect();
 
         if !validation_errors.is_empty() {
-            validation_errors.sort_by_key(|e| match e {
-                TableValidationError::NoIdentityColumns(table) => table.name.clone(),
-            });
+            validation_errors.sort_by_key(|e| e.table.name.clone());
```
levkk: nit: could sort by schema and table name so errors for the same schema appear next to each other.

pgdog/src/backend/replication/logical/publisher/queries.rs
```diff
@@ -181,6 +187,60 @@ impl From<DataRow> for PublicationTableColumn {
...
+      JOIN pg_catalog.pg_attribute a
+        ON a.attrelid = i.indrelid AND a.attnum = k.attnum
+      WHERE NOT a.attnotnull
+    )
+  )
+LIMIT 1";
+
+/// Returns `true` when the named table has at least one unique index that is
+/// fully built and valid on the given server connection.
+pub async fn has_unique_index(
```
levkk: For schemas that have 1000s of tables, this could be slow. Try maybe returning a map like this:

| schema | table | unique_index |
|-|-|-|
| `public` | `users` | `t` |
| `orders` | `items` | `f` |

levkk (commented):
Still reviewing, but looks really good so far! The LB test is flaky (I need to fix it). The copy data test is also flaky I think, but I haven't looked into why yet.

pgdog/src/backend/replication/logical/subscriber/stream.rs
```diff
@@ -196,12 +207,27 @@ impl StreamSubscriber {
...
+        let omni_full: Vec<Table> = self
+            .tables
+            .values()
+            .filter(|t| {
+                t.is_identity_full() && !t.is_sharded(&self.cluster.sharding_schema().tables)
+            })
+            .cloned()
+            .collect();
+        for table in &omni_full {
+            self.validate_full_identity_omni_has_unique_index(table)
```
levkk: nit: this should be executed as one query, ideally, although probably not critical since this runs only once.

levkk (approved):
Really cool stuff!

levkk (approved):

meskill:
made a small update to remove some errors that actually shouldn't have been there for identity full and they should not block the execution flow.

levkk (approved):
Very cool!

Issue #938, opened 2026-04-27
Vault dynamic roles


larrywax:
Hello! First of all, thanks for this amazing project, it's really impressive.
I would like to adopt pgDog in my current organisation, but we currently can't because all our roles are dynamically generated by Vault, which is a strict requirement from our security team.

Would you be open to a contribution that adds support for dynamic roles? I'd be happy to work on this if it's something you'd like to see in the project
Thanks!

levkk:
Hey! Thanks!

Could you tell me a bit more about how this could work / what we need to add?

larrywax:
Here is what I’ve put together, though I'm not familiar with this codebase and probably have overlooked or oversimplified things.
Take it as a starting point to kick off the discussion.

Disclaimer: Since users are inherently dynamic when using vault's database secrets engine, I'm assuming they are being granted a static role to keep object ownership and default privileges consistent across different set of credentials with the same access level.
Example vault DDL: 
```
CREATE ROLE "{{name}}" WITH LOGIN PASSWORD '{{password}}' VALID UNTIL '{{expiration}}' IN ROLE "my-app-role"
```

### backend side: Vault credential provider per role

pgdog generates its own Vault credentials for each configured backend pool. This follows the same pattern as pgdog's existing RDS IAM and Azure Workload Identity backend auth providers.

The key difference from per-connection token generation (RDS IAM model) is **per-role credential**:

```
  Vault Credential Manager (runs in pgdog, one per role)

  my-app-role pool:
    username: v-approle-my-app-pQrStU
    password: s3cr3t-a
    TTL: XXh
  
  my-other-role pool:
    username: v-approle-my-other-AbCXyZ
    password: s3cr3t-b
    TTL: YYh
```

A background task will be responsible for refreshing credentials before expiration


### client side: `pg_auth_members` for role resolution

Like I said before, assuming vault creates PG roles like `v-approle-my-app-xYzAbC` and grants them membership in a parent role (e.g., `my-app-role`). This membership is stored in PostgreSQL's own system catalog.

```
PostgreSQL role hierarchy (created by Vault):

  my-app-role (parent, permanent)
    └── v-approle-my-app-xYzAbC  (member, ephemeral)

  my-other-app-role (parent, permanent)
    └── v-approle-my-other-aBcDeF  (member, ephemeral)

  readonly-role (parent, permanent)
    └── v-approle-ro-gHiJkL   (member, ephemeral)
```

[...]

PR #958, opened 2026-05-04
Vault backend integration

larrywax:
Draft implementation as discussed in https://github.com/pgdogdev/pgdog/issues/938

### DIFF
--- a/pgdog-config/src/users.rs
+++ b/pgdog-config/src/users.rs
@@ -88,6 +88,32 @@ impl Users {
                 );
             }
 
+            if user.server_auth == ServerAuth::Vault {
+                if user.vault_path.is_none() {
+                    warn!(
+                        r#"user "{}" (database "{}") uses Vault server auth but "vault_path" is not set"#,
+                        user.name, user.database
+                    );
+                }
+
+                if config.vault.is_none() {
+                    warn!(
+                        r#"user "{}" (database "{}") uses Vault server auth but the [vault] section is missing from pgdog.toml"#,
+                        user.name, user.database
+                    );
+                }
+
+                if let Some(percent) = user.vault_refresh_percent
+                    && (percent == 0 || percent > 100)
+                {
+                    warn!(
+                        r#"user "{}" (database "{}") has "vault_refresh_percent" of {}, expected 1-100, using default"#,
+                        user.name, user.database, percent
+                    );
+                    user.vault_refresh_percent = None;
+                }
+            }
+
             if !user.database.is_empty() && !user.databases.is_empty() {
                 warn!(
                     r#"user "{}" is configured for both "{}" and "{:?}", defaulting to "{}""#,
@@ -153,11 +179,16 @@ pub enum ServerAuth {
     RdsIam,
     /// Generate an Azure Workload Identity auth token per connection attempt.
     AzureWorkloadIdentity,
+    /// Fetch dynamic credentials from HashiCorp Vault.
+    Vault,
 }
 
 impl ServerAuth {
     pub fn is_external_identity(&self) -> bool {
-        matches!(self, Self::RdsIam | Self::AzureWorkloadIdentity)
+        matches!(
+            self,
+            Self::RdsIam | Self::AzureWorkloadIdentity | Self::Vault
+        )
     }
 }
 
@@ -251,6 +282,13 @@ pub struct User {
     pub server_auth: ServerAuth,
     /// Optional region override for RDS IAM token generation.
     pub server_iam_region: Option<String>,
+    /// Vault path to fetch dynamic database credentials from, e.g. `database/creds/my-role`.
+    /// Required when `server_auth` is set to `vault`.
+    pub vault_path: Option<String>,
+    /// Percentage of the Vault credential lease after which credentials are refreshed.
+    ///
+    /// _Default:_ `80`
+    pub vault_refresh_percent: Option<u8>,
     /// Statement timeout.
     ///
     /// Sets the `statement_timeout` on all server connections at connection creation. This allows you to set a reasonable default for each user without modifying `postgresql.conf` or using `ALTER USER`.
@@ -630,5 +668,42 @@ server_auth = "azure_workload_identity"
         let users: Users = toml::from_str(source).unwrap();
         let user = users.users.first().unwrap();
         assert_eq!(user.server_auth, ServerAuth::AzureWorkloadIdentity);
+    }
+
+    #[test]
+    fn test_user_server_auth_vault() {
+        let source = r#"
+[[users]]
+name = "alice"
+database = "db"
+server_auth = "vault"
+vault_path = "database/creds/pgdog"
+vault_refresh_percent = 75
+"#;
+
+        let users: Users = toml::from_str(source).unwrap();
+        let user = users.users.first().unwrap();
+        assert_eq!(user.server_auth, ServerAuth::Vault);
+        assert!(user.server_auth.is_external_identity());
+        assert_eq!(user.vault_path.as_deref(), Some("database/creds/pgdog"));
+        assert_eq!(user.vault_refresh_percent, Some(75));
+    }
+
+    #[test]
+    fn test_vault_refresh_percent_out_of_range_resets_to_default() {
+        let mut users = Users {
+            users: vec![User {
+                name: "alice".into(),
+                database: "db".into(),
+                server_auth: ServerAuth::Vault,
+                vault_path: Some("database/creds/pgdog".into()),
+                vault_refresh_percent: Some(150),
+                ..Default::default()
+            }],
+            ..Default::default()
+        };
+
+        users.check(&crate::Config::default());
+        assert!(users.users.first().unwrap().vault_refresh_percent.is_none());
     }
 }
--- a/pgdog-config/src/users.rs
+++ b/pgdog-config/src/users.rs
@@ -91,23 +91,23 @@ impl Users {
             if user.server_auth == ServerAuth::Vault {
                 if user.vault_path.is_none() {
                     warn!(
-                        r#"user "{}" (database "{}") uses Vault server auth but "vault_path" is not set"#,
+                        r#"user "{}" (database "{}") uses "server_auth" = "vault" but "vault_path" is not set"#,
                         user.name, user.database
                     );
                 }
 
                 if config.vault.is_none() {
                     warn!(
-                        r#"user "{}" (database "{}") uses Vault server auth but the [vault] section is missing from pgdog.toml"#,
+                        r#"user "{}" (database "{}") uses "server_auth" = "vault" but the [vault] section is missing from pgdog.toml"#,
                         user.name, user.database
                     );
                 }
 
                 if let Some(percent) = user.vault_refresh_percent
-                    && (percent == 0 || percent > 100)
+                    && (percent == 0 || percent > 80)
                 {
                     warn!(
-                        r#"user "{}" (database "{}") has "vault_refresh_percent" of {}, expected 1-100, using default"#,
+                        r#"user "{}" (database "{}") has "vault_refresh_percent" of {}, expected 1-80, using default"#,
                         user.name, user.database, percent
                     );
                     user.vault_refresh_percent = None;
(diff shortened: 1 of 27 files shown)


pgdog-config/src/vault.rs
```diff
@@ -0,0 +1,315 @@
...
+
+    /// TLS verification mode for connections to the Vault server.
+    /// Default: `verify-full`.
+    #[serde(default)]
+    pub tls_verify: VaultTlsVerify,
+
+    /// Path to a PEM-encoded CA certificate bundle used to verify the Vault
+    /// server's TLS certificate. When provided, the bundle is added on top of
+    /// the system-wide certificate store. When omitted, only system certs are used.
+    pub tls_server_ca_certificate: Option<String>,
```
levkk: nit: you can use `PathBuf`, it implements `Deserialize`.

pgdog/src/backend/auth/vault/api.rs
```diff
@@ -0,0 +1,399 @@
...
+        .build()
+        .map_err(|e| Error::Http(format!("vault: failed to build HTTP client: {e}")))
+}
+
+// ── public API ────────────────────────────────────────────────────────────────
+
+/// Authenticate to Vault via AppRole and return a client token.
+pub async fn approle_login(
+    client: &reqwest::Client,
+    addr: &str,
```
levkk: nit: when you have so many args of the same type, the caller can get confused and pass the wrong one. Couple ways around this:

1. Define a struct that keeps those as its own fields and make this method use `&self`.
2. Combine the args into a `struct`, so the caller has to do something like this:

```rust
struct Args<'a> {
    addr: &'a str,
    role_id: &'a str,
}

approve_login(&Args { addr: "localhost", role_id: "1234" }).await?;
```

pgdog/src/main.rs
```diff
@@ -115,9 +115,24 @@ async fn pgdog(command: Option<Commands>) -> Result<(), Box<dyn std::error::Erro
     // are async, so doing this after Tokio launched seems prudent.
     net::tls::load()?;
 
+    // Fetch Vault credentials BEFORE creating pools so Address::new() picks up
+    // the vault-generated username/password from the very first connection.
+    let cfg = config::config();
```
levkk: nit: re-use `cfg` below for `let general =`

levkk:
Really cool, thanks for implementing this! I'm still reviewing, will get back to you with more comments if any asap.

pgdog/src/backend/auth/vault/mod.rs
```diff
@@ -0,0 +1,683 @@
...
+                    .map_err(|e| Error::SecretId(e.to_string()))?;
+                api::approle_login(&self.client, &self.cfg.address, role_id, &secret_id).await
+            }
+            VaultAuthMethod::Kubernetes => {
+                let role = self.cfg.kubernetes_role.as_deref().ok_or_else(|| {
+                    Error::SecretId(
+                        "vault: kubernetes_role is required for Kubernetes auth".into(),
+                    )
+                })?;
+                let jwt = tokio::fs::read_to_string(self.cfg.jwt_path())
```
levkk: nit: consider importing functions and structs, e.g.:

```rust
use tokio::fs::read_to_string;

read_to_string(&path).await?;
```

pgdog/src/backend/auth/vault/mod.rs
```diff
@@ -0,0 +1,683 @@
...
+                &self.client,
+                &self.cfg.address,
+                &token.client_token,
+                vault_path,
+            )
+            .await?;
+            if cred.lease_duration == 0 {
+                tracing::warn!(
+                    pool = %pool_name,
+                    "vault: lease_duration is 0 — credentials may not be renewable; check Vault backend config"
```
levkk:
```suggestion
                    "[vault] "lease_duration" is 0, credentials may not be renewable, check Vault backend configuration"
```

pgdog/src/backend/auth/vault/mod.rs
```diff
@@ -0,0 +1,683 @@
...
+        Ok(c) => c,
+        Err(err) => {
+            error!("vault: failed to build HTTP client: {err}");
+            return Duration::ZERO;
+        }
+    };
+
+    match client.fetch_credentials(&pools, update_config).await {
+        Ok(min_lease) => {
+            info!(pools = pools.len(), "vault: initial credentials fetched");
```
levkk:
```suggestion
            info!(pools = pools.len(), "[vault] initial credentials fetched for {} users", users.len());
```

pgdog/src/main.rs
```diff
@@ -115,9 +115,24 @@ async fn pgdog(command: Option<Commands>) -> Result<(), Box<dyn std::error::Erro
...
+    let vault_initial_delay = if let Some(vault_cfg) = cfg.config.vault.as_ref() {
+        pgdog::backend::auth::vault::init(vault_cfg, &cfg.users.users).await
```
levkk: Reloading the config (`RELOAD` admin command) will erase these I think. You may want to run the vault loop externally and fetch credentials from an in-memory cache instead, using the `database` and `user` as key.

levkk:
Great start. A few comments on style and one important architectural issue: the vault credentials are loaded on boot only, but pgdog config is dynamic and can be reloaded without restarting, I think the current implementation will reset the credentials when this happens.

Fixed my incorrect merge conflict fix in `884f14e6391a69b11b8da22ca074ca62e5e36190` (you should be able to cherry-pick it from https://github.com/pgdogdev/pgdog/pull/960).

larrywax:
> Great start. A few comments on style and one important architectural issue: the vault credentials are loaded on boot only, but pgdog config is dynamic and can be reloaded without restarting, I think the current implementation will reset the credentials when this happens.
> 
> Fixed my incorrect merge conflict fix in `884f14e6391a69b11b8da22ca074ca62e5e36190` (you should be able to cherry-pick it from #960).

Ah I see, was not aware of it. Makes sense, thanks!

pgdog/src/backend/auth/vault/mod.rs
```diff
@@ -0,0 +1,590 @@
...
+    const MAX_BACKOFF: Duration = Duration::from_secs(60);
+
+    loop {
+        match client.fetch_credentials(&pools).await {
+            Ok((creds, min_lease)) => {
+                backoff = Duration::from_secs(1);
+                for (pool_name, cred) in &creds {
+                    cache_set(pool_name, &cred.username, &cred.password);
+                }
+                if let Err(e) = reload_from_existing() {
```
levkk: I think you may want to reload the connection pools if and only if the credentials actually changed. I think vault would rotate them every 15min or so (not sure), but this would log it every second and reload the pools once per second too? Please let me know if I misunderstood.

larrywax: I haven't implemented credentials refresh yet. For now, we fetch a new set of credentials every time they are close to expiration.

This will only log the error and then wait `rotation_interval` seconds to refresh the credentials and reload the connections. The backoff kicks in only when we fail to get credentials from vault.

Perhaps spawning a single task that refreshes credentials in a loop is not ideal for error handling, it would probably be better to spawn a task for each pool

levkk: I think I agree. You can spawn per-pool tasks in `monitor.rs` (feel free to create a separate module under `backend::pool`).

larrywax: Are you sure about moving the tasks into `backend::pool`? I'm trying to implement it, but I’m not sure if calling `reload_from_existing` from this point in the code is 'correct.' Also, this way I’d have to 'leak' some Vault details into the underlying structures

levkk: Ah, good point. I forgot that `Pool` is immutable, so reloading it from there isn't right. We recently added a cache for AWS IAM and Azure Worload Identity tokens, maybe that's a good place for Vault credentials refreshing too? Check it out: https://github.com/pgdogdev/pgdog/blob/main/pgdog/src/backend/pool/token_cache.rs

larrywax: I've updated the code to use the token cache. I had to add some minor extensions because Vault, unlike other "providers", returns both username and password.

For credential refreshing, since I can no longer use `reload_from_existing`, I introduced the concept of `credentials generation`.
Every time credentials are rotated, I bump the generation counter so that checked-out connections are dropped as soon as they return to the pool -> https://github.com/pgdogdev/pgdog/pull/958/changes#diff-2521cf07ff94037c585d263e602354bdec098576f95fa82b051cb44157992a9dR386
Let me know what you think!

pgdog-config/src/users.rs
```diff
@@ -88,6 +88,32 @@ impl Users {
                 );
             }
 
+            if user.server_auth == ServerAuth::Vault {
+                if user.vault_path.is_none() {
+                    warn!(
+                        r#"user "{}" (database "{}") uses Vault server auth but "vault_path" is not set"#,
```
levkk: nit: match log format to what we do above, i.e.:

```rust
r#"user "{}" database "{}" uses "server_auth" = "vault" but "vault_path" is not set"
```

pgdog/src/backend/auth/azure_workload_identity.rs
```diff
@@ -60,6 +60,8 @@ mod tests {
             database_number: 0,
             server_auth: ServerAuth::AzureWorkloadIdentity,
             server_iam_region: None,
+            vault_path: None,
```
levkk: nit: You could do instead:

```rust
..Default::default()
```

pgdog/src/backend/auth/vault.rs
```diff
@@ -0,0 +1,263 @@
...
+use serde_json::json;
+use tracing::{debug, info};
+
+use crate::backend::pool::token_cache::{Credentials, FetchedCredentials};
+use crate::backend::{Error, pool::Address};
+use crate::config::config;
+use pgdog_config::vault::{DEFAULT_REFRESH_PERCENT, Vault, VaultAuthMethod};
+
+/// Re-login this long before the Vault client token expires.
+const TOKEN_EXPIRY_BUFFER: Duration = Duration::from_secs(60);
```
levkk: I'd make this configurable.

pgdog/src/backend/pool/inner.rs
```diff
@@ -48,6 +48,9 @@ pub(super) struct Inner {
     id: u64,
     /// Replica lag.
     pub(super) replica_lag: ReplicaLag,
+    /// Bumped each time Vault credentials rotate. Connections stamped with
+    /// an older generation are closed on check-in rather than reused.
+    pub(super) credentials_generation: u8,
```
levkk: Wouldn't this wrap around really quickly? Why not use `u64`?

pgdog/src/backend/pool/inner.rs
```diff
@@ -364,6 +381,13 @@ impl Inner {
             return Ok(result);
         }
 
+        // Vault credentials rotated while this connection was checked out.
+        // The old role will be revoked; close rather than reuse.
+        if server.credentials_generation() != self.credentials_generation {
+            server.disconnect_reason(DisconnectReason::CredentialsRefresh);
```
levkk, line 387: Curious if this is really necessary. Postres connections, once authenticated, won't get closed if the credential they authenticated with is removed. Double check me on this one, since Vault also drops users; not entirely sure what the Postgres behavior would be in that case.

levkk (approved):
This looks really good!

pgdog/src/backend/pool/inner.rs
larrywax: Vault issues these revocation statements by default
```
REVOKE ALL PRIVILEGES ON ALL TABLES IN SCHEMA public FROM "$ROLE";
REVOKE ALL PRIVILEGES ON ALL SEQUENCES IN SCHEMA public FROM "$ROLE";
REVOKE USAGE ON SCHEMA public FROM "$ROLE";
REVOKE CONNECT ON DATABASE "$DB" FROM "$ROLE";
DROP ROLE IF EXISTS "$ROLE";
```
so it's definitely needed 😅 

That's why I've set the max possible `vault_refresh_percent` to 80, otherwise the risk of errors for long-running queries would be too big IMHO

levkk:
Just need one pass with the formatter and we're good to go.

larrywax:
> Just need one pass with the formatter and we're good to go.

Ops, my bad :D should be good now

levkk (approved):
💯

PR #963, opened 2026-05-06
feat(pool): jitter for server_lifetime

### LINKED ISSUE #962: Jitter for server_lifetime to break synchronized backend reconnect waves
## Problem

When PgDog opens a cohort of backend connections in a tight window — e.g. during a traffic ramp at the start of a sale — those connections share a near-identical birth time. With a finite `server_lifetime`, the entire cohort hits the cap at near-identical times, producing one large reconnect wave at every `server_lifetime` interval. Clients that need a slot during the wave queue up while PgDog reopens connections, which surfaces as a periodic latency stripe in application metrics.

This is the same failure mode HikariCP addresses with built-in `maxLifetime` jitter and that AWS RDS Proxy avoids by spreading recycle events across a window.

## Proposal

Add a configuration option to randomize `server_lifetime` per connection.

Option A — explicit jitter knob (preferred, matches HikariCP):

```toml
[general]
server_lifetime = 3600000          # 1h base
server_lifetime_jitter = 600000    # +/- up to 10 min, sampled per connection at creation time
```

Each backend connection's effective lifetime would be `server_lifetime + uniform(-server_lifetime_jitter, +server_lifetime_jitter)`. Default `server_lifetime_jitter = 0` preserves current behavior.

Option B — implicit fractional jitter:

```toml
[general]
server_lifetime = 3600000
server_lifetime_jitter_fraction = 0.1   # +/- 10% of server_lifetime
```

Either works; (A) is more explicit and easier to reason about for operators.

### Behavioral details that matter

- Jitter must be sampled **once at connection creation**, not on every check, so a connection's lifetime is deterministic from its birth time. Otherwise a connection could keep "rolling" past its cap.
- Jitter should be additive, not subtractive-only — operators want the *average* lifetime to remain `server_lifetime`, not skew shorter.
- The randomness source can be process-global; cryptographic randomness is not required.

## Workarounds today

The only mitigations available are increasing `server_lifetime` (reduces wave frequency but doesn't eliminate the cohort effect) or removing it entirely (gives up backend-memory recycling). Neither is a substitute for jitter.

## Prior art

- HikariCP: [`maxLifetime`](https://github.com/brettwooldridge/HikariCP/wiki/MBean-(JMX)-Monitoring-and-Management) is jittered internally by 2.5% by default; documented for exactly this reason.
- AWS RDS Proxy spreads idle connection recycling across a window to avoid synchronized retirement.

Happy to send a PR if there's interest in (A). Would also be open to a `min_lifetime`/`max_lifetime` shape if maintainers prefer that.

frangz:
Closes #962.

## Summary

Adds `server_lifetime_jitter` (milliseconds) at the general, database, and user config levels. When non-zero, each backend connection samples a per-connection offset uniformly from `[-jitter, +jitter]` once at creation time, so its effective lifetime is fixed at birth and the comparison sites in the pool use that effective value instead of the bare `server_lifetime`.

This breaks the synchronized retirement of cohorts that opened together (e.g. during a traffic ramp), which otherwise produces a periodic reconnect wave at every `server_lifetime` interval — visible as a clean N-minute latency stripe in client metrics. See #962 for context and prior art.

Default is `0` (no jitter), preserving existing behavior.

## Implementation

**Config (`pgdog-config`)**
- `general.rs`: new `server_lifetime_jitter: u64` with default 0, `_Default:_` annotation, and docs URL placeholder per `pgdog-config/CONTRIBUTING.md`. Env override `PGDOG_SERVER_LIFETIME_JITTER`.
- `database.rs`, `users.rs`: `server_lifetime_jitter: Option<u64>` overrides, mirroring the existing `server_lifetime` precedence.
- `url.rs`: query-param parsing for the database-level override.

**Pool config (`pgdog-stats`, `pgdog/src/backend/pool/config.rs`)**
- New `Config::max_age_jitter: Duration` (default `Duration::ZERO`). Resolved User → Database → General, same chain as `server_lifetime`.

**Per-connection sampling (`pgdog/src/backend/pool/monitor.rs`)**
- After a successful `Server::connect`, sample once via `rand::rng().random_range(-jitter..=+jitter)` and store on the connection. Skipped (and free) when jitter is 0.

**Effective lifetime (`pgdog/src/backend/server.rs`)**
- `Server::set_max_age_offset_ms(i64)` and `Server::effective_max_age(base) -> Duration`. Saturates at zero on negative overflow; returns `base` unchanged for offset 0.

**Comparison sites (`pgdog/src/backend/pool/inner.rs`)**
- `close_old` and `maybe_check_in` now compare `server.age(now)` against `server.effective_max_age(self.config.max_age)`.

## Behavioral details (matches the issue's contract)

- Sampled **once at connection creation**, not on every check, so each connection's lifetime is deterministic from its birth time.
- Additive (uniform `[-jitter, +jitter]`) so the **average** effective lifetime stays at `server_lifetime`, not skewed shorter.
- Process-global RNG (non-cryptographic).

## Tests

10 new tests, all passing locally:

- `pgdog-config`: URL parsing of `server_lifetime_jitter`.
- `pgdog::backend::pool::config`: precedence chain (general → database → user), default-zero, and the existing precedence test extended.
- `pgdog::backend::server`: `effective_max_age` for default / positive / negative / negative-saturating cases.

Existing pool/server tests (27 in `pool::inner`, full `pgdog-config` suite) continue to pass.

`cargo fmt` applied; no new clippy warnings introduced at the touched call sites.

## Docs

The `///` doc comments reference `https://docs.pgdog.dev/configuration/pgdog.toml/{general,databases}/#server_lifetime_jitter` per `pgdog-config/CONTRIBUTING.md`. Happy to follow up with a docs PR (or update those URLs) once you tell me where the new headings should land.

### DIFF
--- a/pgdog/src/backend/pool/inner.rs
+++ b/pgdog/src/backend/pool/inner.rs
@@ -182,12 +182,12 @@ impl Inner {
     /// Close connections that have exceeded the max age.
     #[inline]
     pub(crate) fn close_old(&mut self, now: Instant) -> usize {
-        let max_age = self.config.max_age;
+        let base_max_age = self.config.max_age;
         let mut removed = 0;
 
         self.idle_connections.retain_mut(|c| {
             let age = c.age(now);
-            let keep = age < max_age;
+            let keep = age < c.effective_max_age(base_max_age);
             if !keep {
                 removed += 1;
             }
@@ -354,7 +354,7 @@ impl Inner {
         }
 
         // Close connections exceeding max age.
-        if server.age(now) >= self.config.max_age {
+        if server.age(now) >= server.effective_max_age(self.config.max_age) {
             server.disconnect_reason(DisconnectReason::Old);
             return Ok(result);
         }
--- a/pgdog/src/backend/pool/monitor.rs
+++ b/pgdog/src/backend/pool/monitor.rs
@@ -337,7 +337,8 @@ impl Monitor {
         let mut error = Error::ServerError;
         let now = Instant::now();
 
-        let max_age_jitter_ms = pool.config().max_age_jitter.as_millis() as i64;
+        let max_age = pool.config().max_age;
+        let max_age_jitter = pool.config().max_age_jitter;
 
         for attempt in 0..connect_attempts {
             match timeout(
@@ -354,7 +355,7 @@ impl Monitor {
                         guard.stats.counts.connect_time += elapsed;
                         guard.stats.counts.auth_attempts += conn.password_attempts();
                     }
-                    conn.apply_lifetime_jitter(max_age_jitter_ms);
+                    conn.apply_lifetime_jitter(max_age, max_age_jitter);
                     return Ok(conn);
                 }
 
--- a/pgdog/src/backend/pool/monitor.rs
+++ b/pgdog/src/backend/pool/monitor.rs
@@ -337,6 +337,8 @@ impl Monitor {
         let mut error = Error::ServerError;
         let now = Instant::now();
 
+        let max_age_jitter_ms = pool.config().max_age_jitter.as_millis() as i64;
+
         for attempt in 0..connect_attempts {
             match timeout(
                 connect_timeout,
@@ -344,7 +346,7 @@ impl Monitor {
             )
             .await
             {
-                Ok(Ok(conn)) => {
+                Ok(Ok(mut conn)) => {
                     let elapsed = now.elapsed();
                     {
                         let mut guard = pool.lock();
@@ -352,6 +354,7 @@ impl Monitor {
                         guard.stats.counts.connect_time += elapsed;
                         guard.stats.counts.auth_attempts += conn.password_attempts();
                     }
+                    conn.apply_lifetime_jitter(max_age_jitter_ms);
                     return Ok(conn);
                 }
 
--- a/.schema/pgdog.schema.json
+++ b/.schema/pgdog.schema.json
@@ -97,6 +97,7 @@
         "resharding_parallel_copies": 1,
         "rollback_timeout": 5000,
         "server_lifetime": 86400000,
+        "server_lifetime_jitter": 0,
         "shutdown_termination_timeout": null,
         "shutdown_timeout": 60000,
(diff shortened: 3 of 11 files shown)


pgdog/src/backend/pool/inner.rs
```diff
@@ -182,12 +182,12 @@ impl Inner {
...
     #[inline]
     pub(crate) fn close_old(&mut self, now: Instant) -> usize {
-        let max_age = self.config.max_age;
+        let base_max_age = self.config.max_age;
         let mut removed = 0;
 
         self.idle_connections.retain_mut(|c| {
             let age = c.age(now);
-            let keep = age < max_age;
+            let keep = age < c.effective_max_age(base_max_age);
```
levkk, line 190: I don't think you need to save this on the server struct. You can calculate the real max age using the offset + jitter here, since the pool has all the config settings you need. The server just needs to provide its creation time, which is already does.

levkk: Nevermind, I get it. This makes sense, instead of calculating a random jitter per connection when cleaning up the pool, you calculate it once on connection creation. This makes sense.

levkk (commented):
Just a minor stylistic change, otherwise this is great!

pgdog/src/backend/pool/monitor.rs
```diff
@@ -337,21 +337,24 @@ impl Monitor {
         let mut error = Error::ServerError;
         let now = Instant::now();
 
+        let max_age_jitter_ms = pool.config().max_age_jitter.as_millis() as i64;
```
levkk: nit: starting from here, let's propagate it as `Duration` so we are sure not to confuse units - I've done this before, even when the name of the variable is `*_ms`, it's easy to forget.

levkk (approved):
Nice!

If you'd like to follow up with docs, you can just add the setting description here: https://github.com/pgdogdev/docs/blob/main/docs/configuration/pgdog.toml/general.md#server_lifetime

I'm going to let the CI finish and merge.

🙏

PR #966, opened 2026-05-07
feat(replication): add retry logic

meskill:
- Adds lock_timeout configuration setting to pass it to pg server
- Adds retry mechanism for the replication.

This at some degree could help with the cross-shared dead-locks and fix the issue with tests in https://github.com/pgdogdev/pgdog/pull/924 if configured


levkk:
#968 feels like the better/right approach.

meskill:
> #968 feels like the better/right approach.

Yes, the #968 is more targeted approach to solve the issue with tests. But I suggest to consider this pr as a general improvement that adds new configuration and adds retries to replication process. The features could be used to fix the resharding test but this could be useful in general

levkk (approved):
Awesome!

PR #967, opened 2026-05-07
fix(replication): omni sequential send and read

meskill:
"fix" for the https://github.com/pgdogdev/pgdog/pull/924 that trades latency to the lower changes of cross-shard deadlocks for the same row update for the omni tables.


levkk:
#968 is definitely better - we can't sacrifice performance since that could block us from sharding busy databases (and that's where sharding is needed the most).

PR #968, opened 2026-05-07
fix(replication): distribute destination omni among the subscribers

meskill:
Attempt to fix issue with test https://github.com/pgdogdev/pgdog/pull/924 by distributing the destination shards between streams, so they don't overlap and therefore can't deadlock each other. At the cost of consistency in case input omni tables are not equal in the content or the definition of the table changes from source to destination.

### DIFF
--- a/integration/resharding/dev.sh
+++ b/integration/resharding/dev.sh
@@ -151,10 +151,16 @@ wait_for_no_copy_rows tasks title
 wait_for_no_copy_rows task_comments body
 wait_for_no_copy_rows settings name
 
+
+# pg_count PORT TABLE — row count via a direct postgres connection (bypasses pgdog).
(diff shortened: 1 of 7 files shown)


levkk (approved):
This is my preferred solution for sure. It doesn't sacrifice performance and fixes the original deadlock issue.

One thing to double check is that we send the WAL acknowledgement back to the source for data received via logical replication for omni tables that the connection "doesn't own"; I think this is not critical since other tables will receive writes too and that will move the WAL forward on the source, but it could be possible to have omni-only sharded clusters, theoretically.

meskill:
> One thing to double check is that we send the WAL acknowledgement back to the source for data received via logical replication for omni tables that the connection "doesn't own"; I think this is not critical since other tables will receive writes too and that will move the WAL forward on the source, but it could be possible to have omni-only sharded clusters, theoretically.

Should be ok, since the changes are made on the send level and they do not affect neither how acknowledgement to source is sent nor the lsn status update

Issue #943, opened 2026-04-28
Re-prepare statements after schema changes


levkk:
Automatically re-prepare and retry transactions that hit this error:

```
cached plan must not change result type.
```

The implementation should do something like this:

1. Catch the error
2. Re-prepare statement in request
3. Retry the request

To make this safe, make sure the request has only _one_ `Bind` & `Execute` messages - any more and the request is not safe to retry because it can run the same query twice.

### Considerations

This cannot work inside multi-statement transactions. Make sure the statement that's being re-prepared executed in a single-statement transaction.

dowzhong:
Hey, i'm guessing that beyond having only 1 bind and execute, we would also have to make sure that the statement is currently the only one in-flight right? 

for example, if we had 2 statements:

```
---- statement 1 ----
parse
bind
execute <- cached plan must not change result type
sync
---- statement 2 ----
bind
execute
sync
```

retrying statement 1 would cause it's response to come back out of order to the client (since we're not buffering statement 2, it's sent immediately after).

i could be misunderstanding this though, what do you think

levkk:
This is correct. However, our client handling code guarantees that there is only one request / statement in-flight at any given point, so the implementation here doesn't have to handle that edge case.

dowzhong:
thanks for that @levkk! gave it a shot in the above PR. I hope it's not too far off a decent implementation

levkk:
We should also double check that we are not inside a multi-statement transaction since any error will cause the transaction to be aborted. We also can't retry statements that were previously executed and it may not be safe to do so.

jkaczman:
Bind/Execute cannot be safely re-tried (see comments in #1449).

PR #970, opened 2026-05-08
feat: reprepare and retry prepared statements

dowzhong:
https://github.com/pgdogdev/pgdog/issues/943

My attempt at implementing automatic reprepare+retry of prepared statements that hit a cache error.

I thought the place place to do this would be in the server impl. That way, we can avoid dealing with sharding and duplicating the retry logic at different points. Please let me know if I should move it elsewhere though.

The main piece of work here is a small state machine in each server that keeps track of what the current batch of messages are, and if they've returned with expected. That way, we can know which messages we need to retry, and which messages we shouldn't send again. An example being

```
Parse
Describe
--- later ---
Describe
Bind <-- cache error happens here
Execute
Sync
```

In this case, we want to reprepare and only retry from bind onwards.


dowzhong:
weird, the python integration tests locally for me

meskill:
> weird, the python integration tests locally for me

integration tests are flaky sometimes, there are no errors after the restart

dowzhong:
Thanks for that, just resolved the conflicts!

PR #972, opened 2026-05-11
Add slow query parse logging

frodsan:
- Adds `log_min_duration_parse` (ms) config option: when set, emits a `WARN` log for any query whose parse time meets or exceeds the threshold.
- Adds `log_query_sample_length` (default: 1000) to cap how many characters of the query are included in the log line, keeping logs readable for long queries.

### DIFF
--- a/pgdog/src/backend/pool/cluster.rs
+++ b/pgdog/src/backend/pool/cluster.rs
@@ -76,7 +76,7 @@ pub struct Cluster {
     connection_recovery: ConnectionRecovery,
     client_connection_recovery: ConnectionRecovery,
     query_parser_engine: QueryParserEngine,
-    log_min_duration_parse: Option<u64>,
+    log_min_duration_parse: Option<Duration>,
     log_query_sample_length: usize,
     reload_schema_on_ddl: bool,
     load_schema: LoadSchema,
@@ -99,7 +99,7 @@ pub struct ShardingSchema {
     pub rewrite: Rewrite,
     /// Query parser engine.
     pub query_parser_engine: QueryParserEngine,
-    pub log_min_duration_parse: Option<u64>,
+    pub log_min_duration_parse: Option<Duration>,
     pub log_query_sample_length: usize,
 }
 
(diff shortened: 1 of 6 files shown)


pgdog/src/backend/pool/cluster.rs
```diff
@@ -76,6 +76,8 @@ pub struct Cluster {
     connection_recovery: ConnectionRecovery,
     client_connection_recovery: ConnectionRecovery,
     query_parser_engine: QueryParserEngine,
+    log_min_duration_parse: Option<u64>,
```
levkk: Yup, this is great. Just one small change, make this a `Duration` type. Once its typed, all subsequent callers don't need to guess the unit anymore.

pgdog/src/backend/pool/cluster.rs
```diff
@@ -97,6 +99,8 @@ pub struct ShardingSchema {
     pub rewrite: Rewrite,
     /// Query parser engine.
     pub query_parser_engine: QueryParserEngine,
+    pub log_min_duration_parse: Option<u64>,
```
levkk: Same as above (`Duration`).

pgdog/src/frontend/router/parser/cache/ast.rs
```diff
@@ -105,6 +106,16 @@ impl Ast {
         let mut stats = Stats::new();
         stats.parse_time += elapsed;
 
+        if let Some(threshold_ms) = schema.log_min_duration_parse {
```
levkk: Once `log_min_duration_parse` is a `Duration`, you'll be able to do a simple `>` check without worrying about units.

levkk:
One small type change and also don't forget to run `cargo fmt`.

To fix the jsonschema error, run `cargo run --bin pgdog-jsonschema` and commit the diff (`.json` change).

pgdog/src/frontend/router/parser/cache/context.rs
```diff
@@ -62,4 +62,9 @@ impl<'a> AstQuery<'a> {
             original_query: query,
         }
     }
+
+    /// Return the first `sample_len` characters of the original query, including any comment.
+    pub fn truncated_query(&self, sample_len: usize) -> String {
```
levkk: You could return a slice here, e.g.:

```rust
self.original_query.query()[0..sample_len]
```

after checking for bounds. This avoids allocating a new string (memory alloc), especially since this is used only for logging.

frodsan:
@levkk thanks for the review! I fixed the comments in https://github.com/pgdogdev/pgdog/pull/972/changes/4052ea5dae4dd495bf0e7c4b6e13516852fbf798 and https://github.com/pgdogdev/pgdog/pull/972/changes/f916d57d0c2c5b22657387714e9f10958904da9c 😄 

should we add a default to `log_min_duration_parse` to enable this feature by default? 🤔

levkk (approved):
Amazing, thank you.

PR #985, opened 2026-05-19
feat(tls): load full certificate chain for server acceptor

### LINKED ISSUE #834: TLS Certificate Chain support
There is a configuration field `tls_certificate` which can be configured to point to a PEM file, which does appear to in fact _accept_ certificate chains, but it appears the entire chain but the certificate itself is discarded. The result of this is validation of a certificate not signed _directly_ by a CA will always fail.

The code in `build_acceptor` seems to likely be where the additional certs are being discarded, and could probably be adapted with no additional changes to make TLS work correctly.

I am not aware of any way to work around this or anything I may have misconfigured.

SAY-5:
Fixes #834. `build_acceptor` read only the first certificate from `tls_certificate`, so server certs not signed directly by a root CA failed validation. Per the maintainer's note that `build_acceptor` "could probably be adapted with no additional changes to make TLS work correctly", this reads the whole PEM file with `pem_file_iter` (mirroring `build_client_cert_verifier`) and passes the full chain to `with_single_cert`. Adds a regression test with a two-cert chain fixture.


levkk (approved):
Awesome! Would be happy to merge this if you can sign the CLA.

SAY-5:
@levkk the CLA check has been green since May. I also just rebased onto main to clear the tls.rs conflict, cargo check is clean on my side.

PR #987, opened 2026-05-19
Add flattened JSON log format support

### LINKED ISSUE #986: Flattened JSON log output
Recently, JSON log format was added (in https://github.com/pgdogdev/pgdog/pull/822) which outputs the JSON in this format: `{"timestamp":"2026-05-19T10:40:00.330523Z","level":"ERROR","fields":{"message":"this is the message"}}`. This works great, but some log processors expect the level and other fields at the top level rather than nested under `fields`.

I'd like a way to configure flat JSON as the output. This could be a new `json_flattened` log_format option or a new setting altogether.

Rust library already supports this via [`flatten_event`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/struct.Layer.html#method.flatten_event), so it should be straightforward to add.

kitop:
Close https://github.com/pgdogdev/pgdog/issues/986

Builds on top of https://github.com/pgdogdev/pgdog/pull/822 and adds a new `json_flattened` `log_format` option that puts the fields at the top level.

Example output:
```
$ PGDOG_LOG_FORMAT=json_flattened RUST_LOG=info \
    cargo run -q -p pgdog -- --config /tmp/missing-pgdog.toml --users /tmp/missing-users.toml configcheck 2>&1

{"timestamp":"2026-05-19T11:24:44.191700Z","level":"WARN","message":"\"/tmp/missing-pgdog.toml\" doesn't exist, loading defaults instead","target":"pgdog_config::core"}
{"timestamp":"2026-05-19T11:24:44.192415Z","level":"WARN","message":"\"/tmp/missing-users.toml\" doesn't exist, loading defaults instead","target":"pgdog_config::core"}
{"timestamp":"2026-05-19T11:24:44.192447Z","level":"INFO","message":"[debug only] admin password: _pgdog_BApCRAcrBvNp","target":"pgdog_config::core"}
{"timestamp":"2026-05-19T11:24:44.193621Z","level":"INFO","message":"✅ config valid","target":"pgdog"}
```


levkk:
Awesome. Just run `cargo run --bin pgdog-jsonschema` and this should be good to go.

kitop:
Thanks!
Updated in [eef0b07](https://github.com/pgdogdev/pgdog/pull/987/commits/eef0b077fcf3869ca78b7f2f709c169ae7c4f93e)

PR #989, opened 2026-05-19
Use rust_decimal's builtin PG encode/decode

sgrif:
We had enabled the feature on the rust_decimal crate to compile these functions, but were never actually using them. The postgres-types crate already existed as a transitive dependency of rust_decicmal, it's only added to our Cargo.toml so we can use the trait impls that decimal has.

Our tests for all the various malformed inputs were helpful here, the only one I think is excessive is rejecting NaN with digits. All the other cases caught places where the decimal crate hit unchecked overflow. It wouldn't have resulted in UB but the fact that it returns nonsense instead of an error is something that's good for us to catch.

This does slightly change our behavior, as their encoding implementation fails to remove leading zeroes from the output when the result is an exact power of 10000. This shouldn't matter, as the two values are completely equivalent as far as PG is concerned (and I've verified as such manually). I've updated the test case to reflect this behavior, though I think if we accept this change we should probably just delete it since we're testing a crate's behavior and not our own.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3286,6 +3286,7 @@ dependencies = [
  "bytes",
  "chrono",
  "pgdog-vector",
+ "postgres-types",
  "rust_decimal",
  "schemars",
  "serde",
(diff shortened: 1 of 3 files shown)


sgrif:
@CLAassistant You can't take my rights you don't even have a soul

levkk (approved):
Image

Image

PR #990, opened 2026-05-19
Yeet unused serde impls for Numeric

sgrif:


sgrif:
@blacksmith-sh No you didn't, stop lying. Also I would rather die than fix with cursor.

levkk (approved):
I cannot approve more

meskill (approved):

PR #992, opened 2026-05-20
Maintain Numeric scale when computing AVG across shards

sgrif:
When computing this across shards, our behavior could slightly diverge from PG's. PG will maintain the scale of the column in the resulting average (or the default scale when converting a integer to a numeric). However, when we divide a Rust `Decimal`, it will potentially increase the scale and track more significant digits.

If these were floats, I would argue we should ignore this and as long as our behavior is within epsilon of what vanilla PG would do that is acceptable. But numeric is a type you use specifically in order to be able to specify the scale, so returning a more precise result isn't acceptable.


pgdog/src/backend/pool/connection/aggregate.rs
```diff
@@ -636,12 +636,14 @@ fn divide_for_average(sum: &Datum, count: &Datum) -> Option<Datum> {
...
-            let decimal = numeric.as_decimal()?.to_owned();
-            let divisor = Decimal::from_i128_with_scale(divisor_i128, 0);
+            let decimal = numeric.as_decimal()?;
+            let divisor = Decimal::from_i128(divisor_i128)?;
             if divisor == Decimal::ZERO {
                 Some(Datum::Null)
             } else {
-                Some(Datum::Numeric(Numeric::from(decimal / divisor)))
+                let mut result = decimal / divisor;
+                result.rescale(decimal.scale());
```
sgrif, line 645: This line is the relevant change. Everything else is light cleanup

levkk (approved):
Image

levkk:
The LB test is flaky.

PR #993, opened 2026-05-20
Normalize integer numerics when performing aggregate queries

sgrif:
The AVG function in PG will always convert any integer types to numeric. The scale of that result does not appear to be clearly defined, but appears to be a scale of 16 for small numbers and 0 for larger ones. Unfortunately, since this behavior isn't documented or defined, it's difficult for us to match.

That said, normalizing integers in this case will prevent our values from diverging with larger values (e.g. very small fractional parts that a single shrad query would not return), while only causing us to diverge from PG by not including trailing zeroes on small integers (e.g. we will now return `85` instead of `85.0000000000000000`), which I believe is a net win overall


levkk (approved):
Hell yes

PR #996, opened 2026-05-21
feat: add identity user parameter and use that for mtls

levkk:
```toml
[[users]]
identity = "pgdog.dev"
user = "prod"
database = "prod"
server_auth = "rds_iam"
```

Instead of matching on `user`, match on `identity`, allowing multiple users with the same identity.

### DIFF
--- a/pgdog-config/src/users.rs
+++ b/pgdog-config/src/users.rs
@@ -49,9 +49,9 @@ impl Users {
     pub fn check(&mut self, config: &Config) {
         for user in &mut self.users {
             if user.passwords().is_empty() {
-                if !config.general.passthrough_auth() {
+                if !config.general.passthrough_auth() && user.identity.is_none() {
                     warn!(
-                        r#"user "{}" (database "{}") doesn't have a password and passthrough auth is disabled"#,
+                        r#"user "{}" (database "{}") doesn't have a password, passthrough auth and mTLS are disabled"#,
                         user.name, user.database,
                     );
                 }
@@ -64,7 +64,10 @@ impl Users {
                     };
 
                     for database in databases {
-                        if min_pool_size > 0 {
+                        if min_pool_size > 0
+                            && user.server_password.is_none()
+                            && user.server_auth == ServerAuth::Password
+                        {
                             warn!(
                                 r#"user "{}" (database "{}") does not have a password configured, PgDog cannot connect to the server to maintain "min_pool_size" of {}, setting it to 0"#,
                                 user.name, database, min_pool_size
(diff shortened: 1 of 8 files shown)


pgdog-config/src/users.rs
```diff
@@ -64,7 +64,10 @@ impl Users {
                     };
 
                     for database in databases {
-                        if min_pool_size > 0 {
+                        if min_pool_size > 0
+                            && user.server_password.is_none()
```
sgrif, line 68: This and the line below feel related enough to justify encapsulating in a method

pgdog/src/backend/pool/cluster.rs
```diff
@@ -405,6 +410,12 @@ impl Cluster {
         &self.passwords
     }
 
+    /// Get user identity which should match the TLS certificate it provided
+    /// when connecting.
+    pub fn identity(&self) -> Option<&str> {
+        self.identity.as_ref().map(|s| s.as_str())
```
sgrif:
```suggestion
        self.identity.as_deref()
```

pgdog/src/backend/databases.rs
```diff
@@ -314,6 +314,15 @@ impl Databases {
         }
     }
 
+    /// Get the user TLS identity.
+    pub fn identity(&self, user: impl ToUser) -> Option<&str> {
+        if let Some(cluster) = self.databases.get(&user.to_user()) {
```
sgrif: Can't figure out how to do a multi line selection on mobile but I'd write this as `self.databases.get(&user.to_user()).and_then(|cluster| cluster.identity())`

pgdog/src/frontend/client/mod.rs
```diff
@@ -182,12 +182,15 @@ impl Client {
...
-            // matches the user name exactly. If the client is not connecting with TLS,
+            // matches the user identity exactly. If the client is not connecting with TLS,
             // this will fail.
             //
             // This is part of our mTLS implementation.
             //
-            stream.tls_cn() == Some(user)
+            let identity = databases::databases()
+                .identity((user, database))
+                .map(|s| s.to_string());
```
sgrif: Why are we cloning this only to immediately go back to a reference on the next line?

sgrif (approved):
Minor nits

sgrif:
Also feels like code that justifies a unit test imo

pgdog/src/frontend/client/mod.rs
levkk: It's coming from a global `ArcSwap` and we don't want to keep a reference to that while we check because of its fast path/slow path optimization that scares me.

levkk:
> Also feels like code that justifies a unit test imo

If I can figure out how to write one for this. TLS is ...like the worst.

pgdog-config/src/users.rs
levkk: Yes. I don't even know if it's right though yet, I just added this to make the warning go away. Need to 🤔 🤔 🤔 🤔 🤔

pgdog/src/frontend/client/mod.rs
levkk: Also because borrow checker said so

PR #997, opened 2026-05-21
Fix healthcheck for empty pools

### LINKED ISSUE #983: ELB HTTP healthcheck fails with passthrough auth
**PgDog version**
`0.1.40`

**Description**
The load balancer health check endpoint returns a false positive when passthrough auth is enabled but no pools have been configured yet.

We need to add an exception there if no pools are set, return 200 OK.

**Configuration**

```toml
[general]
passthrough_auth = "enabled"
```

sunil-2000:
## Summary
- Fixes #983
- Treat an empty pool set as healthy for the HTTP healthcheck.
- Preserve the existing behavior where PgDog reports down when all existing pools are unhealthy.
## Tests
- cargo fmt
- cargo nextest run --profile dev no_pools_is_healthy

### DIFF
--- a/pgdog/src/healthcheck.rs
+++ b/pgdog/src/healthcheck.rs
@@ -10,7 +10,7 @@ use hyper_util::rt::TokioIo;
 use tokio::net::TcpListener;
 use tracing::info;
 
-use crate::backend::databases::databases;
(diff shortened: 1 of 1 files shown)


levkk (approved):
Easy 🚀

PR #1001, opened 2026-05-26
Delete some unused config

sgrif:
When I was looking through our integration test config I saw this section which worried me slightly since it implied Rails apps (or anything using an ORM which queries schema information at runtime) aren't working out of the box.

@levkk claims we route queries to `information_schema` correctly now so this config shouldn't be needed anymore. If you're reading this on main, it would seem CI agreed.

### DIFF
--- a/integration/pgdog.toml
+++ b/integration/pgdog.toml
@@ -456,38 +456,3 @@ shard = 1
 
 [admin]
 password = "pgdog"
-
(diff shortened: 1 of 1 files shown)


levkk:
Ruby passed, so you're good to go!

levkk (approved):

PR #1002, opened 2026-05-26
feat: use dNSName primarily as TLS identity, fallback to CN

levkk:
- feat: use `dNSName` as TLS identity, fallback to `CN`
- refactor: remove `[general.client_tls_validate_cn]` setting; if `[[users]]` sets `identity`, we assume TLS auth is used


pgdog/src/frontend/client/mod.rs
```diff
@@ -180,62 +179,68 @@ impl Client {
...
+                            vec![PasswordKind::Plain(admin_password.clone())]
                         } else {
+                            cluster.passwords().to_vec()
+                        };
+
+                        if passwords.is_empty() {
                             false
+                        } else {
+                            match auth_type {
+                                AuthType::Md5 => {
```
sgrif:
```suggestion
                                AuthType::Md5 => panic!("It is 2026 my dude what are you even doing")
```

levkk: You'd be surprised

sgrif: 😭 🔫

levkk: Haha it's like 10 times faster than scram! Scram is sloowwwww

sgrif: I'd think not being able to reverse the hash on a consumer GPU from a decade ago would be a bigger concern but ¯\\\_(ツ)\_/¯

PR #1005, opened 2026-05-26
`cargo update`

sgrif:
sqlx 0.9 for some reason starts complaining that a &str we're passing it needs to be `'static` despite the signature not changing. I'm too tired to figure out why, adding a temporary would mean that the robots win, so for now I just locked that version


levkk:
ive never not had problems upgrading sqlx

sgrif:
It's a shame nobody has made an alternative that takes stability more seriously 🙃

sgrif:
Okay I'm really curious what could have broken that is only visible in Ruby

levkk:
#1006

levkk (approved):

PR #1011, opened 2026-05-27
chore: clippy

levkk:
1. Clippy pass
2. Refactor `BackendKeyData` secret into an enum and remove `Copy`

### DIFF
--- a/pgdog/src/backend/pool/error.rs
+++ b/pgdog/src/backend/pool/error.rs
@@ -1,8 +1,6 @@
 //! Connection pool errors.
 use thiserror::Error;
 
-use crate::net::BackendKeyData;
-
 #[derive(Debug, Error, PartialEq, Clone, Copy)]
 pub enum Error {
     #[error("checkout timeout")]
@@ -68,8 +66,8 @@ pub enum Error {
     #[error("pool is not healthy")]
     PoolUnhealthy,
 
-    #[error("checked in untracked connection: {0}")]
-    UntrackedConnCheckin(BackendKeyData),
+    #[error("checked in untracked connection: pid={0}")]
+    UntrackedConnCheckin(i32),
 
     #[error("mapping missing: {0}")]
     MappingMissing(usize),
--- a/pgdog/src/backend/pool/inner.rs
+++ b/pgdog/src/backend/pool/inner.rs
@@ -239,8 +239,8 @@ impl Inner {
     pub(super) fn take(&mut self, request: &Request) -> Result<Option<Box<Server>>, Error> {
         if let Some(conn) = self.idle_connections.pop() {
             self.taken.take(&Mapping {
-                client: request.id,
-                server: *(conn.id()),
+                client: request.id.clone(),
+                server: conn.id().clone(),
             })?;
 
             Ok(Some(conn))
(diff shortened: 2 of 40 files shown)


pgdog/src/backend/pool/error.rs
```diff
@@ -68,8 +66,8 @@ pub enum Error {
     #[error("pool is not healthy")]
     PoolUnhealthy,
 
-    #[error("checked in untracked connection: {0}")]
-    UntrackedConnCheckin(BackendKeyData),
```
sgrif, line 72: This seems fine if we don't need it, but I really don't think it'd be a problem to do `Box<BackendKeyData>` either

pgdog/src/backend/pool/inner.rs
```diff
@@ -548,7 +548,7 @@ mod test {
         };
 
         let server = Box::new(Server::default());
-        let server_id = *server.id();
+        let server_id = server.id().clone();
```
sgrif, line 551: These are all due to removing `Copy` from the `BackendKeyData` structs, right? I can't imagine clippy recommending this 😅

pgdog/src/backend/server.rs
```diff
@@ -314,7 +314,11 @@ impl Server {
         // so they don't send BackendKeyData.
         // Generating a random one is fine, it just won't work when we try to
         // cancel a query with this secret.
-        let id = key_data.unwrap_or(BackendKeyData::new());
+        let id = if let Some(key_data) = key_data {
+            key_data
+        } else {
+            BackendKeyData::new()
+        };
```
sgrif, line 321:
```suggestion
        let id = key_data.unwrap_or_else(BackendKeyData::new)
```

pgdog/src/backend/pool/error.rs
levkk: This was probably to preserve the `Copy` derive on the `Error` struct.

pgdog/src/backend/server.rs
levkk: That's actually what clippy complained about and told me not to do. It also made an incorrect automatic edit (`unwrap_or_default()`).

pgdog/src/backend/pool/inner.rs
levkk: This is not clippy, this is a refactor. I removed `Copy` from `BackendKeyData`, yup. It's now an enum, with the most likely variant resulting in a copy of 4 bytes, and the unlikely one an alloc of up to 256 bytes (variable secret key size in Postgres protocol 3.2 which will probably not be adopted for another decade).

sgrif (approved):
I'd prefer that the semantic changes around `BackendKeyData` have been a separate PR, since doing that and clippy at the same time.

Given that we're keeping the same semantics as the code had before, this seems fine with one minor nit. However, I do think it'd be worth actually evaluating each of those clones to see if they really need to be passing owned data or not.

pgdog/src/backend/server.rs
sgrif: Wait clippy complained about using unwrap_or_else? I would 100% expect it to complain about unwrap_or but not what I suggested. You could also do `unwrap_or_else(|| BackendKeyData::new())` if passing the function pointer seems too clever but I would be shocked if clippy actually told you to write this instead of my suggestion

sgrif:
Looks like the clippy portion of this has been fully covered by #1056 and #1057. I think this should be closed and a new PR opened to move forward with the `BackendKeyData` refactor in the future

meskill:
> Looks like the clippy portion of this has been fully covered by #1056 and #1057. I think this should be closed and a new PR opened to move forward with the `BackendKeyData` refactor in the future

the `BackendKeyData` refactor is already done https://github.com/pgdogdev/pgdog/pull/1025
so nothing left here

sgrif:
Lev clean up ur prs

levkk:
Oh yeah

sgrif:
giphy

PR #1013, opened 2026-05-27
test(resharding): add benchmarks for copy_data & replication

meskill:
Add benchmarks scripts (with hyperfine inside) for pure copy_data and replication to measure its performance.

It looks like this 
image

This is example of run for copy_data test for comparing text vs binary copy_format.

Here are only benchmarks on base of which I'm going to make some improvements and fixes


levkk (approved):
Really cool!

PR #1015, opened 2026-05-28
feat: add `query_log_stdout` config for pgcat-style query logging

marcomicera:
# Summary

Adds a `query_log_stdout` general config option that logs every query to stdout in pgcat-style format: `[pool: db][user: user] query`.

- Configurable via `pgdog.toml` or `PGDOG_QUERY_LOG_STDOUT`
- Defaults to `false`
- Logs at the start of query parsing so queries are captured even when the parser is bypassed

# Test plan

- [x] Run pgdog with `query_log_stdout = true` and confirm queries appear in logs

### DIFF
--- a/pgdog-config/src/general.rs
+++ b/pgdog-config/src/general.rs
@@ -269,6 +269,10 @@ pub struct General {
     #[serde(default)]
     pub query_log: Option<PathBuf>,
 
+    /// Log queries to stdout (pgcat-style). Format: `[pool: db][user: user] query`
+    #[serde(default = "General::query_log_stdout")]
+    pub query_log_stdout: bool,
+
     /// Minimum parse duration in milliseconds that triggers a warning log with the query text.
     /// Queries whose parsing takes longer than this value are logged at WARN level.
     /// Set to `0` or omit to disable.
@@ -800,6 +804,7 @@ impl Default for General {
             broadcast_address: Self::broadcast_address(),
             broadcast_port: Self::broadcast_port(),
             query_log: Self::query_log(),
+            query_log_stdout: Self::query_log_stdout(),
             log_min_duration_parse: Self::default_log_min_duration_parse(),
             log_query_sample_length: Self::log_query_sample_length(),
             openmetrics_port: Self::openmetrics_port(),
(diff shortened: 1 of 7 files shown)


marcomicera:
Adding tests to cover these 6 missing lines... 👨‍💻

pgdog-config/src/general.rs
```diff
@@ -269,6 +269,10 @@ pub struct General {
     #[serde(default)]
     pub query_log: Option<PathBuf>,
 
+    /// Log queries to stdout (pgcat-style). Format: `[pool: db][user: user] query`
```
levkk: I would put the db/user at the end to match our existing log format. Not sure if it's right, but at least it's consistent.

```
SELECT * FROM users WHERE id = $1 [database: oranges, user: apples]
```

pgdog/src/frontend/router/parser/query/mod.rs
```diff
@@ -98,6 +98,15 @@ impl QueryParser {
 
     /// Parse a query and return a command.
     pub fn parse(&mut self, context: RouterContext) -> Result<Command, Error> {
+        if config().config.general.query_log_stdout {
```
levkk: I would put this somewhere at the top of `QueryEngine::handle`, in its own function. The parser is not guaranteed to be used if, for example, `query_parser = "off"` in the config.

The query will be available as:

```rust
if context.query_log_stdout {
  if let Ok(Some(query)) = context.client_request.query() {
      info!("{} [database: {}, user: {}]", query.query(), ...);
   }
}
```

Also, try to follow the same pattern as other settings, e.g. load them at request start only here: https://github.com/pgdogdev/pgdog/blob/beabd7d0ccadcfbe56a54097e095f8152b7f7276/pgdog/src/frontend/client/mod.rs#L573

Loading the config (`ArcSwap` + `Arc` clone) is effectively free, but not entirely, so we try to do it once per query at most.

pgdog-config/src/general.rs
marcomicera: Fixed in 58ab86bf0534891debf91b7ced9f0670e5b1a825

pgdog/src/frontend/router/parser/query/mod.rs
marcomicera: I hope I addressed all of your points in ab15bd025d9c068952f0901c99d831f833cba86a

levkk (approved):
Cool! Let's let CI run but otherwise, good to go.

PR #1016, opened 2026-05-28
feat(resharding): use binary protocol for replication

meskill:
Use binary protocol for replication slot if `resharding_copy_format` set to binary.

Performance benefits are minimal, guess mostly because of the current sequential streaming handling in replication (send -> wait for reply for every event). 

I managed to get some improvements on benchmarks only with toxiproxy in case of limited bandwidth
image

otherwise no difference

### DIFF
--- a/pgdog/src/backend/replication/logical/publisher/slot.rs
+++ b/pgdog/src/backend/replication/logical/publisher/slot.rs
@@ -1,5 +1,6 @@
 use super::super::status::ReplicationSlot as ReplicationSlotTracker;
 use super::super::Error;
+use crate::config::config;
 use crate::{
     backend::{self, pool::Address, ConnectReason, Server, ServerOptions},
     frontend::client::query_engine::two_pc::TwoPcTransactions,
@@ -9,6 +10,8 @@ use crate::{
     },
     util::random_string,
 };
+
+use pgdog_config::CopyFormat;
 use std::{fmt::Display, str::FromStr, time::Duration};
 use tokio::time::timeout;
 use tracing::{debug, info, trace, warn};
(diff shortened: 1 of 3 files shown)


levkk (approved):
Cool! It should be a bit faster but maybe the cost of decoding the WAL is too high to matter.

pgdog/src/backend/replication/logical/publisher/slot.rs
```diff
@@ -293,10 +296,11 @@ impl ReplicationSlot {
 
     /// Start replication.
     pub async fn start_replication(&mut self) -> Result<(), Error> {
+        let is_binary = config().config.general.resharding_copy_format == CopyFormat::Binary;
         // TODO: This is definitely Postgres version-specific.
         let query = Query::new(format!(
-            r#"START_REPLICATION SLOT "{}" LOGICAL {} ("proto_version" '4', origin 'any', "publication_names" '"{}"')"#,
-            self.name, self.lsn, self.publication
+            r#"START_REPLICATION SLOT "{}" LOGICAL {} ("proto_version" '4', origin 'any', "publication_names" '"{}"', "binary" '{}')"#,
```
sgrif, line 302: I know we're not passing user input, but seeing `format!` instead of prepared statements still makes me wince 😅

pgdog/src/net/messages/replication/logical/tuple_data.rs
```diff
@@ -96,18 +97,22 @@ impl TupleData {
...
+                         caller should strip or fill toasted columns first — sending NULL"
+                    );
+                    (Parameter::new_null(), Format::Text)
                 }
+                Identifier::Format(fmt) => (Parameter::new(&c.data), *fmt),
             })
-            .collect::<Vec<_>>();
-        Bind::new_params(name, &params)
+            .unzip();
+        Bind::new_params_codes(name, &params, &codes)
```
sgrif, line 115: Not entirely relevant to this PR since you didn't add this function, but this smells like we should use a builder

sgrif (approved):

pgdog/src/backend/replication/logical/publisher/slot.rs
levkk: The logical replication protocol doesn't support prepared statements.

sgrif: RIP

sgrif: Looking at it again I don't even know why my brain parsed this as a SQL query.
IgnoreMeTheVentureBrosGIF

levkk:
Speaking of sync streaming ...we should definitely work on that - push queries without waiting for replies. There isn't any downside afaik, and we can always error-out if we eventually receive an error.

Basically have two tokio task loops run side by side:

1. Send queries
2. Receive replies and check for errors, send error via `mpsc` to parent

I bet this would make replication super quick.

meskill:
> Speaking of sync streaming ...we should definitely work on that - push queries without waiting for replies. There isn't any downside afaik, and we can always error-out if we eventually receive an error.
> 
> Basically have two tokio task loops run side by side:
> 
> 1. Send queries
> 2. Receive replies and check for errors, send error via `mpsc` to parent
> 
> I bet this would make replication super quick.

:+1: https://github.com/pgdogdev/pgdog/issues/1036

PR #1019, opened 2026-05-28
Clean up some tests

sgrif:
This removes a bunch of boilerplate, and also isolates the call to `Aggregate::parse` which I'm planning on changing the signature of in a way that would otherwise cause a lot of churn.

The &str -> SelectStmt helper is duplicated across multiple files. I plan to extract it at some point but in the short term I want to keep the changes contained.


levkk (approved):
I believe!

PR #1022, opened 2026-05-28
Remove useless Result

sgrif:
Our `ToBytes` trait returns a `Result` instead of just `Bytes`. This is a reasonable forward looking decision, except not a single impl anywhere in the code base ever returns an error. And it's a good thing they don't, because the vast majority of the code which calls this method doesn't handle the error. It just unwraps.

If it weren't for the dozens of unwraps everywhere, I would actually be in favor of leaving the Result there if we thought we might conceivably return an error in the future. But since we're not handling the errors, better to just get rid of it entirely.

### DIFF
--- a/pgdog/src/admin/server.rs
+++ b/pgdog/src/admin/server.rs
@@ -46,7 +46,7 @@ impl AdminServer {
             return Err(Error::SimpleOnly);
         }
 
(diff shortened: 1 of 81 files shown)


levkk (approved):
Yes

PR #1023, opened 2026-05-29
Remove unused QueryWithoutComment

sgrif:
We reimplemented our own Cow for this, which isn't necessarily unreasonable but wasn't needed here for a few reasons. The first is that we don't actually have an Owned case. The code appears to be written assuming that you can't take a subslice of a string with the same lifetime as the original, which isn't true.

The only other reason we'd keep this is if we cared whether a query had a comment or not at some future point, but no code ever checks for this. We can reasonably determine this by checking if role and shard are both Some.


levkk (approved):
neat

PR #1026, opened 2026-05-30
ci: fix flaky prepared statements test

### LINKED ISSUE #1006: test: flaky Ruby test
```
...F.

Failures:

  1) prepared_statements = disabled fails named extended-protocol statements in transaction pool mode
     Failure/Error: expect(errors).not_to be_empty
       expected `[].empty?` to be falsey, got true
     # ./prepared_spec.rb:80:in `block (2 levels) in <top (required)>'

Finished in 0.17507 seconds (files took 0.23506 seconds to load)
5 examples, 1 failure

Failed examples:

rspec ./prepared_spec.rb:61 # prepared_statements = disabled fails named extended-protocol statements in transaction pool mode
```

meskill:
fixes #1006

Improve reliability of prepared statement test by following this structure for testing inside the transaction pooling mode:

1. Prepare statement on some backend connection from pool
2. Pin the used connection with fake transaction so it can't be reused by other clients while transaction is alive. It'll be the same transaction since the pooled connection are reused in LIFO mode
3. Start multiple parallel connections that just picking some connections from pool and changing it's order
4. Test multiple times if the initial prepared statement works as expected depending on the configuration

These tests should basically verify the behavior of the prepared_statement options and to do this we need to use different backend connection to verify pgdog properly proxies these calls. There is no mechanism to control it from the client so the test might look clunky but the intention is to test multiple pool connection that they behave as expected when different clients use prepared statements

### DIFF
--- a/integration/ruby/prepared_disabled/prepared_spec.rb
+++ b/integration/ruby/prepared_disabled/prepared_spec.rb
@@ -44,60 +44,44 @@ describe 'prepared_statements = disabled' do
     conn2.close
   end
 
-  # Prepare on backend A. pgdog intercepts bare BEGIN without checking out a
-  # backend (deferred until the first real query), so a SELECT 1 is needed to
-  # force the actual checkout and keep backend A pinned in the open transaction.
-  # exec_prepared then lands on backend B which never saw the Parse.
-  # In disabled mode pgdog forwards Bind as-is — 'prepared statement does not exist'.
-  #
-  # Mirror: full/'executes named extended-protocol statements in
-  # transaction pool mode' — identical structure, opposite expectation.
+  # Test that the prepared statement that is created inside the connection
+  # could not be reliably executed with prepared_statements=false since it
+  # could land on another backend connection that doesn't have this
+  # statement prepared.
   it 'fails named extended-protocol statements in transaction pool mode' do
     conn = connect
-    begin
-      conn.prepare('ext_stmt', 'SELECT $1::bigint AS val')
-      pin = connect
-      begin
-        pin.exec('BEGIN')
-        pin.exec('SELECT 1') # force actual pool checkout; pin now holds backend A
-        expect { conn.exec_prepared('ext_stmt', [42]) }.to raise_error(PG::Error)
-      ensure
-        pin.exec('ROLLBACK') rescue nil
-        pin.close rescue nil
-      end
-    ensure
-      conn.close rescue nil
-    end
-  end
+    conn.prepare('ext_stmt', 'SELECT $1::bigint AS val')
 
-
-  # Same mechanism as the extended-protocol test above, but for the
-  # simple-protocol PREPARE/EXECUTE path. disabled mode forwards the SQL
-  # statement text as-is, so EXECUTE on a backend that never saw the
-  # PREPARE fails with 'prepared statement does not exist' or, if two
-  # threads hit the same backend, 'already exists'.
-  #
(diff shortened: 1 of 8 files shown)


integration/copy_data/setup.sql
```diff
@@ -93,18 +93,18 @@ SELECT
                     END,
         'notifications', (random() > 0.5)
     ) AS settings
-FROM generate_series(1, 10000) AS gs(id);
+FROM generate_series(1, 100000) AS gs(id);
```
sgrif: It's not clear from the commit context why increasing the users by 10x makes this test less flaky. Can you add that to the commit message?

integration/ruby/prepared_disabled/prepared_spec.rb
```diff
@@ -44,42 +44,33 @@
...
-        conn = connect
-        begin
-          conn.prepare('ext_stmt', 'SELECT $1::bigint AS val')
-          20.times { conn.exec_prepared('ext_stmt', [42]) }
-        rescue PG::Error => e
-          mutex.synchronize { errors << e.message }
-        ensure
-          conn.close rescue nil
-        end
+    conn = connect
```
sgrif, line 52: This seems like it pretty fundamentally changes the semantics of the test. Do we not care about the behavior of multiple competing threads? Can you clarify why we don't care in your commit message or PR description?

integration/ruby/prepared_full/prepared_spec.rb
```diff
@@ -71,35 +71,29 @@
...
-          20.times do |i|
-            res = conn.exec_prepared('ext_stmt', [i])
-            val = res[0]['val'].to_i
-            raise "expected #{i}, got #{val}" unless val == i
-          end
-        rescue => e
-          mutex.synchronize { errors << e.message }
-        ensure
-          conn.close rescue nil
+    conn = connect
```
sgrif, line 74: Same issue as the previous comment here. Can you clarify why we no longer care about the behavior with multiple competing threads?

integration/ruby/prepared_spec.rb
```diff
@@ -47,35 +47,30 @@
...
-          20.times do |i|
-            res = conn.exec_prepared('ext_stmt', [i])
-            val = res[0]['val'].to_i
-            raise "expected #{i}, got #{val}" unless val == i
-          end
-        rescue => e
-          mutex.synchronize { errors << e.message }
-        ensure
-          conn.close rescue nil
+    conn = connect
```
sgrif, line 55: Same thing here

sgrif (changes requested):
Would like to see more context given about why these changes make the tests less flaky. For the Ruby tests in particular it seems like this pretty fundamentally changes what they're testing, instead of testing the same thing in a more reproducible manner. If we're okay with that change it'd be nice to have more context as to why we're okay with that.

integration/ruby/prepared_disabled/prepared_spec.rb
meskill: I've changed the tests and the pr description to address this and other comments related to prepared statements, hope this helps

integration/copy_data/setup.sql
meskill: moved to another pr https://github.com/pgdogdev/pgdog/pull/1028

levkk (approved):
Testing a negative is hard for sure. Good job!

sgrif (approved):
ty 💜

PR #1027, opened 2026-06-01
Cache queries on whether they are direct-to-shard + SQL

sgrif:
This change allows the query rewriter to safely modify its behavior for direct-to-shard queries, and skip rewrite steps that are only needed for cross-shard queries.

Though this commit does not take advantage of this, future changes will require this behavior and changing the cache key required enough code to justify being pulled into its own commit.

The signature of Map::get and Borrow means this needs a bit more boilerplate than I would have liked. There is no way to implement Borrow<(str, _)> for (String, _) (or any other owned/ref types). Instead we have to erase the types to a trait object so we can get individual references to the fields when we need them for comparison and hashing.

This trait is implemented for any combination of values/refs, and a special case for (Arc<A>, B). If in the future this trait is needed with additional pointer types, or Arc in the right position, additional manual impls will need to be added.

### DIFF
--- a/pgdog/src/frontend/router/parser/cache/cache_impl.rs
+++ b/pgdog/src/frontend/router/parser/cache/cache_impl.rs
@@ -10,8 +10,8 @@ use parking_lot::{Mutex, RawMutex};
 use std::sync::Arc;
 use tracing::debug;
 
-use super::super::{Error, Route};
-use super::{super::parse_edge_comment, Ast, AstContext, AstQuery};
+use super::super::{Error, Route, ShardOptionExt};
+use super::{super::parse_edge_comment, key_pair::KeyPair, Ast, AstContext, AstQuery};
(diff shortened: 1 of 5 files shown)


levkk:
> This change allows the query rewriter to safely modify its behavior for direct-to-shard queries, and skip rewrite steps that are only needed for cross-shard queries.

Based on my reading, this change only actually detects if the user passed in `/* pgdog_shard: x */` manual routing hint. That's not enough to detect whether the query is direct-to-shard or not, you need to cache the output of the "parser" in `parser/query/mod.rs`. There is a re-usable struct you can use for some of it, but there are quite a few edge cases, e.g. schema-based sharding which users the value of `search_path` or the schema on a fully-qualified table, e.g. `SELECT * FROM shard_0.users;` and also the value of the `pgdog.shard` connection parameter which is another way to specify manual routing hints.

PR #1028, opened 2026-06-02
fix(resharding): fix issues with copy_data (caused copy_data test to fail often)

meskill:
Found multiple issues that caused copy_data/retry_test to fail occasionally.

- Sometime copy_data failed because it finished too fast and the retry_test haven't caught on some checkpoints because of this -- I increased the amount of generated data for test to lower such events
- Not all suitable errors were handled as retryable, mostly because of wrappers -- fixed in the error definitions
- The fail on second shard caused the partial commit that broke the retries due to unique constraints on first shard. Multiple things to resolve this:
a. refactor flow, check and the error message in the case we fail and found some rows on destination tables. 
b. add new test commit_sync that specifically challenge the partial commit and fails if we reproduce it
c. wrap copy_data in transaction to make sure the last CopyDone actually reached all the server (means the backends are healthy) and only after that commit the transaction. (2pc desirable?)

### DIFF
--- a/integration/copy_data/commit_sync/cleanup.sql
+++ b/integration/copy_data/commit_sync/cleanup.sql
@@ -1,0 +1,5 @@
+-- Tear down the commit_sync test objects. Shared by setup_schema.sql, setup.sql,
+-- and run.sh's cleanup trap. DROP IF EXISTS keeps it a safe no-op on the shards
+-- (which have no publication) and on a fresh database.
+DROP PUBLICATION IF EXISTS commit_sync;
+DROP SCHEMA IF EXISTS commit_sync CASCADE;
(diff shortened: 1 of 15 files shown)


levkk (approved):
>  wrap copy_data in transaction to make sure the last CopyDone actually reached all the server (means the backends are healthy) and only after that commit the transaction. (2pc desirable?)

Agreed. We could also add 2pc support here if its enabled on the shards (we are running it in prod now, it's good, so worth putting on the backlog).

meskill:
> Agreed. We could also add 2pc support here if its enabled on the shards (we are running it in prod now, it's good, so worth putting on the backlog).

👍🏻 https://github.com/pgdogdev/pgdog/issues/1035

PR #1030, opened 2026-06-03
Allow disabling idle healthchecks

### LINKED ISSUE #1029: Allow disabling idle backend healthchecks
## Proposal: allow disabling idle backend healthchecks

I’m using PgDog primarily as a Postgres connection pooler with passthrough auth, mainly to cap the number of backend connections to the final DB instance.

My setup has many distinct Postgres users connecting through PgDog to the same database. With passthrough auth, PgDog can create per-user/per-database pools dynamically. I already use:

 ```toml
pooler_mode = "transaction"
min_pool_size = 0
```

This works well for avoiding a permanent backend connection per user. However, the independent idle healthcheck loop can still create backend activity for otherwise idle pools. In particular, if a pool has no idle backend connection, the idle healthcheck path appears to create a temporary backend connection just to run the healthcheck, then closes it.

For deployments with many low-traffic passthrough users, this can lead to periodic connection/auth churn even when no client is actively querying. It is especially noticeable when a user has been deleted or credentials are no longer valid: the healthcheck attempts are expected to fail, but PgDog keeps retrying them in the background.

### Desired behavior

I’d like a way to say:

> If a pool has no backend connections and no clients waiting, do not open a backend connection just for an idle healthcheck.

More generally, it would be useful to disable the independent idle healthcheck loop while keeping normal on-demand behavior intact. For example:

```toml
idle_healthcheck_interval = 0
```

  or a more explicit setting such as:

```toml
idle_healthchecks = false
```

With this enabled/disabled behavior:

  - Existing idle backend connections can still be closed by idle_timeout.
  - New backend connections are opened only when a client actually needs one.
  - Checkout-time healthchecks could remain unchanged for reused backend connections.
  - PgDog would not proactively validate completely idle/empty passthrough pools.

### Tradeoff

The tradeoff seems acceptable for this use case: PgDog would no longer proactively discover that an unused pool/database/user is unhealthy. The first real query after a quiet period may pay the cost of connection creation or discover that auth/network/backend access no longer works.

For passthrough-auth deployments with many rarely-used users, that seems preferable to continuously opening backend connections for healthchecks that may have no user-visible benefit.

### Question

Would this behavior make sense for PgDog? If so, is the preferred implementation an explicit boolean setting, or should an existing numeric config value such as idle_healthcheck_interval = 0 be treated as “disabled”?

I'm happy to send a PR once this is clear.

kitop:
Close https://github.com/pgdogdev/pgdog/issues/1029

Treat `idle_healthcheck_interval = 0` as disabled so deployments can avoid background idle healthchecks when they are not useful.

This only disables background idle healthchecks. Checkout-time healthchecks controlled by `healthcheck_interval` should keep their current behavior, with `healthcheck_interval = 0` meaning “check every checkout.” - Please let me know if you'd like me to work on changing behaviour there too, as now `0` can mean something different. I think this different behaviour makes sense but it has to be clearly documented.

### DIFF
--- a/.schema/pgdog.schema.json
+++ b/.schema/pgdog.schema.json
@@ -785,7 +785,7 @@
           "minimum": 0
         },
         "idle_healthcheck_interval": {
-          "description": "Frequency of healthchecks performed by PgDog on idle connections.\n\n_Default:_ `30000`\n\nhttps://docs.pgdog.dev/configuration/pgdog.toml/general/#idle_healthcheck_interval",
+          "description": "Frequency of healthchecks performed by PgDog on idle connections.\nSet to `0` to disable idle healthchecks.\n\n_Default:_ `30000`\n\nhttps://docs.pgdog.dev/configuration/pgdog.toml/general/#idle_healthcheck_interval",
           "type": "integer",
           "format": "uint64",
           "default": 30000,
--- a/example.pgdog.toml
+++ b/example.pgdog.toml
@@ -49,6 +49,7 @@ pooler_mode = "transaction"
 healthcheck_interval = 30_000
 # How often to check databases with a health check. This happens independently from clients
 # and runs on a separate loop. This is helpful if databases aren't frequently used.
+# Set to 0 to disable idle health checks.
 #
 # Default: 30 seconds
 idle_healthcheck_interval = 30_000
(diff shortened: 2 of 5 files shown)


levkk (approved):
That is a good point regarding potential confusion between `idle_healthcheck_interval` and `healthcheck_interval` meanings of "0". I think some tools use -1 to disable something, but that's very "C-like" and doesn't feel like it belongs in a Rust project :smile:.

I think you're right and with the power of good documentation we can overcome this ambiguity. I don't expect people to frequently change these unless they know what they want exactly, in which case, the docs will help.

Happy to merge as is, just letting CI run.

PR #1032, opened 2026-06-04
security: use constant_eq for password checks

meskill:
Let's just be on par with Postgres itself on this
https://www.postgresql.org/support/security/CVE-2026-6478/


levkk (approved):
Really cool!

PR #1038, opened 2026-06-05
Give a more meaningful error message when a helper column is missing

sgrif:
Any time we're decoding a row expecting a specific column to exist, such as when we're grabbing the value of a helper column we added for an aggregate function, the absence of that column represents a bug in pgdog. We should provide a more meaningful error message than "decoder was missing data". Unfortunately, since the error comes from the absence of the description for a column, we don't have any useful information to provide for debugging beyond an index. But at minimum we can be clear this is due to a bug in pgdog, and ask the user to open an issue with details about the query that we can use to debug.


levkk (approved):
love it

sgrif:
Failures are from using the wrong Rust version, I failed to extract the version bump from #1020. Just going to wait for that one before merging since it's in the final steps

PR #1039, opened 2026-06-05
Change Datum to only perform checked addition

sgrif:
Datum does not represent a type that can unconditionally add to itself. Currently, any time we encounter two types which do not match, or any types which can't be added, we silently return NULL. This changes that behavior to return a specific error. The Add impl is removed, as these cases will most likely represent a bug in pgdog and shouldn't be silently swallowed.

Two potential future changes stand out to me. The first is that we currently unconditionally error on types that differ, even though we can absolutely add Integer to Bigint for example. I'm not sure that we want to be responsible for maintaining the whole matrix of cross type addition that PG supports, so for the time being I maintained the current behavior.

The second is that how we handle NULL is currently the exact opposite of what PG does. I suspect this will be because all calls to addition come from the implementation of aggregate functions, where we might get NULL from a shard that returned no rows, and we don't want that NULL to cancel out the results from other shards.

That behavior in general is reasonable, but in such a generic function like Datum::add, diverging from PG is surprising behavior, and a potential footgun waiting to go off. I think instead we should have the aggregate functions be responsible for handling NULLs, and mirror PG in this more general case.

However, I opted not to make that change in this PR, as I'm not confident we have test coverage for this case, where a cross shard query returns results from one query and NULL from another. As I refactor this code further, I intend to make that change down the line.

### DIFF
--- a/pgdog-postgres-types/src/array.rs
+++ b/pgdog-postgres-types/src/array.rs
@@ -6,7 +6,7 @@ use crate::{DataType, Datum};
 #[derive(Debug, Clone, PartialOrd, PartialEq, Eq, Hash)]
 pub struct Array {
     elements: Vec<Datum>,
-    element_oid: i32,
+    pub(crate) element_oid: i32,
     dim: Dimension,
 }
 
--- a/pgdog-postgres-types/src/datum.rs
+++ b/pgdog-postgres-types/src/datum.rs
@@ -1,5 +1,5 @@
 use std::cmp::{Ordering, PartialOrd};
-use std::ops::Add;
+use std::fmt;
 
 use bytes::Bytes;
 use pgdog_vector::{Float, Vector};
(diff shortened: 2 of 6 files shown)


levkk (approved):

meskill:
I like the direction to more explicit handling that return Result instead of implicit nulls somewhere.

But, considering the 2 points you've mentioned I do have strong feeling that the current add implementation is aggregate specific and exactly that's why the types coercion is not relevant here and nulls are handled like this. I'd actually suggest to move this implementation to the aggregate module and use it only there without affecting the base logic for Datum type (and I think we don't have a case for Datum math outside aggregate?). 

Also, please note that the things we are discussing are also related to Ord, Eq implementations on Datum. 

P.S.: do we need to encode NULL as the part of Datum or Option<Datum> could be better fit in some cases?

sgrif:
My personal preference would be Option, but NULL as part of Datum matches PG's semantics most closely.

sgrif:
I agree with you that Datum should not implement Ord or Eq, I'm currently
working on that in another branch

sgrif:
Dealt with Ord in #1041, laid out my reasoning for why this convo actually doesn't apply to `Eq` in that commit

PR #1043, opened 2026-06-09
Unify env var manipulation in tests

sgrif:
In the 2024 edition, set_var and remove_var have been changed to be unsafe functions. The behavior remains the same, however even in tests we should not have an `unsafe` block without a safety comment. To avoid having 1023721983 comments saying `SAFTEY: cargo nextest runs one test per process and does not use threads`, we should move all this behavior into the same place.


sgrif:
Thank you, codecov, for telling me that the test code I changed doesn't have test coverage. That was so useful.

PR #1047, opened 2026-06-09
Refactor and improve the functionality of AVG

sgrif:
This is pulled out from a larger refactor of all aggregate functions. As such, the state of this commit is a bit messy, and you'll have to trust me that it's a "it'll get worse before it gets better". But I pulled this out of that branch as it has significant behavioral improvements, is chunky enough to be reviewed on its own, and to prevent a long running branch that could become stale

The new home of the AVG logic represents what I plan on the shape of the aggregate accumulator to look like long term. Each function in a smaller module that is concerned *only* with accumulating state, and not with deserializing data, error handling, etc. The result is code that is easier to read and *much* easier to unit test. Things are a bit funky in the short term though, such as the unused firts argument to Avg::new. This is because long term I intend for Accumulator to just be an enum, where we store the target column in each variant.

The new Avg code also has several improvements over the old code:

- Reduced floating point errors, making us much less likely to diverge from the answer unsharded PG would give (this is demonstrated in `avg_with_double_data`, where the old code diverged by far more than epsilon
- Error handling that happens earlier, and more strictly, reducing the amount of "if we set the unsupported flag at some point in the past bail and maybe error but we don't actually have that info anymore"
- Removal of overflow. The old code required the sum of the averages from each shard to be representable. The new code does not.

I've also simplified and clarified how we rescale numeric results. As part of writing this it became clear that this isn't just a "match the bits PG sends" issue, but one of correctness. Any precision we calculate beyond what PG sends us will be a lie, since we aren't operating on the original data

Our handling of agg(distinct) hasn't changed, we'll still silently do the wrong thing in many cases. Explicit erroring for that followed by full support will both fall out of the larger structural change in the refactor once everything else is moved.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4225,9 +4225,20 @@ dependencies = [
  "postgres-types",
  "rand 0.8.6",
  "rkyv",
+ "rust_decimal_macros",
  "serde",
  "serde_json",
  "wasm-bindgen",
+]
+
+[[package]]
+name = "rust_decimal_macros"
+version = "1.40.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "74a5a6f027e892c7a035c6fddb50435a1fbf5a734ffc0c2a9fed4d0221440519"
+dependencies = [
+ "quote",
+ "syn 2.0.117",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


levkk (approved):
Here we go 🚀 🚀

sgrif:
inb4 codecov complains that I don't have test coverage for some error branches that only exist to give more meaningful error messages to the user

sgrif:
Called it lol (only uncovered lines in that file) 
image

PR #1048, opened 2026-06-09
Refactor the trivial aggregate functions to more easily unit test

sgrif:
This takes the really trivial aggregate functions, and moves their behavior into the structure I am moving all of them towards over time. The primary benefit for these functions is purely that they're much easier to unit test, and have fewer edge cases to consider in those tests.

At this point the shape that I'm moving towards should be a bit more clear. After we've moved everything out of Accumulator, it and State will become a single type, and we'll refactor the helper columns to always be present and not a mess of options.

`State::accumulate` has a decent bit of duplication of fetching the column. But I do not want the variants to care about decoding things, and keeping it as a field separate from those structs will lead to spaghetti. In the short term, duplication is cheaper than the wrong abstraction

There's a `Dummy` variant on `State` that I don't want to have there long term. It's there because I don't want `State::new` to return an `Option`, and `Accumulator::finalize` takes `&mut self` instead of `self` and I can't flip that switch until everything is moved. Ideally `State::Dummy` would never make it into main, but a long running branch becoming stale is a worse outcome than a bit of scaffold code living in main for a day or two

### DIFF
--- a/pgdog/src/backend/pool/connection/aggregate.rs
+++ b/pgdog/src/backend/pool/connection/aggregate.rs
@@ -23,6 +23,8 @@ use rust_decimal::prelude::{FromPrimitive, ToPrimitive};
 use super::Error;
 
 mod avg;
+mod cmp;
+mod count;
 
 /// GROUP BY <columns>
 #[derive(Hash, PartialEq, Eq, Debug)]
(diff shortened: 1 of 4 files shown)


levkk (approved):
Image

cool cool cool

sgrif:
Oh codecov... Complaining about a file where the coverage % improved, and the only uncovered lines added are error branches for things we don't support yet. Never change.

PR #1049, opened 2026-06-09
Refactor Sum into its own testable module

sgrif:
This doesn't add any behavioral changes beyond explicitly erroring on `sum(distinct ...)` and adding some more unit tests


levkk (approved):
i believe you

PR #1053, opened 2026-06-10
fix: admin password not checked when using passthrough auth

levkk:
When `passthrough_authentication` is enabled, we weren't checking the admin password correctly, allowing logins into the admin DB. We now always use password auth for admin DBs, using the password configured in `pgdog.toml`/`users.toml`.


sgrif (approved):

levkk:
Pushing a regression test, then should be good to go.

PR #1054, opened 2026-06-10
Yeeeeeeeeeeeeeeeeeeeeeet

sgrif:
The deleted tests are either redundant with unit tests in the refactored code, or sufficiently covered by integration tests. CodeCov will still complain

image

### DIFF
--- a/pgdog/src/frontend/router/parser/cache/test.rs
+++ b/pgdog/src/frontend/router/parser/cache/test.rs
@@ -104,7 +104,7 @@ async fn bench_ast_cache() {
         faster
     ); // 32x on my M1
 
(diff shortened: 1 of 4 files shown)


pgdog/src/frontend/router/parser/cache/test.rs
```diff
@@ -104,7 +104,7 @@ async fn bench_ast_cache() {
         faster
     ); // 32x on my M1
 
-    assert!(faster > 10.0);
+    assert!(faster > 9.0);
```
sgrif, line 107: See y'all when I have to lower this to 8

levkk (approved):
Red is good. Green is evil. This is good.

PR #1062, opened 2026-06-11
fix chunked copying of large csv rows

LnL7:
I ran into this issue when importing data using psql from a csv for a table with a relatively big jsonb config column.

```sql
\copy public.table from 'table.csv' WITH NULL 'CSV_NONE_SENTINEL' CSV HEADER;
```

For large rows a partial message seems to be forwarded to the postgres backend resulting in an error like this. I'm not sure if this patch is the best approach to handle this case but so far it seems to be working for my usecase.

```
[88115] ERROR:  unterminated CSV quoted field
[88115] CONTEXT:  COPY copy_csv_large_test, line 1: "1,{"value":"eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee..."
[88115] STATEMENT:  COPY copy_csv_large_test (id, data) FROM STDIN (FORMAT CSV)
```

### DIFF
--- a/pgdog/src/backend/pool/connection/mod.rs
+++ b/pgdog/src/backend/pool/connection/mod.rs
@@ -307,10 +307,8 @@ impl Connection {
                 .map_err(|e| Error::Router(e.to_string()))?;
             if !rows.is_empty() {
                 self.send_copy(rows).await?;
-                self.send(&client_request.without_copy_data()).await?;
-            } else {
-                self.send(client_request).await?;
             }
+            self.send(&client_request.without_copy_data()).await?;
         } else {
             // We split up the extended protocol exhange as soon as we see
             // a Flush or Sync that doesn't actually execute anything. This
--- a/integration/rust/tests/integration/copy.rs
+++ b/integration/rust/tests/integration/copy.rs
@@ -1,0 +1,69 @@
+use rust::setup::{connections_sqlx, connections_tokio};
+
+#[tokio::test]
+async fn copy_from_csv() -> Result<(), Box<dyn std::error::Error>> {
+    let conns = connections_tokio().await;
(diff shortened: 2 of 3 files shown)


levkk:
Interesting. I need to dig into this to understand the root cause (unless you have already).

LnL7:
Not entirely to be honest, maybe part of the query gets sent twice once partial and once fully buffered the entire row?
The test case reproduces the scenario so that hopefully helps as a starting point.

levkk:
I understand what's going on. The row is crossing the CooyData message boundary and we're sending it as-is to all shards instead of parsing it. Nice catch. I'll just double check the implementation of the fix one more time for performance, but otherwise... really good find!

pgdog/src/backend/pool/connection/mod.rs
```diff
@@ -307,10 +307,8 @@ impl Connection {
                 .map_err(|e| Error::Router(e.to_string()))?;
             if !rows.is_empty() {
                 self.send_copy(rows).await?;
-                self.send(&client_request.without_copy_data()).await?;
-            } else {
-                self.send(client_request).await?;
             }
+            self.send(&client_request.without_copy_data()).await?;
```
levkk, line 311: There is an assumption of protocol correctness here: if the client sends incomplete `CopyData` rows and a `CopyDone`, we will forward the `CopyDone` to the shards, causing them to close the copy operation. We should make an assertion here that the request doesn't contain both `CopyData` and `CopyDone` messages.

It would technically be the client's fault for sending an incomplete request, but we still shouldn't allow it go fly through.

PR #1065, opened 2026-06-12
feat: check if tuples belongs to same shard before rewriting

### LINKED ISSUE #1012: [Rewrite] Check INSERT tuples belong on different shards before rewriting
Currently, the multi-row INSERT rewrite doesn't check the values and blindly rewrites the statement into one INSERT per value tuple. Perform a value/shard check instead, and if all values belong to one shard, skip the rewrite and send the statement as-is to the destination shard.

### Example

```sql
INSERT INTO sharded_table (id, value) VALUES (1, 2), (2, 2)
```

Same sharding key - same shard, no need to split.

murex971:
fixes #1012 

Adds `uniform_shard()` helper that iterates the split requests and returns `Some(n)` only if every split has `Shard::Direct(n)` for the same `n`, otherwise `None`.

## Testing
added and ran following unit test:
cargo nextest run --profile dev -E 'test(same_shard_insert) | test(cross_shard_insert)'


integration/rust/tests/integration/rewrite.rs
```diff
@@ -139,6 +139,73 @@ async fn split_inserts_rewrite_moves_rows_across_shards() {
...
+
+    let row2: Option<String> = sqlx::query_scalar(&format!(
+        "SELECT value FROM {SHARDED_INSERT_TABLE} WHERE id = 2"
+    ))
+    .fetch_optional(&pool)
+    .await
+    .expect("fetch row with id=2");
+
+    assert_eq!(row1.as_deref(), Some("one"), "row id=1 should be present");
+    assert_eq!(row2.as_deref(), Some("two"), "row id=2 should be present");
```
sgrif: Don't use `fetch_optional` if the result isn't optional.

integration/rust/tests/integration/rewrite.rs
```diff
@@ -139,6 +139,73 @@ async fn split_inserts_rewrite_moves_rows_across_shards() {
...
+
+    // Both rows must land on the same shard: the optimizer must have sent
+    // the INSERT as a single statement rather than splitting it.
+    let shard0_count: i64 = sqlx::query_scalar(&format!(
+        "/* pgdog_shard: 0 */ SELECT COUNT(*)::bigint FROM {SHARDED_INSERT_TABLE} WHERE id IN (1, 2)"
```
sgrif:
```suggestion
        "/* pgdog_shard: 0 */ SELECT COUNT(*) FROM {SHARDED_INSERT_TABLE} WHERE id IN (1, 2)"
```

integration/rust/tests/integration/rewrite.rs
```diff
@@ -139,6 +139,73 @@ async fn split_inserts_rewrite_moves_rows_across_shards() {
...
+    ))
+    .fetch_one(&pool)
+    .await
+    .expect("count on shard 0");
+
+    let shard1_count: i64 = sqlx::query_scalar(&format!(
+        "/* pgdog_shard: 1 */ SELECT COUNT(*)::bigint FROM {SHARDED_INSERT_TABLE} WHERE id IN (1, 2)"
```
sgrif:
```suggestion
        "/* pgdog_shard: 1 */ SELECT COUNT(*) FROM {SHARDED_INSERT_TABLE} WHERE id IN (1, 2)"
```

pgdog/src/frontend/client/query_engine/multi_step/insert.rs
```diff
@@ -30,6 +34,25 @@ impl<'a> InsertMulti<'a> {
...
+                Shard::Direct(n) => *n,
+                _ => return None,
+            };
+            match target {
+                None => target = Some(n),
+                Some(t) if t != n => return None,
+                _ => {}
+            }
+        }
+        target
```
sgrif: This can be written much more succinctly and idiomatically using `find`

sgrif (changes requested):

sgrif:
It would also be nice to have a unit test for this in addition to the integration tests

levkk:
Agreed. The integration test doesn't assert the changed behavior - it will pass in both scenarios (if the rewrite happened or not).

sgrif:
Yeah, now that I think about it I'd just delete the integration test. This is essentially an implementation detail and not something that can be meaningfully tested at the client level

murex971:
@sgrif  @levkk removed the integration test in favour of unit tests, resolved other comments as well, thanks!

sgrif (approved):
Don't see anything worth blocking this over

Issue #1017, opened 2026-05-28
`Monitor::stats` task silently freezes on every pool simultaneously at exactly 2^30 ms uptime


frangz:
## Summary

After **exactly 2^30 milliseconds** (≈ 12 days 10 h 15 m 42 s) of pgdog uptime, the per-pool `Monitor::stats` task stops firing on every pool simultaneously, with no log signal. `Monitor::maintenance` and the request-handling path keep running normally. `stats.averages` for every pool is left frozen at whatever value it held in the last `calc_averages` call.

This has reproduced twice in production, each freeze occurring within ~20 minutes of the 2^30 ms mark from process start:

| Run | Started (UTC) | Froze (UTC) | Δ (seconds) | vs 2^30 ms (1,073,742 s) |
|---|---|---|---|---|
| 1 | Apr 16 12:20 | Apr 28 22:15 | 1,072,500 | −20 min |
| 2 | May 14 16:25 | May 27 02:50 | 1,074,300 | +9 min |

Both windows are inside the noise of "deploy timestamp vs actual process start" and "metrics scrape granularity", and the two intervals differ from each other by only 30 minutes out of 12.4 days. The periodicity is not coincidence.

2^30 ms is the level-4 / level-5 boundary in tokio's hashed timer wheel. Tokio is 1.52.3 in this build.

## Environment

- pgdog **v0.1.40**
- tokio **1.52.3**
- Linux x86_64
- `stats_period` default (15 s)

## Symptoms

For every pool on the affected host, every `avg_*` field in `SHOW POOLS` is stuck at constant decimal values. Example:

```
avg_xact_count         | 1956
avg_query_count        | 3313
avg_received           | 24626894
avg_sent               | 5648070
avg_xact_time          | 3.004
avg_idle_xact_time     | 2.348
avg_query_time         | 0.813
avg_bind_count         | 2455
avg_server_parse_count | 2455
... (every avg_* field frozen)
```

OpenMetrics gauges (`pgdog.avg_query_time`, `pgdog.avg_xact_time`, etc.) show min == max == avg over many hours, while counters like `pgdog.total_query_count.count` keep incrementing at the expected rate. Every database on the host is affected at the same instant.

## What's still working after the freeze

- Process is alive; same PID throughout.
- `Monitor::maintenance` fires every 333 ms: continuous \`closing server connection ... reason: max age\` and \`new connection requested: reason=min\` entries for every pool.
[...]

PR #1068, opened 2026-06-12
fix: stop Monitor::stats freezing at the 2^30 ms tokio timer-wheel boundary

mvanhorn:
## Summary

The per-pool stats updater in the connection-pool monitor no longer freezes. The loop's timing now uses a `tokio::time::interval` (with `MissedTickBehavior::Skip`) instead of the previous mechanism that could stop firing, so pool stats keep updating for the life of the pool.

## Why this matters

#1017: the monitor's stats timer could freeze, leaving pool statistics stale. Switching to an interval-driven tick keeps the cadence stable. The fix also guards the `stats_period = 0` config (and `PGDOG_STATS_PERIOD=0`): `tokio::time::interval(Duration::ZERO)` panics, so a zero period now disables the stats loop instead of crashing the updater at launch.

## Testing

`cargo check` passes on the touched crate. The interval path handles a normal positive period (steady ticks, missed ticks skipped rather than bursting) and the zero-period config (loop disabled, no panic).

### DIFF
--- a/pgdog/src/backend/pool/monitor.rs
+++ b/pgdog/src/backend/pool/monitor.rs
@@ -383,6 +383,10 @@ impl Monitor {
 
     async fn stats(pool: Pool) {
         let duration = pool.config().stats_period;
+        if duration.is_zero() {
+            return;
+        }
+
         let mut tick = interval(duration);
         tick.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
         tick.tick().await;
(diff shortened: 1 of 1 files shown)


levkk (approved):
Cool. The root cause is really not clear to me. While I believe it happened, the why seems suspicious at best and hallucinated at worst. Since this introduces no new behavior, I think we can roll with it and see what happens.

levkk:
Closing as dup of #1082

PR #1069, opened 2026-06-12
Handle pg_catalog.set_config session state

### LINKED ISSUE #1055: Handle `pg_catalog.set_config`
This modifies the session state (just like `SET`) except we don't handle it. `pg_dump` uses it and it blows up the `session_path` for apps and causes scary errors like "table not found".

We should parse the param and value out of this statement and update it in the client params state, so the servers remain in sync with client state.

erayack:
## Summary

   Fixes PgDog session-state tracking for `pg_catalog.set_config(...)`, which is used by tools like `pg_dump` to mutate settings such as `search_path`.

   Before this change, PgDog forwarded:

   ```sql
   SELECT pg_catalog.set_config('search_path', '...', false)
 ```

 to PostgreSQL normally, but did not update the client-side Parameters state. If the client later received a different pooled backend connection, PgDog would not replay the
 updated setting, so queries could run with a stale search_path.

 This PR:

 - recognizes simple top-level SELECT pg_catalog.set_config(<literal name>, <literal value>, <literal is_local>)
 - also supports the existing unqualified SELECT set_config(...) form for backward compatibility
 - adds a dedicated Command::SetConfig
 - executes the SQL normally so clients still receive PostgreSQL’s SELECT result row
 - applies the parameter update only after PostgreSQL sends CommandComplete
 - clears pending state on backend error
 - reuses the existing SetParam / Parameters transaction machinery
 - routes dynamic or untracked set_config(...) calls as writes

 Semantics

 set_config(name, value, false) is tracked like session SET:

 - outside a transaction: persists in client params and is replayed on future backend reuse
 - inside a transaction: stored as transaction params and committed/rolled back with the transaction

 set_config(name, value, true) is tracked like SET LOCAL:

 - inside a transaction: visible for the transaction only
 - outside a transaction: not persisted in PgDog client params

 Unsupported forms remain normal queries, for example:

 ```sql
   SELECT set_config('search_path', current_setting('search_path'), false);
   SELECT set_config($1, $2, false);
   SELECT set_config(name, value, false) FROM settings;
 ```

 Tests

 - cargo fmt
 - cargo test -p pgdog frontend::router::parser::query::test::test_set::test_set_config --no-fail-fast

 Added coverage for:

 - parser recognition of pg_catalog.set_config(...)
 - parser recognition of unqualified set_config(...)
 - is_local = true
 - dynamic arguments remaining normal write queries
 - complex/multi-statement SELECTs remaining normal write queries
 - SQLx integration coverage for backend reuse
 - SQLx rollback coverage for transaction-scoped state

 Fixes #1055

### DIFF
--- a/integration/rust/tests/sqlx/params.rs
+++ b/integration/rust/tests/sqlx/params.rs
@@ -45,6 +45,71 @@ async fn test_params() {
 }
 
 #[tokio::test]
+async fn test_set_config_search_path_replayed() {
+    let mut conn1 = PgConnection::connect("postgres://pgdog:pgdog@127.0.0.1:6432/pgdog")
+        .await
+        .unwrap();
+
+    let mut conn2 = PgConnection::connect("postgres://pgdog:pgdog@127.0.0.1:6432/pgdog")
+        .await
+        .unwrap();
+
+    let row = conn1
+        .fetch_one("SELECT pg_catalog.set_config('search_path', 'pg_catalog', false)")
+        .await
+        .unwrap();
+    assert_eq!(row.get::<String, usize>(0), "pg_catalog");
+
+    for _ in 0..25 {
+        // Conn 2 takes the connection conn1 just used.
+        conn2.execute("BEGIN").await.unwrap();
+
+        // Conn 1 is forced to get a new one, which should now be synchronized
+        // with the search_path recorded from set_config.
+        let row = conn1.fetch_one("SHOW search_path").await.unwrap();
+        assert_eq!(row.get::<String, usize>(0), "pg_catalog");
+
+        conn2.execute("COMMIT").await.unwrap();
+    }
+}
+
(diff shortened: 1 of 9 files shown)


erayack:
Stacked follow-up for incremental review: https://github.com/erayack/pgdog/pull/1

sgrif:
Thank you for your submission. I apologize for not assigning myself to the issue sooner. Unfortunately this pull request overlaps with work the pgdog staff already had in progress. Duplicate of #1072

erayack:
No worries, thanks for letting me know. If any part of this implementation or the tests are useful, you’re welcome to cherry-pick or copy them.

> Thank you for your submission. I apologize for not assigning myself to the issue sooner. Unfortunately this pull request overlaps with work the pgdog staff already had in progress. Duplicate of #1072

PR #1071, opened 2026-06-12
feat: add query_size_limit to block/warn on big queries

### LINKED ISSUE #1058: Long query logging and blocking
Log when a large query comes in and allow for us to block it early to prevent OOMs.

```toml
[general]
query_size_limit = 123456 # bytes
query_size_limit_action = "warn" # or block
```

The trick is to catch these queries _before_ they are loaded into memory. For blocking, that's possible here: https://github.com/pgdogdev/pgdog/blob/0482b6dde097c4df58c4bd7507dfd9d979179024/pgdog/src/net/messages/buffer.rs#L42

Basically, read the message size first, and if it's above the limit, return error immediately and disconnect. For warn, load it up, and log the large query maybe here: https://github.com/pgdogdev/pgdog/blob/0482b6dde097c4df58c4bd7507dfd9d979179024/pgdog/src/frontend/client/query_engine/mod.rs#L113

The `warn` action should probably just log the first few and the last few bytes of the query so the caller can debug its origin.

meskill:
fixes #1058

### DIFF
--- a/pgdog/src/net/messages/buffer.rs
+++ b/pgdog/src/net/messages/buffer.rs
@@ -7,9 +7,12 @@ use bytes::{Buf, BytesMut};
 use pgdog_stats::MessageBufferStats;
 use tokio::io::AsyncReadExt;
 
+use crate::config::config;
 use crate::net::stream::eof;
+use crate::util::sanitize_log_sample;
 
 use super::{Error, Message};
+use tracing::error;
 
 const HEADER_SIZE: usize = 5;
 
@@ -18,12 +21,15 @@ pub struct MessageBuffer {
     buffer: BytesMut,
     capacity: usize,
     stats: MessageBufferStats,
+    /// If specified the messages exceeding this number
+    /// will be rejected and cause fatal abruption.
+    size_limit_block: Option<usize>,
 }
 
 impl MessageBuffer {
     /// Create new cancel-safe
     /// message buffer.
-    pub fn new(capacity: usize) -> Self {
+    pub fn new(capacity: usize, size_limit_block: Option<usize>) -> Self {
         Self {
             buffer: BytesMut::with_capacity(capacity),
             capacity,
@@ -31,7 +37,13 @@ impl MessageBuffer {
                 bytes_alloc: capacity,
                 ..Default::default()
             },
+            size_limit_block,
         }
+    }
+
+    /// Update the size limit used to block oversized query messages.
+    pub fn set_size_limit_block(&mut self, size_limit_block: Option<usize>) {
+        self.size_limit_block = size_limit_block;
     }
 
     /// Buffer capacity.
@@ -45,6 +57,24 @@ impl MessageBuffer {
     ) -> Result<Message, Error> {
         loop {
             if let Some(size) = self.message_size() {
+                if let Some(limit) = self.size_limit_block
+                    && size > limit
+                    && self.is_query_message()
+                {
+                    error!(
+                        "[large_query] blocking message: size={}B query_size_limit={}B partial_query='{}...'",
+                        size,
+                        limit,
+                        self.log_sample(size),
+                    );
+                    // Returning here leaves `size - buffer.len()` bytes unread on the
+                    // socket. The caller must be abrupt and reconnect;
+                    // We may actually recover the connection
+                    // instead - drain the remain message from socket
+                    // and return the error.
+                    return Err(Error::MessageTooLarge { size, limit });
+                }
+
                 if self.have_message() {
                     return Ok(Message::new(self.buffer.split_to(size).freeze()));
                 }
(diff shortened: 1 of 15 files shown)


pgdog/src/net/messages/buffer.rs
```diff
@@ -66,6 +96,28 @@ impl MessageBuffer {
...
+            .clamp(HEADER_SIZE, HEADER_SIZE.saturating_add(sample_size));
+        sanitize_log_sample(
+            &String::from_utf8_lossy(&self.buffer[HEADER_SIZE..sample_end]),
+            sample_size,
+        )
+    }
+
+    /// Whether the buffered message carries SQL that the query parser
+    /// will see: Query ('Q', simple protocol) or Parse ('P', extended).
+    /// The size limit protects the parser, so only these are subject to it.
```
levkk, line 117: Worth noting the invariant that the buffer will contain the message character since we already scanned the first 5 bytes at minimum (since we have the message length when we check this).

levkk (approved):
Pretty cool!

pgdog/src/net/messages/buffer.rs
```diff
@@ -66,6 +96,28 @@ impl MessageBuffer {
         }
     }
 
+    /// Sample of the (partial) oversized message for log output, bounded
+    /// to the message's own bytes (the buffer may already hold pipelined
+    /// traffic past it) before the lossy conversion; sanitize_log_sample
+    /// caps it at log_query_sample_length and strips control characters.
+    fn log_sample(&self, size: usize) -> String {
+        let sample_size = config().config.general.log_query_sample_length;
```
sgrif, line 104: Do we really need a user config for this? This seems really seems like there should be an answer that's good enough for everyone here.

pgdog/src/net/messages/buffer.rs
```diff
@@ -66,6 +96,28 @@ impl MessageBuffer {
...
+        let sample_end = size
+            .min(self.buffer.len())
+            .clamp(HEADER_SIZE, HEADER_SIZE.saturating_add(sample_size));
+        sanitize_log_sample(
+            &String::from_utf8_lossy(&self.buffer[HEADER_SIZE..sample_end]),
```
sgrif: Are we assuming the only cases where we have invalid UTF-8 are if we're slicing into a char boundary? Thoughts on 
```suggestion
            str::from_utf8(&self.buffer[HEADER_SIZE..sample_end])
                .or_else(|e| str::from_utf8(&self.buffer[HEADER_SIZE..e.valid_up_to()]))
                .unwrap_or("invalid utf-8")
```

pgdog/src/net/messages/error_response.rs
```diff
@@ -255,6 +255,19 @@ impl ErrorResponse {
...
         }
     }
+
+    pub fn query_too_large(size: usize, limit: usize) -> Self {
+        Self {
+            severity: "FATAL".into(),
+            code: "54000".into(),
+            message: "query size exceeds query_size_limit".into(),
+            detail: Some(format!(
+                "message is {} bytes, query_size_limit is {} bytes",
```
sgrif, line 299:
```suggestion
                "message is nearly as big as yo momma, query_size_limit is {} bytes",
```

pgdog/src/util.rs
```diff
@@ -266,6 +266,28 @@ pub fn raise_nofile_limit() -> u64 {
...
+pub fn truncate_utf8(s: &str, limit: usize) -> &str {
+    if s.len() <= limit {
+        s
+    } else {
+        let mut end = limit;
+        while !s.is_char_boundary(end) {
+            end -= 1;
+        }
+        &s[..end]
+    }
```
sgrif:
```suggestion
    &s[..s.floor_char_boundary(limit))]
```

sgrif (commented):
A few nits. My gut says that adding a config option shouldn't require this much code but I haven't touched this part of the code base enough to have a strong opinion there

pgdog/src/net/messages/buffer.rs
levkk: Dunno I learned it the hard way...people want to configure _everything_.

sgrif: 🙄 

I don't disagree but this just seems like such a dumb thing for someone to want to configure

meskill: the `log_query_sample_length` is already present on main, so haven't thought about it.
the default is 1000, it's a big chunk to get some scope of the query for debug, does increasing may help with it? 
not sure it makes much sense, but we probably need to provide more info in the log to help identify the client that actually generated such queries

sgrif: My bad, I missed that this wasn't new in this PR

pgdog/src/net/messages/buffer.rs
meskill: cool, I like this assumption and the suggestion

pgdog/src/util.rs
meskill: annoyingly helpful!

PR #1072, opened 2026-06-12
Handle the special case of `SELECT set_config(...)`

### LINKED ISSUE #1055: Handle `pg_catalog.set_config`

sgrif:
This intercepts queries that are in the simplest form of `set_config`, where it is a select statement containing exactly one call to that function, and all arguments are literals. Any queries like this could have easily been written using `SET` instead, but this is a form that is emitted by pg_dump, which results in sessions being left in a bad state if we don't handle it explicitly. So we have to handle it.

I took great care to avoid any allocations when we're checking if the select statement is in a form that we can handle, which was slightly more painful than I'd like since we need to check for both qualified an unqualified function names.

Eventually we need to handle this more broadly. `set_config` can appear anywhere in expression context (with undefined execution order, dear god), and can take any expression as an argument including cross-shard queries. When we have a cheaper and more walkable AST, we should at least warn when we encounter this function in any cases where we don't handle.

We can't *just* treat it exactly as `SET`, since the function does return a value and so the client may be expecting rows. I've left the implementation of this somewhat extensible so that we can at least handle selections with multiple calls to `set_config` down the road. Realistically, long term, we need to go through the whole AST and replace any calls to this with their value, while figuring out which config settings to apply when... somehow.

If we encounter a form that we can't parse (non-literals)

Ultimately the very existence of this function is something of an abomination, and I suspect we'll be avoiding dealing with more complex forms for as long as we possibly can.

Fixes #1055

### DIFF
--- a/pgdog/src/frontend/client/query_engine/fake.rs
+++ b/pgdog/src/frontend/client/query_engine/fake.rs
@@ -1,8 +1,8 @@
 use tokio::io::AsyncWriteExt;
 
 use crate::net::{
-    BindComplete, CloseComplete, CommandComplete, NoData, ParameterDescription, ParseComplete,
-    ProtocolMessage, ReadyForQuery, RowDescription,
+    BindComplete, CloseComplete, CommandComplete, DataRow, Field, NoData, ParameterDescription,
+    ParseComplete, ProtocolMessage, ReadyForQuery, RowDescription, parameter::ParameterValue,
 };
 
 use super::*;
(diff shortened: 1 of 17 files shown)


pgdog/src/frontend/client/query_engine/fake.rs
```diff
@@ -14,8 +14,23 @@ impl QueryEngine {
...
         context: &mut QueryEngineContext<'_>,
         command: &str,
+        return_value: Option<impl IntoIterator<Item = Option<&'_ ParameterValue>> + Clone>,
     ) -> Result<(), Error> {
         let mut sent = 0;
+        let return_fields = return_value
+            .clone()
+            .into_iter()
+            .flatten()
+            .map(|_| Field::text(""))
```
levkk: I think the columns need to have names?

sgrif: Columns do, result fields don't. I can double check, but I'm fairly certain this is the behavior of vanilla pg for selections that aren't columns and don't have aliases

sgrif: Double checked vanillas behavior, it does set it to the function name. I thought for sure that was something psql did on the client side. Since it's the end of the week, I'm just hard coding it for now. I'll definitely remember to fix it on Monday

PR #1073, opened 2026-06-13
Return error when ORDER BY references unknown column

Akash504-ai:
## Summary

Replace the silent no-op behavior when resolving named ORDER BY columns fails.

Previously, `Buffer::sort()` would silently skip sorting when an `ORDER BY` column name could not be found in the row description. This change returns an explicit error instead.

## Changes

* Add `OrderByColumnNotFound(String)` to multi-shard error handling.
* Make `Buffer::sort()` return `Result<(), Error>`.
* Return an error when resolving:

  * `OrderBy::AscColumn`
  * `OrderBy::DescColumn`
  * `OrderBy::AscVectorL2Column`
    fails due to a missing column.
* Propagate the error from the multi-shard execution path.

## Tests

Added tests covering:

* Missing named ORDER BY columns return `OrderByColumnNotFound`.
* Existing named ORDER BY columns continue to sort correctly.

## Notes

* Replaces the existing `TODO` comments that indicated missing-column resolution should error instead of silently not sorting.
* `cargo fmt` was run locally.

### DIFF
--- a/pgdog/src/backend/pool/connection/buffer.rs
+++ b/pgdog/src/backend/pool/connection/buffer.rs
@@ -50,7 +50,11 @@ impl Buffer {
     }
 
     /// Sort the buffer.
-    pub(super) fn sort(&mut self, columns: &[OrderBy], decoder: &Decoder) {
+    pub(super) fn sort(
+        &mut self,
+        columns: &[OrderBy],
+        decoder: &Decoder,
+    ) -> Result<(), super::multi_shard::Error> {
         // Calculate column indices once, since
         // fetching indices by name is O(number of columns).
         let mut cols = vec![];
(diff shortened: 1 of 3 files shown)


sgrif (changes requested):
This has no tests. I'm fairly certain there is no case where we would trigger this error where the shards would not have errored resolving the query. Can you provide an example?

Akash504-ai:
Thanks for taking a look. I picked this up because of the existing TODOs in `Buffer::sort()` that say it should error instead of silently not sorting. My assumption was that if a column name couldn't be resolved from the row description, returning an error would be better than silently skipping the sort.
That said, I don't currently have a real query that would hit this path. The tests I added only cover the buffer logic directly.

If these branches are effectively unreachable because query validation happens earlier, then my assumption was probably wrong. If so, I'm happy to close the PR.

sgrif:
I'm fine with us verifying the invariant and erroring if it's not upheld. But if we expect this code path to never be hit unless there's a bug in pgdog we should have the error reflect that. We shouldn't add a new error variant that implies this is user error

PR #1074, opened 2026-06-13
fix: windows build

imor:
Latest code did not build on windows.

### DIFF
--- a/integration/rust/tests/integration/tls_reload.rs
+++ b/integration/rust/tests/integration/tls_reload.rs
@@ -1,3 +1,5 @@
+// Test cases in this file are run only on unix because of missing sighup
+// support on Windows.
 #![cfg(unix)]
 
 use std::{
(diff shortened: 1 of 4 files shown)


pgdog/src/net/tweaks.rs
```diff
@@ -17,6 +17,7 @@ pub fn tweak(socket: &TcpStream, config: &Tcp) -> Result<()> {
     if let Some(interval) = config.interval() {
         params = params.with_interval(interval);
     }
+    #[cfg(target_os = "linux")]
     if let Some(retries) = config.retries() {
         params = params.with_retries(retries);
     }
```
imor, line 23: `with_retries` is not available on Windows.

integration/rust/tests/integration/tls_reload.rs
```diff
@@ -1,3 +1,7 @@
+// Test cases in this file are run only on unix because of missing sighup
+// support on Windows.
+#![cfg(unix)]
+
```
imor, line 4: Tests in this file rely on using the sighup signal on unix OSes to reload the config. Since sighup is not available on windows, this test is skipped on windows.

levkk (approved):
Thank you!

PR #1077, opened 2026-06-15
refactor(sharding): rehaul the sharded_mapping configuration

meskill:
Rehaul of sharded_mapping configuration

## What is done

- embed mapping configuration inside the `[[sharded_tables]]`instead of it being defined separately with `[[sharded_mapping]]`. The benefits of doing this:
	- make the relation between those explicit and without ambiguity when the mapping should apply, no need to duplicate fields between two and build the search keys to find the related mapping. 
	- solve the original issue - the mapping/table without table name, how they should be connected and applied together. The mapping is always defined on the proper table/catch-all table
	- no decoupled mappings in config due to misconfiguration
- the old `[[sharded_mapping]]` is left for backward compatibility, but no fixes for it were provided, therefore it's not working properly in some cases and this is intentionally:
	- if table name is not defined for mapping it will be applied only for the sharded_table without name. No more guessing if it should provided for some other tables etc. as I initially did in this pr.
	- the schema is not checked for the search and could be applied wrong if the schema is defined on the table/mapping
- the mapping is now fully compound and all the settings are used in order (List -> Range -> Default kinds) as it was previously either Range or List + Default
- split some types to config and runtime shapes - the config shape with only values parsed from toml and runtime shape with calculated/omitted fields
- add multiple validations for mapping configuration (currently shown as warn)
- add ..Ref types to avoid clonning but still be able to search in maps
- drop unreliable Hash implementations (they often diverge from the type itself)  and use other approaches instead

## The config difference

image

## Roadmap

Preserve the old definition for some time and remove in some future release leaving only the new way for configuring mapping.


levkk (approved):
Kinda complicated, but I get it. I think we may want a more tight config for sharding here. `name` is really optional if the `column` is the same in all tables, for example. We may even want to provide some kind of config builder based on the schema and users' input.

meskill:
Completely redone the mapping configuration instead of band-aid fixes, please see the updated description

levkk:
How do you identify mappings for two different tables without its name / database?

meskill:
> How do you identify mappings for two different tables without its name / database?

`[[sharded_tables.mapping]]` becomes part of `[[sharded_tables]]` definition above.

Here are the identical representation in toml

image

levkk:
Oh interesting. I guess I don't know toml that well :smile: I figured all entries were independent and you kinda had to map them together yourself, but ordering matters? Or are you relying on the user to provide the config in the right order?

levkk:
One more thing I would remove is the `kind` argument: it's redundant. We can detect if its a list or a range by the value in the mapping.

meskill:
> Oh interesting. I guess I don't know toml that well 😄 I figured all entries were independent and you kinda had to map them together yourself, but ordering matters? Or are you relying on the user to provide the config in the right order?

yes, ordering matters, so to connect mapping it should come just after the corresponding `[[sharded_table]]` and it'll error if it'll be somewhere else. Or you can just use inline array to make it obvious, like on the right side

meskill:
> One more thing I would remove is the `kind` argument: it's redundant. We can detect if its a list or a range by the value in the mapping.

done

levkk (approved):

PR #1088, opened 2026-06-18
feat: 2pc logic for resharding

### LINKED ISSUE #1035: Add 2pc support during resharding
In case prepared transactions on the shard are enabled we can use 2pc to prevent possible issues with partial commits on the destination shards during resharding

ygxio:
Use two-phase commit for `COPY` commands when resharding. This allows for a clean rollback on error.

fixes #1035

### DIFF
--- a/pgdog/src/backend/replication/logical/subscriber/copy.rs
+++ b/pgdog/src/backend/replication/logical/subscriber/copy.rs
@@ -1,9 +1,13 @@
 //! Shard COPY stream from one source
 //! between N shards.
 
+use futures::future::join_all;
 use pg_query::{NodeEnum, parse_raw};
 use pgdog_config::QueryParserEngine;
 use tracing::debug;
+
+use crate::frontend::client::query_engine::TwoPcPhase;
+use crate::frontend::client::query_engine::two_pc::{Manager, TwoPcTransaction};
 
 use crate::{
     backend::{Cluster, ConnectReason, replication::subscriber::ParallelConnection},
@@ -232,15 +236,64 @@ impl CopySubscriber {
         // scope). Shards not yet committed roll back on connection close. The
         // destination_has_rows() guard in parallel_sync.rs prevents a doomed retry if this
         // window is ever hit.
-        for (shard, server) in self.connections.iter_mut().enumerate() {
-            if let Err(error) = Self::send_and_confirm(server, Query::new("COMMIT").into()).await {
-                tracing::error!(
-                    "COMMIT failed on destination shard {shard} during copy_done: {error}; \
-                     shards committed before it stay committed, the rest roll back on \
-                     connection close"
-                );
-                return Err(error);
+        if self.cluster.two_pc_enabled() {
+            self.commit_two_pc().await?;
+        } else {
+            for (shard, server) in self.connections.iter_mut().enumerate() {
+                if let Err(error) =
+                    Self::send_and_confirm(server, Query::new("COMMIT").into()).await
+                {
+                    tracing::error!(
+                        "COMMIT failed on destination shard {shard} during copy_done: {error}; \
+                         shards committed before it stay committed, the rest roll back on \
+                         connection close"
+                    );
+                    return Err(error);
+                }
             }
+        }
+
+        Ok(())
+    }
+
+    async fn commit_two_pc(&mut self) -> Result<(), Error> {
+        let manager = Manager::get();
+        let txn = TwoPcTransaction::new();
+        let identifier = self.cluster.identifier();
+
+        let _guard_phase_1 = manager
+            .transaction_state(&txn, &identifier, TwoPcPhase::Phase1)
+            .await?;
+        self.two_pc_on_shards(&txn, TwoPcPhase::Phase1).await?;
+
+        let _guard_phase_2 = manager
+            .transaction_state(&txn, &identifier, TwoPcPhase::Phase2)
+            .await?;
+        self.two_pc_on_shards(&txn, TwoPcPhase::Phase2).await?;
+
+        manager.done(&txn).await?;
+
+        Ok(())
+    }
+
+    async fn two_pc_on_shards(
+        &mut self,
+        txn: &TwoPcTransaction,
+        phase: TwoPcPhase,
+    ) -> Result<(), Error> {
+        let mut futures = Vec::new();
+
+        for (shard, server) in self.connections.iter_mut().enumerate() {
+            let query = match phase {
+                TwoPcPhase::Phase1 => format!("PREPARE TRANSACTION '{txn}_{shard}'"),
+                TwoPcPhase::Phase2 => format!("COMMIT PREPARED '{txn}_{shard}'"),
+                _ => unreachable!(),
+            };
+            futures.push(Self::send_and_confirm(server, Query::new(query).into()));
+        }
+
+        for result in join_all(futures).await {
+            result?;
         }
 
         Ok(())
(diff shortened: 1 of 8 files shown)


levkk (approved):

meskill:
Could you please uncomment the tests in the `integration/copy_data/run.sh` and make sure you have 2pc enabled for them? We had issues with these tests and I'd like to check if 2pc fixes them

ygxio:
@meskill these tests pass for me even if two_pc flag is disabled

meskill:
> @meskill these tests pass for me even if two_pc flag is disabled
> 

yeah, they are just flaky sometimes on ci. don't worry about it, I'll be just checking ci status

levkk:
@meskill are we ok to merge or do we need anything else?

pgdog/src/backend/replication/logical/subscriber/copy.rs
```diff
@@ -232,20 +236,69 @@ impl CopySubscriber {
...
+        &mut self,
+        txn: &TwoPcTransaction,
+        phase: TwoPcPhase,
+    ) -> Result<(), Error> {
+        let mut futures = Vec::new();
+
+        for (shard, server) in self.connections.iter_mut().enumerate() {
+            let query = match phase {
+                TwoPcPhase::Phase1 => format!("PREPARE TRANSACTION '{txn}_{shard}'"),
+                TwoPcPhase::Phase2 => format!("COMMIT PREPARED '{txn}_{shard}'"),
```
meskill: this should work rn since the same logic is used for tx name as in the binding https://github.com/pgdogdev/pgdog/blob/280798a71d5bdf25c8d09aa0400e4f1e03a0c392/pgdog/src/backend/pool/connection/binding.rs#L377

but it could diverge if something changed on one side. It's especially important since the rollback is used exactly from the binding. 

Can we share some of the logic of the query or tx name creation between here and binding.rs? that'll prevent possible issues

pgdog/src/backend/replication/logical/subscriber/copy.rs
```diff
@@ -232,20 +236,69 @@ impl CopySubscriber {
...
+                _ => unreachable!(),
```
meskill: the rollback is kinda implicit and is done by the 2pc manager directly, maybe add a mention of it as the comment in the code?

meskill: There could be some issues due to the nature of cli calls i.e. when we run `pgdog data-sync` instead admin command on the running instance:

- the `Manager::monitor` should be running anyway, but we don't wait for it's shutdown in the cli path. That means if we fail the prepared transactions could be left on some shards and they won't be cleaned up.
- wal is also is not enabled in cli path, but on running instance it'll be enabled. We can get different behavior depending how exactly the copy command was run.

We probably should unify this and make sure we wait for shutdown and enable wal for the cli path. But that also means we should think how the old transactions should be recovered..

pgdog/src/backend/replication/logical/subscriber/copy.rs
```diff
@@ -232,20 +236,69 @@ impl CopySubscriber {
...
+            };
+            futures.push(Self::send_and_confirm(server, Query::new(query).into()));
+        }
+
+        for result in join_all(futures).await {
+            result?;
```
meskill, line 307: I'm not sure if it's really a problem, but the commit/rollback for prepared transactions happens inside `Manager::monitor` handler and it's async to this code meaning we may actually start new attempt before the manager logic is fired and the cleanup could happen in the middle of new copy attempt.

@levkk wdyt? should we care about it?

meskill (changes requested):

pgdog/src/backend/replication/logical/subscriber/copy.rs
ygxio: should i take up this as another PR, after thinking about crash recovery

pgdog/src/backend/replication/logical/subscriber/copy.rs
levkk: Maybe we can add a waiter to make sure the transaction is rolled back / committed?

pgdog/src/backend/replication/logical/subscriber/copy.rs
levkk: We can call this waiter on shutdown: https://github.com/pgdogdev/pgdog/blob/main/pgdog/src/frontend/client/query_engine/two_pc/manager.rs#L357

pgdog/src/backend/replication/logical/subscriber/copy.rs
levkk: That makes sense. Should we add some code to wait for the transaction to be rolled back?

meskill: yeah, something like this, but this should be probably done not just after the error, but when starting the new attempt, so we won't wait twice until the transaction is rolled back and then the sleep timeout between retries. We could wait for manager's queue exhausting, before starting to push data.

pgdog/src/backend/replication/logical/subscriber/copy.rs
ygxio: @meskill have added statement.rs for this

pgdog/src/backend/replication/logical/subscriber/copy.rs
ygxio: @meskill @levkk have added a v1 for this problem, could you review this

levkk: Taking a look!

ygxio: Sure thanks @levkk

levkk (approved):
This looks good to me. @meskill thoughts?

ygxio:
@meskill could u give a review for the changes

pgdog/src/frontend/client/query_engine/two_pc/manager.rs
```diff
@@ -148,12 +148,24 @@ impl Manager {
...
+    /// Block until the monitor has removed this transaction from the manager.
+    ///
+    /// No-op if the transaction was never registered or is already gone.
+    pub async fn wait_until_cleaned_up(&self, transaction: TwoPcTransaction) {
+        loop {
+            if !self.inner.lock().transactions.contains_key(&transaction) {
+                return;
+            }
+            self.notify.notify.notified().await;
+        }
```
meskill, line 185: Let's add some limits to the loop so we won't end up indefinitely trying to rollback the transaction. Maybe single notify and some fixed timeout, since the monitor already retries?

pgdog/src/backend/replication/logical/subscriber/copy.rs
meskill: still relevant - probably adding await for 2pc shutdown should be enough when running cli copy. just make sure it's awaited on error

meskill (changes requested):

meskill:
> @meskill could u give a review for the changes

Sorry, missed the updates. Couple of small fixes and this'll be ready to merge

pgdog/src/frontend/client/query_engine/two_pc/manager.rs
ygxio: addressed

meskill (approved):

meskill:
@yogesh1801 please, fmt fix and we'll merge

ygxio:
@meskill done

PR #1090, opened 2026-06-18
feat: prefer_primary role config

### LINKED ISSUE #1046: Default load balancing to primary, require manual opt-in to route to replica
Hello! First, as far as we can tell, the blessed way currently of configuring pgdog to default to primary is to set `pgdog.role=primary` either before each query, or in our local query pooler (e.g. TypeORM or Drizzle or whatever). Is that correct?

If so, it would be useful to us if we were able to configure this in pgdog, so there's no need to change our application code, until we progressively opt-in to serving queries from read replicas.

Thanks!

Jeremy
Ginger Labs

stewart-glabs:
Fixes: https://github.com/pgdogdev/pgdog/issues/1046

Hey I work with @gl-jkeys who reported this ticket. So, same motivation. I want to add a read replica to my setup but make it opt-in for all queries. All queries go to primary, unless a comment or parameter suggests go to a replica.

@levkk suggested adding this to `load_balancing_strategy` however this setting is for how to pick a among replicas. Prefer primary is about the default role. I decided to use a separate configuration. Let me know if I'm off base here.

There are comprehensive tests. I also created an integration test, to be cautious and cover all bases.

I could not figure out how to get all integration tests to pass. There seems to be some dependencies missing that are not documented. What do you suggest about this?

PS I used ai to help me write, and review the code and to make sure I'm following the repo guidelines, style and convention.


levkk:
Ah yeah, my bad, I meant this: https://docs.pgdog.dev/configuration/pgdog.toml/general/#read_write_split

It's two lines above where you put the new config in `pgdog-config/general.rs`.

stewart-glabs:
I thought that might be a good place to put this config, but it overshadows the behavior of `include_primary_if_replica_banned`. but if you think that's a better place, I can move it over.

levkk:
Should be fine. The two are super similar.

stewart-glabs:
Ok cool, I'll make the change, probably early next week.

stewart-glabs:
@levkk good to go!

levkk (approved):
Awesome! If you want the integration tests to run in CI, add the `run.sh` script there (can make it a separate job), but since this is already fully covered in tests, we're good to here.

PR #1095, opened 2026-06-22
feat: improve reload to preserve replica connection and paused state

### LINKED ISSUE #356: Improve RELOAD command
### Description

Improve `RELOAD` command to handle the following scenarios:

1. If adding more replicas to cluster, preserve connections to existing replicas (if their address is the same)
2. If database was paused, keep it paused after reload

murex971:
## Description

When adding new replicas to a cluster, `RELOAD` previously dropped all existing backend connections because the target count changed. Connection preservation now uses address-based matching instead of requiring identical replica counts. Existing replicas keep their connections, and new replicas start with empty pools.

If a database was paused via `PAUSE` before a `RELOAD`, the new pools now inherit the `paused` state so the database stays paused.

## Testing

replica test
```
cargo nextest run --profile dev -E 'test(test_can_move_conns_to_with_added_replica) or test(test_can_move_conns_to_with_removed_replica) or test(test_move_conns_to_with_added_replica_matches_by_address)'
```
pause test
```
cargo nextest run --profile dev -E 'test(test_move_conns_to_propagates_pause_state) or test(test_move_conns_to_does_not_pause_destination_when_source_is_not_paused)'
```

## Related Issues/Comments

closes: #356

### DIFF
--- a/pgdog/src/backend/pool/lb/mod.rs
+++ b/pgdog/src/backend/pool/lb/mod.rs
@@ -224,29 +224,42 @@ impl LoadBalancer {
     }
 
     /// Move connections from this replica set to another.
+    ///
+    /// Uses address-based matching so existing pools survive replica additions or
+    /// removals: each old target is paired with the new target that shares its
+    /// address. New targets with no matching old target start empty; old targets
+    /// with no match in the new config have their connections dropped.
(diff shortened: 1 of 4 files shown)


levkk (approved):
Nice! Thank you.

PR #1097, opened 2026-06-22
feat: add pgdog psql CLI subcommand

### LINKED ISSUE #430: Open psql shell to a database
### Description

Have the ability to create a psql shell to a database in the config from the CLI.

#### Example

```
pgdog psql --database prod --user pgdog
```

This directly opens up psql.

murex971:
Adds a psql subcommand that opens an interactive psql shell (or runs a psql command) directly against a backend database defined in pgdog.toml.

Usage
```
# interactive shell
pgdog psql --database prod --user pgdog
# run a single statement
pgdog psql --database prod --user pgdog -- -c "SELECT version()"
# run a SQL file
pgdog psql --database prod --user pgdog -- -f migration.sql
# connect to a specific shard
pgdog psql --database prod --user pgdog --shard 1
```
Any flags after -- are forwarded verbatim to psql.

closes #430

### DIFF
--- a/pgdog/src/cli.rs
+++ b/pgdog/src/cli.rs
@@ -13,7 +13,7 @@ use crate::backend::databases::databases;
 use crate::backend::replication::orchestrator::Orchestrator;
 use crate::backend::schema::sync::config::ShardConfig;
 use crate::backend::schema::sync::pg_dump::SyncState;
-use crate::config::{Config, Users};
+use crate::config::{Config, ConfigAndUsers, Role, Users};
 use crate::frontend::router::cli::RouterCli;
 
 /// PgDog is a PostgreSQL pooler, proxy, load balancer and query router.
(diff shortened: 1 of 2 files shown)


pgdog/src/cli.rs
```diff
@@ -443,3 +472,98 @@ pub async fn route(commands: Commands) -> Result<(), Box<dyn std::error::Error>>
...
+/// then exec-replaces this process with `psql`. On success this function never returns.
+/// `extra_args` are forwarded verbatim to psql after the connection flags.
+pub fn psql(
+    database: &str,
+    user: Option<&str>,
+    shard: usize,
+    extra_args: &[String],
+    config: &ConfigAndUsers,
+) -> Result<(), Box<dyn std::error::Error>> {
+    use std::os::unix::process::CommandExt;
```
meskill, line 488: that'll break the compilation on non unix platforms. Although, it's not our main target we'd prefer the pgdog to be buildable on most platforms.
Let's think how we can avoid this platform-specific features or gate it behind cfg

PR #1100, opened 2026-06-23
feat: block cross-shard sharding key updates with ON DELETE

### LINKED ISSUE #899: Handle foreign key constraints during sharding key updates
### Description

At the very least, block a sharding key update if it moves the row _and_ it has a foreign key `ON DELETE` constraint. They typically won't be correctly executed, i.e. `SET NULL` is bad and `CASCADE` is bad. The right way to handle this is to fetch all corresponding rows and move them as well, but that's a big job.

So for milestone 1.0, let's just return an error; we have the schema with FKs loaded, we can check if the `UPDATE` should be allowed.

murex971:
## Description 

Block cross-shard sharding key updates when the row is referenced by a foreign key with a destructive `ON DELETE` action (`CASCADE`, `SET NULL`, or `SET DEFAULT`). Those moves are implemented as `DELETE + INSERT`, so the delete step could silently affect related rows. PgDog now checks loaded schema metadata before the delete/insert and returns a recoverable error; same-shard updates and `ON DELETE RESTRICT / NO ACTION` are unchanged.

## Testing

```
cargo nextest run --profile dev test_foreign_key_on_delete_sharding_key_update

 cargo nextest run --profile dev 'frontend::router::parser::rewrite::statement::update::test'

 cargo nextest run --profile dev 'frontend::client::query_engine::multi_step::test'
```

## Related issues/comments

closes #899

### DIFF
--- a/pgdog-stats/src/schema.rs
+++ b/pgdog-stats/src/schema.rs
@@ -32,6 +32,10 @@ impl ForeignKeyAction {
             _ => Self::NoAction,
         }
     }
+
+    pub fn is_destructive_on_delete(&self) -> bool {
+        matches!(self, Self::Cascade | Self::SetNull | Self::SetDefault)
+    }
 }
 
 /// A foreign key reference to another table's column.
@@ -182,6 +186,27 @@ impl Schema {
             .relations
             .get(schema)
             .and_then(|tables| tables.get(name))
+    }
+
+    /// Returns true if a column is referenced by a foreign key whose ON DELETE
+    /// action would be unsafe during a sharding-key row move.
+    pub fn has_destructive_on_delete_reference(
+        &self,
+        schema: &str,
+        table: &str,
+        column: &str,
+    ) -> bool {
+        self.relations
+            .values()
+            .flat_map(|tables| tables.values())
+            .flat_map(|relation| relation.columns.values())
+            .flat_map(|column| column.foreign_keys.iter())
+            .any(|foreign_key| {
+                foreign_key.schema == schema
+                    && foreign_key.table == table
+                    && foreign_key.column == column
+                    && foreign_key.on_delete.is_destructive_on_delete()
+            })
     }
 }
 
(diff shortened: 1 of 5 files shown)


levkk (approved):
Nice!

pgdog/src/frontend/router/parser/rewrite/statement/update.rs
```diff
@@ -96,6 +97,36 @@ impl Deref for ShardingKeyUpdate {
     }
 }
 
+impl ShardingKeyUpdate {
+    pub(crate) fn target_table(&self) -> Option<Table<'_>> {
+        self.insert.table.as_ref().map(Table::from)
+    }
+
+    pub(crate) fn sharded_table<'a>(
```
levkk, line 105: I feel like this logic is duplicated, but I could be wrong.

pgdog-stats/src/schema.rs
```diff
@@ -183,6 +187,27 @@ impl Schema {
             .get(schema)
             .and_then(|tables| tables.get(name))
     }
+
+    /// Returns true if a column is referenced by a foreign key whose ON DELETE
+    /// action would be unsafe during a sharding-key row move.
+    pub fn has_destructive_on_delete_reference(
```
levkk, line 193: This is O(n), which gives me pause. For databases with 1000+ tables, this could get expensive. I wonder if we shouldn't create a reverse map of foreign key references to the tables they reference, so we can check this in O(1)?

levkk (commented):
This is really close, just a couple of comments to address.

pgdog-stats/src/schema.rs
murex971: For this milestone I kept the linear scan intentionally: it only runs on the cross-shard sharding key update path (rare), walks in-memory FK metadata rather than hitting Postgres, and the cost is really O(FK edges), not O(tables). 
Happy to add an index at Schema::new / load time in a follow-up if we add more callers.

pgdog/src/frontend/router/parser/rewrite/statement/update.rs
murex971: The name/schema matching is similar to `ShardedTables::get_table`, but not same. Here we need the configured sharding column to be in the `UPDATE SET` mapping, not match a parsed query column. I kept it on `ShardingKeyUpdate` since the runtime check only has the `rewrite plan` and `cluster config`.

PR #1101, opened 2026-06-23
~BUG: support kubernetes/docker native signals

pgray:
```
support kubernetes/docker native SIGTERM
kubernetes/docker default to sigterm so support is better than
relying on any version of kubelet out there to respect the custom signal
this adds support for sigterm through the same pathway as ctrl_c/SIGINT
```

thanks for your consideration!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,5 +24,9 @@ COPY --from=builder /build/target/release/pgdog /usr/local/bin/pgdog
 COPY --from=builder /build/target/release/libpgdog_primary_only_tables.so /usr/lib/libpgdog_primary_only_tables.so
 
 WORKDIR /pgdog
+# PgDog drains gracefully on both SIGINT and SIGTERM, so the STOPSIGNAL value no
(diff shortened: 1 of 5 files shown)


levkk:
SIGTERM currently implements a not-so-graceful (aka immediate) shutdown, so this is a breaking change and also removes a somewhat important functionality: `SIGINT` is sent first, and `SIGTERM` is typically sent afterwards if PgDog is refusing to shut down (due to long transactions).

pgray:
that's what `SIGKILL` is for though, which docker and kubernetes will send eventually

[kubernetes docs](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination)

levkk:
Interesting. Still, I kind of don't want to introduce a breaking change and I'm also confused why `STOPSIGNAL` which we configure in our `Dockerfile` isn't working for you. Could you clarify why this change is needed?

pgray:
ok, I'll try to sort it out in the helm chart and see if there's anything to contribute back. thanks for your time.

PR #1102, opened 2026-06-23
Begin porting the codebase to use the new parser

sgrif:
This is the first in a series of (hopefully smaller from this point)
commits that port the codebase over to use the new parser. The function
this commit focuses on is StatementParser::run_walk.

This commit leaks *way* outside that function, since it required most of
the glue code to pass the new AST around. The parsed AST must live at
the cache layer, since `pg_raw_parse::ParseResult` is what holds the
`MemoryContext`, and all nodes contain a lifetime referring to that
struct.

We're doing a funky thing where we parse once at the same time as the
old parser runs, but then later we unparse and reparse and start passing
that around. The reason for this is that the rewriter hasn't been ported
yet, so it only operates on the old parser's AST. If we didn't
deparse/reparse at that point, code after that which gets ported would
be based on the untransformed query.

The actual porting code is relatively straightforward. It's mostly 1:1
changes with fewer layers of indirection, but I did clean up some edge
cases in advisory locks while I was at it. The new code can handle
advisory locks coming from values clauses without an explicit column
name given, and also correctly handles completely silly cases where the
advisory lock comes form a subselect.

These edge cases are unlikely to ever show up in the wild, but it's
imporant that if we ever do encounter them, we do something sensible
instead of silently doing the wrong thing. And handling the tricky edge
cases also gave me an opportunity to stress test the capability of the
new parser's AST walking API.

### DIFF
--- a/pgdog/src/frontend/router/parser/cache/ast.rs
+++ b/pgdog/src/frontend/router/parser/cache/ast.rs
@@ -38,6 +38,9 @@ pub struct Ast {
 pub struct AstInner {
     /// Cached AST.
     pub ast: ParseResult,
+    // FIXME(sage): Rename to ast when parser port is done
+    #[cfg(feature = "new_parser")]
+    pub new_ast: pg_raw_parse::ParseResult,
     /// AST stats.
     pub stats: Mutex<Stats>,
     /// Rewrite plan.
@@ -52,6 +55,8 @@ impl AstInner {
     /// Create new AST record, with no rewrite or comment routing.
     pub fn new(ast: ParseResult) -> Self {
         Self {
+            #[cfg(feature = "new_parser")]
+            new_ast: pg_raw_parse::parse(&ast.deparse().unwrap()).unwrap(),
             ast,
             stats: Mutex::new(Stats::new()),
             rewrite_plan: RewritePlan::default(),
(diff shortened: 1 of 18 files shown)


pgdog/src/frontend/router/parser/cache/ast.rs
```diff
@@ -123,6 +128,8 @@ impl Ast {
             query_parser_engine: schema.query_parser_engine,
             inner: Arc::new(AstInner {
                 stats: Mutex::new(stats),
+                #[cfg(feature = "new_parser")]
```
levkk, line 135: 🙈

levkk:
Uber cool

pgdog/src/frontend/router/parser/cache/ast.rs
levkk: Maybe call it somewhere higher in the stack. Deparse is pretty expensive - totally fine to keep around for testing, etc.

sgrif: As I mentioned in the commit message, I can't do it earlier since that would lead to the code running without the rewriter operating on it. This is a temporary measure to allow me to parse individual functions without having to port the whole codebase at once. I do not intend for deparse/reparse to ever make it into prod

levkk:
image

sgrif:
This is a nasty diff. It's 99% glue code. The only meaningful code changes happen in pgdog/src/frontend/router/parser/statement.rs

levkk:
image

levkk (approved):
Image

PR #1105, opened 2026-06-25
Update `StatementParser::run_walk` to not recurse twice

sgrif:
I changed the behavior of the parser to not pass the root node to the callback. I found recursing based on pointer equality to be too painful as I work on porting the routing functions.

As a side effect, this should be more efficient. We did a full AST walk of each subselect looking for advisory locks, and then walked it again looking for tables. With this change they're now a single walk, and the code should be easier to follow


levkk (approved):
fancy!

PR #1106, opened 2026-06-26
refactor(api): rework of async tasks

meskill:
Rework of AsyncTasks used for running tasks for replication, resharding etc.:

- define new architecture for AsyncTasks in src/api to define generic way to manage such tasks
- specify new tasks in api that now the single source of truth for cli and admin commands, migrate cli and admin commands to api tasks
- allow tasks to run in hierarchy which makes the relation explicit and manageable from the tasks storage
- introduce lifecycle and inner states for tasks to easier track its progress through logs and SHOW TASKS
- the generic way to gracefully cancel the tasks and subtasks through shared CancellationToken
- handle errors and panics for running tasks and track this status with the error
- add tracing to track the task related logs and their relations
- update the output of SHOW_TASKS to show more info and to show historic info about finished/failed tasks.
- fix some issues with the cancellation of some tasks when they were using tokio::spawn and these tasks were continue to run after stop
- fix the issue with stopping multiple replications running when `notify.notify_one()` was used and only one stream was actually stopped instead of all of them
- fix the config cutover when the database_name was not provided (caused the wrong connection)
- add integration tests for admin commands

Screenshot 2026-06-26 160007
Screenshot 2026-06-26 160039
Screenshot 2026-06-26 162830
Screenshot 2026-06-26 163039


levkk (approved):
Image

sgrif:
image

☠️

PR #1111, opened 2026-06-26
Port Value::try_from to the new parser

sgrif:
This port diverges from the old implementation in that it doesn't try to stick any int looking ConstValue::String into an integer, as there wasn't any clear reason for us to be doing so.


levkk (approved):
True true

PR #1113, opened 2026-06-26
Port StatementParser::shard to the new parser for inserts

sgrif:
This isn't a straight 1-1 port, we evaluate things in a slightly different order. We also much more cleanly look for any subselects that might be something we can use for routing (though that branch didn't have good test coverage before, and it still doesn't now).

This also fixes a bug where inserts that should have been round robin were broadcast to all shards


sgrif:
Lol GitHub hides the actual code that changed behind "large diff hidden by default" 🙄

sgrif:
@levkk Would still like review on this since it does include behavioral changes

levkk (approved):
It looked good to me

PR #1114, opened 2026-06-29
feat: stream cross-shard ORDER BY results with k-way merge

### LINKED ISSUE #207: [Query engine] Merge/sort for sorting
### Description

Add merge/sort for `ORDER BY` cross-shard queries. This is an optimization on top of the current implementation. Right now, we are buffering all rows in memory. This is not great if result sets are large and QPS is high.

Since we return sorted results from each shard, we can efficiently use merge-sort to stream & sort results as they come in. Barely any buffering is needed.

murex971:
## Description

- Replace full in-memory buffering/sort for eligible cross-shard `ORDER BY` queries with a streaming k-way merge path.
- Keep existing buffer/sort/aggregate path unchanged for non-ORDER BY-only cases (e.g. aggregate/distinct flows).
- added unit/functionality tests and integration tests.

## Related issues/comments

closes #207

### DIFF
--- a/pgdog/src/backend/pool/connection/binding.rs
+++ b/pgdog/src/backend/pool/connection/binding.rs
@@ -103,7 +103,7 @@ impl Binding {
                             return Ok(message);
                         }
                         let mut read = false;
-                        for (position, server) in shards.iter_mut().enumerate() {
+                        for (shard_position, server) in shards.iter_mut().enumerate() {
                             if !server.has_more_messages() {
                                 continue;
                             }
@@ -111,7 +111,7 @@ impl Binding {
                             let message = server.read().await?;
 
                             read = true;
-                            if let Some(message) = state.forward_from(position, message)? {
+                            if let Some(message) = state.forward_from(shard_position, message)? {
                                 return Ok(message);
                             }
                         }
@@ -144,11 +144,11 @@ impl Binding {
                 let mut shards_sent = servers.len();
                 let mut futures = Vec::new();
 
-                for (position, server) in servers.iter_mut().enumerate() {
+                for (shard_position, server) in servers.iter_mut().enumerate() {
                     // Map positional index to actual shard number.
                     // When only a subset of shards is connected (Shard::Multi binding),
                     // positional indices don't match actual shard numbers.
-                    let shard = state.shard_index(position);
+                    let shard = state.shard_index(shard_position);
                     let send = match client_request.route().shard() {
                         Shard::Direct(s) => {
                             shards_sent = 1;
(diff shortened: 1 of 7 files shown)


pgdog/src/backend/pool/connection/binding.rs
```diff
@@ -103,15 +103,15 @@ impl Binding {
...
+                        for (position, server) in shards.iter_mut().enumerate() {
                             if !server.has_more_messages() {
                                 continue;
                             }
 
                             let message = server.read().await?;
 
                             read = true;
-                            if let Some(message) = state.forward(message)? {
+                            if let Some(message) = state.forward_from(position, message)? {
```
levkk: nit: rename `position`  to `shard`

pgdog/src/backend/pool/connection/buffer.rs
```diff
@@ -20,12 +20,212 @@ use pgdog_postgres_types::Datum;
 
 use super::Aggregates;
 
+#[derive(Debug, Clone)]
+struct OrderByMergeState {
```
levkk: nit: would be good to move that to its own file.

levkk:
Looks good. We would need quite a few integration tests for this one though before it's in a mergeable state.

murex971:
> We would need quite a few integration tests for this one though before it's in a mergeable state.

could you elaborate what kind of integration tests should be there ?

levkk:
I'm thinking, use our existing integration test setup with:

- 2 shards
- insert a bunch of rows in each table, at random
- SELECT with a bunch of different ordering
- assert that the order is always correct

Repeat this test with as many data types as you can think of; totally fine to use the same table, can just order on different columns, e.g. varchar, bigint, timestamp(tz), etc. Let me know if you hit any errors anywhere, we need to double check our decoding logic for some more complex types.

Could write them in `integration/rust` project if convenient.

pgdog/src/backend/pool/connection/binding.rs
murex971: since it's not shard_id, rather index in connected servers list, `shard` could be misleading.
would change this to `shard_position`

murex971:
@levkk 
While validating the new cross-shard k-way merge path for ORDER BY, I found an existing limitation, when an ORDER BY key is not present in the projected output (e.g. SELECT id ... ORDER BY name), PgDog currently resolves sort columns from `RowDescription` and may silently skip unresolved named keys, leading to incorrect global ordering. This behavior existed in the previous full-buffer sort path as well; the k-way merge implementation inherits it. For this PR, tests are scoped to queries where sort keys are included in the select list to validate merge correctness. We should open a follow-up to support PostgreSQL-compatible ordering on non-projected keys (likely via hidden helper columns in shard queries and dropping them before returning results, or by returning an explicit error instead of silent fallback).

levkk:
@murex971 Thank you. Yes, this is a known limitation: https://docs.pgdog.dev/features/sharding/cross-shard-queries/select/#sorting. I agree, we should do a follow up and fix it; I'm thinking we use the same approach as we do for aggregates: add the columns temporarily into the query using the rewriter and remove them when sending the result back to the client.

levkk:
My apologies for delay on reviewing this. Will try to get through this by tomorrow or early next week. Please ping me again if you don't hear back in this timeframe.

murex971:
added an issue for this https://github.com/pgdogdev/pgdog/issues/1135

murex971:
@levkk could you take a look now ?

levkk:
Sorry for the delay here. We're deep in the parser rewrite at the moment, it's almost done. We should be able to take a good look (and probably merge this) next week. Thanks for your patience!

PR #1115, opened 2026-06-29
chore: normalize configs for nextest and rust integration tests

meskill:
- more explicit nextest config with profile for integration tests
- fix slow unit test (slow on wsl)
- rename integration tests package rust -> integration_tests_rust
- dedupe some dependencies and make lock file smaller

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,7 +1,12 @@
-[profile.dev]
-fail-fast = true
-test-threads = 1
-default-filter = "not package(rust)"
-
 [profile.default]
-slow-timeout = "15s"
+slow-timeout = { period = "15s" }
+fail-fast = true
+# ignore integration tests since they require a running pgdog instance
+default-filter = "not package(rust)"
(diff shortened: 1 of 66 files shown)


sgrif:
> fix slow unit test (slow on wsl)

I know which test you're talking about, but it's not clear to me how this PR fixes it. Can you clarify?

meskill:
> > fix slow unit test (slow on wsl)
> 
> I know which test you're talking about, but it's not clear to me how this PR fixes it. Can you clarify?

Sure, it's this test https://github.com/pgdogdev/pgdog/blob/main/pgdog/src/backend/databases.rs#L1861 which uses non existing hosts in config (I think we connect to the hosts during the cutover in config) and it's trying to connect with a long timeout on my machine with WSL. I switched to ip addresses to drop any dns resolutions, and that removed long waiting

sgrif (approved):
I love that I'm not the only dev on Windows for once lol

PR #1119, opened 2026-06-30
fix: skip multi-statement SET validation in session mode

### LINKED ISSUE #1087: Session mode incorrectly rejects multi-statement queries mixing SET with other commands
**PgDog version**
`v0.1.44`

**Description**
In session mode, PgDog should pass queries directly to PostgreSQL without parsing or validation. However, multi-statement queries that mix SET with other commands (e.g. SHOW) are rejected with an error even when the pool is confirmed to be in session mode via SHOW POOLS.

**To reproduce**
```bash
psql -U <user> -h localhost -d <database> \
  -c "SET DateStyle='ISO';SET extra_float_digits = 2;show transaction_isolation"
```

Expected: Query succeeds (as it does when connecting directly to PostgreSQL)

Actual:
```
ERROR:  multi-statement queries cannot mix SET with other commands
DETAIL:  SET DateStyle='ISO';SET extra_float_digits = 2;show transaction_isolation
```

**Impact**

eg psqlODBC sends this multi-statement batch during connection startup, making it impossible to connect via ODBC even with pooler_mode = "session".

**Where does it come from**

The check is done in `src/frontend/router/parser/query/set.rs#try_multi_set`. This should take pool mode into consideration and return early in `session` mode.

DirtyRacer1337:
just cherry pick from this https://github.com/pgdogdev/pgdog/commit/923523478cc09ace45d373f523a7f734724047a6

In session mode, PgDog should forward multi-statement queries to the server verbatim without validating whether they mix SET with other commands. The server owns the session and can handle these naturally.

Fixes a bug where queries like:
  SET DateStyle='ISO'; show transaction_isolation

were rejected with 'multi-statement queries cannot mix SET with other commands' even when the pool was in session mode.

Closes #1087

### DIFF
--- a/pgdog/src/frontend/router/parser/context.rs
+++ b/pgdog/src/frontend/router/parser/context.rs
@@ -149,4 +149,14 @@ impl<'a> QueryParserContext<'a> {
     pub(super) fn expanded_explain(&self) -> bool {
         self.expanded_explain
     }
+
+    /// Are we running in session mode?
+    ///
+    /// In session mode, queries are forwarded to the server without
+    /// parsing or validation beyond what's required for routing.
+    pub(super) fn is_session_mode(&self) -> bool {
+        self.router_context
+            .cluster
+            .pooler_mode() == crate::config::PoolerMode::Session
+    }
 }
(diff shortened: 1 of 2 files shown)


levkk:
Just a quick `cargo fmt` (and maybe a small test if you're feeling generous in `parser/test/mod.rs`).

DirtyRacer1337:
I'm not a rust dev, sorry

zefixlluja:
#1157 includes

* cherry-picked commit
* `cargo fmt` on top
* and test case

PR #1121, opened 2026-07-01
fix(replication): prevent premature cutover due to lag miscalculation

meskill:
Fix the issue when `replication_lag` is called before actual replication lag is calculated for all or some shards


levkk (approved):
Nice catch

PR #1122, opened 2026-07-01
fix: replica lag evalution in load balancer

levkk:
- fix: the primary/replica evaluation order was inverted, so the replica ban never triggered
- fix: `prefer_primary` incorrectly excluded primary when all replicas were banned @stewart-glabs :pray: 
- feat: show the reason why pools are unbanned (helps debugging)

### DIFF
--- a/pgdog/src/backend/pool/lb/ban.rs
+++ b/pgdog/src/backend/pool/lb/ban.rs
@@ -1,6 +1,6 @@
 use super::*;
 use parking_lot::RwLock;
-use std::time::Instant;
+use std::{fmt::Display, time::Instant};
 
 use tracing::{error, warn};
 
@@ -9,6 +9,24 @@ use tracing::{error, warn};
 pub struct Ban {
     inner: Arc<RwLock<BanInner>>,
     pool: Pool,
+}
+
+#[derive(Debug, Default, Copy, Clone)]
+pub enum UnbanReason {
+    #[default]
+    AllTargetsBanned,
+    Expired,
+    Manual,
+}
+
+impl Display for UnbanReason {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::AllTargetsBanned => write!(f, "all targets banned"),
+            Self::Expired => write!(f, "expired"),
+            Self::Manual => write!(f, "manual"),
+        }
+    }
 }
 
 impl Ban {
@@ -48,7 +66,7 @@ impl Ban {
     }
 
     /// Unban the database.
-    pub fn unban(&self, manual_check: bool) {
+    pub fn unban(&self, manual_check: bool, reason: UnbanReason) {
         let mut guard = self.inner.upgradable_read();
         if let Some(ref ban) = guard.ban {
             if ban.error != Error::ManualBan || !manual_check {
@@ -56,7 +74,7 @@ impl Ban {
                     guard.ban = None;
                 });
             }
-            warn!("resuming read queries [{}]", self.pool.addr());
+            warn!("resuming read queries: {} [{}]", reason, self.pool.addr());
         }
     }
 
@@ -114,7 +132,11 @@ impl Ban {
         };
         drop(guard);
         if unbanned {
-            warn!("resuming read queries [{}]", self.pool.addr());
+            warn!(
+                "resuming read queries: {} [{}]",
+                UnbanReason::Expired,
+                self.pool.addr()
+            );
         }
         unbanned
     }
@@ -185,7 +207,7 @@ mod tests {
         let pool = Pool::new_test();
         let ban = Ban::new(&pool);
         ban.ban(Error::ServerError, Duration::from_secs(1));
-        ban.unban(false);
+        ban.unban(false, UnbanReason::Expired);
         assert!(!ban.banned());
         assert!(ban.error().is_none());
     }
@@ -195,7 +217,7 @@ mod tests {
         let pool = Pool::new_test();
         let ban = Ban::new(&pool);
         ban.ban(Error::ManualBan, Duration::from_secs(1));
-        ban.unban(true);
+        ban.unban(true, UnbanReason::Expired);
         assert!(ban.banned());
         assert_eq!(ban.error(), Some(Error::ManualBan));
     }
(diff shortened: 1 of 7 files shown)


pgdog/src/backend/pool/lsn_monitor.rs
```diff
@@ -481,6 +481,48 @@ mod test {
...
+        .into()
+    }
+
+    // Regression test: the shard monitor used to call `replica_lag` with
+    // `self` and `primary` swapped, which inverted both subtractions inside
+    // (`primary.lsn - self.lsn`, `primary.timestamp - self.timestamp`). The
+    // result was negative bytes and a duration clamped to 0, so the replica
+    // always reported zero lag and `ban_replica_lag` could never trigger.
+    #[test]
+    fn test_replica_lag_direction() {
```
meskill: I'd like to note that the unit test doesn't validate the actual bug that was present in monitor.

would it better to put the logic from monitor to the method on replica type, so we hide the details and can have some test for it?

levkk: Yes definitely. Haha, that's what I get for using GLM 5.2 😆

pgdog/src/backend/pool/lb/ban.rs
```diff
@@ -11,6 +11,24 @@ pub struct Ban {
     pool: Pool,
 }
 
+#[derive(Debug, Default, Copy, Clone)]
+pub enum UnbanReason {
+    #[default]
```
meskill: I don't think this enum should have default, it looks arbitrary and error-prone in usage

pgdog/src/backend/pool/lb/ban.rs
```diff
@@ -48,15 +66,15 @@ impl Ban {
...
+    pub fn unban(&self, manual_check: bool, reason: UnbanReason) {
         let mut guard = self.inner.upgradable_read();
         if let Some(ref ban) = guard.ban {
             if ban.error != Error::ManualBan || !manual_check {
                 guard.with_upgraded(|guard| {
                     guard.ban = None;
                 });
             }
-            warn!("resuming read queries [{}]", self.pool.addr());
+            warn!("resuming read queries: {} [{}]", reason, self.pool.addr());
```
meskill: the log is unconditional though code above can ignore unban call

pgdog/src/backend/pool/shard/monitor.rs
```diff
@@ -168,6 +193,84 @@ mod test {
...
+        let replica = lsn_stats(true, 200, "2026-07-01 13:33:10.000000+00");
+
+        let lag = calculate_replica_lag(&primary, &replica);
+
+        assert_eq!(lag.bytes, 0);
+        assert_eq!(lag.duration, Duration::default());
+    }
+
+    #[test]
+    #[should_panic(expected = "primary stats must come from the primary pool")]
```
meskill, line 238: this test relies on `debug_assert` inside the tested code. 

probably should also have `#[cfg(debug_assertions)]` to not fail with `test --release`

pgdog/src/backend/pool/lb/ban.rs
```diff
@@ -114,7 +132,11 @@ impl Ban {
         };
         drop(guard);
         if unbanned {
-            warn!("resuming read queries [{}]", self.pool.addr());
+            warn!(
+                "resuming read queries: {} [{}]",
+                UnbanReason::Expired,
```
meskill, line 146: I though there is some issues since the `UnbanReason::Expired` is never passed to `unban` except tests, but it's actually used here. 
So this some kind of another code-smell: a little code duplication.
Should we merge the `unban` and `unban_expired` together? or maybe other way around: private unban (upgrade + log) with unban_manual, unban_expired with different prechecks?

pgdog/src/backend/pool/lb/mod.rs
```diff
@@ -349,12 +350,16 @@ impl LoadBalancer {
...
+            // we read from the primary if we have no replicas
+            ExcludePrimary => !candidates
                 .iter()
                 .any(|target| matches!(target.role(), Role::Replica | Role::Auto)),
+            // PreferPrimary makes all queries writes. If a query lands here,
+            // it's because of pgdog.role=replica. Let it use the primary only if
+            // no replicas are available.
+            PreferPrimary => !candidates.iter().any(|target| {
+                matches!(target.role(), Role::Replica | Role::Auto) && !target.ban.banned()
+            }),
```
meskill, line 362: well, I'd say not very simple function to understand because of juggling with different conditions and flags.

Just feel strange that we check for banned in some cases only for primary_reads flag and then shuffling and sorting the list of replicas that could consist of only banned replicas just to eventually call unban after we traverse and verify it again

meskill (changes requested):

pgdog/src/backend/pool/shard/monitor.rs
levkk: Haha I guess...but who runs their tests in release mode? Who has that kind of patience :D

pgdog/src/backend/pool/lb/mod.rs
meskill: just rant, maybe ignore..

levkk: Yeah, it's weird. I feel like the entire LB logic is one function. I think it used to work ok when we had just a couple of algorithms, but it's getting more interesting for sure and `PreferPrimary` is assuming the caller did something special to get here, so definitely not straightforward to understand.

pgdog/src/backend/pool/lb/ban.rs
levkk: Oh oops, nice catch.

pgdog/src/backend/pool/shard/monitor.rs
meskill: you don't do `test --release` until the moment you doubt you have identical behavior in dev/prod :D

pgdog/src/backend/pool/lb/mod.rs
meskill: looks like it have to stay a "secret sauce" for now

pgdog/src/backend/pool/shard/monitor.rs
levkk: been there, done that...

pgdog/src/backend/pool/lb/ban.rs
levkk: Yeah...maybe. `unban` really means `unban_unless_manual` and `unban_if_expired` means what it means, and runs every 333ms from the shard monitor. `unban` really means unban because the caller really needs it to happen, e.g., all targets are banned.

Yeah, unban reason right now is only used for logging. I agree, we should merge the logic and use the enum as the operand.

```rust
// FIXME(lev): do that
```

meskill (approved):

PR #1123, opened 2026-07-01
fix(replication): update the processed lsn on keep-alive messages

meskill:
Make sure the commited_lsn updated on keep-alive messages in order to prevent stuck replication lag value if no updates to the original publication were made, but other changes advances the global lsn on the source


levkk (approved):
Really getting into the details now, awesome 🚀

PR #1125, opened 2026-07-01
Port QueryParser::has_locking_clause to new parser

sgrif:
The old version of this function failed to detect locks anywhere outside of a CTE or top level query. As part of the port, I've fixed that behavior.


levkk (approved):
Mmmyexcellent

PR #1126, opened 2026-07-01
Remove a bunch of dead code in `parser::function`

sgrif:
`locking_behavior` was never read, and all the logic around lock functions here was redundant with our checks for locking functions that we do elsewhere. The lists we were searching were far to small to justify the complexity of `Lazy<HashMap<...>>`


pgdog/src/frontend/router/parser/query/test/mod.rs
```diff
@@ -562,9 +562,9 @@ fn test_show_shards() {
 
 #[test]
 fn test_write_functions() {
-    let route = query!("SELECT pg_advisory_lock($1)");
+    let route = parse!("SELECT pg_advisory_lock($1)", &["1".as_bytes()]);
     assert!(route.is_write());
-    assert!(!route.is_lock_session()); // Need Bind
+    assert!(route.is_lock_session());
```
levkk: I think this one was to make sure `Parse` doesn't actually lock the session,e.g., when statement is just prepared, but not executed yet, so we checked for `Bind`.

pgdog/src/frontend/router/parser/function.rs
```diff
@@ -1,55 +1,18 @@
-use std::collections::{HashMap, HashSet};
-
-use once_cell::sync::Lazy;
 use pg_query::{Node, NodeEnum, protobuf};
 
-static WRITE_ONLY: Lazy<HashMap<&'static str, LockingBehavior>> = Lazy::new(|| {
-    HashMap::from([
-        ("pg_advisory_lock", LockingBehavior::Lock),
```
levkk, line 8: This moved to the new parser I assume?

sgrif: No, it just seemed redundant with https://github.com/pgdogdev/pgdog/blob/main/pgdog/src/frontend/router/parser/query/select.rs#L56 which already walks the AST looking for advisory locks

pgdog/src/frontend/router/parser/query/test/mod.rs
sgrif: This wasn't testing that though, `query!` was constructing a `Query` message, which does execute. The only reason `!route.is_lock_session()` was the case before was because this is an invalid query. If this test were changed to `SELECT pg_advisory_lock(1)`, `assert!(!route.is_lock_session())` would fail.

pgdog/src/frontend/router/parser/function.rs
levkk: Might of been left over from that refactor? I don't remember tbh. If you say it's dead, it's dead. I think we even have advisory lock test coverage somewhere.

sgrif: Yes, we do have advisory lock test coverage. The line I linked above is where we actually get the advisory locks in order to handle them. These entries were only used to determine if the route should be read or write. The logic all these entries were encapsulating has been replaced with https://github.com/pgdogdev/pgdog/pull/1126/changes#diff-9c64b05c3c85bff899f33445482dfe110e255cbb704e17632dcce59be9b9ed7bR64

The only change in behavior here would be if we received a query such as `SELECT pg_advisory_lock($1)` with no binds, or other forms that we can't handle. We discussed last week that it's unclear what the behavior should be in those cases anyway. The only test coverage that cared about this change was the one test you commented on above, which as I pointed out, wasn't testing what it thought it was.

sgrif: (Note: Please feel free to tell me if I'm missing something on the intended behavior here. I wouldn't be surprised to learn there's some subtle interaction I'm missing, but given the only test asserting this behavior was a single test that seemed to be testing a condition which should never exist, and the sheer amount of dead code this module had, whatever is going on seems to be by coincidence not intent)

levkk: As long as you remove all getters for the stuff you're deleting, you should be good. Tbh, I don't remember when we moved the handling for advisory locks into the StatementParser and what happened to this code.

levkk: If you find a bug along the way, even better.

sgrif: Oh it's definitely extremely buggy. The current code just assumes the first function call it finds is what defines the behavior for the whole query (presumably because it was written assuming `SELECT single_func_call` was the only thing we'd receive)

PR #1130, opened 2026-07-02
feat: Hashicorp Vault static role credentials

larrywax:
Hey, it's me again 😃 

## Summary
- Add Vault static role authentication, for both client-side password verification (`client_vault_path`) and backend PgDog-to-Postgres connections (`server_auth = "vault_static"` + `backend_vault_path`)
- Rename `server_auth = "vault"` → `"vault_dynamic"` and `vault_path` → `backend_vault_path` for clarity/symmetry with `client_vault_path`. Both old names still parse via `#[serde(alias = ...)]`, no config breaking change
- Backend static-role credentials behave like RDS IAM / Azure Workload Identity: username is operator-supplied (`server_user`/`name`), not generated by Vault, and there's no proactive percentage-based refresh
- Prevent a tight refresh loop: Vault only issues a new static-role password once the old one's TTL truly expires and echoes back the same password (shrinking TTL) if read early. Scheduling the refresh shortly before expiry (like RDS/Azure) caused continuous refetching once TTL dropped under the buffer. Static backend credentials set their own `refresh_at`.
- Extract shared Vault login/token-cache/HTTP-fetch logic into `pgdog/src/auth/vault.rs`, reused by both `pgdog/src/backend/auth/vault.rs` and `pgdog/src/frontend/client/mod.rs`

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,7 +291,7 @@ When any user has `server_auth = "vault_dynamic"` or `"vault_static"`, the follo
 
 #### HashiCorp Vault static role authentication
 
-Unlike dynamic credentials, a Vault static database role has a fixed username and only its password rotates, on a schedule Vault manages. PgDog supports two independent uses of a static role,they don't need to point at the same role, and each has its own username setting:
+Unlike dynamic credentials, a Vault static database role has a fixed username and only its password rotates, on a schedule Vault manages. PgDog supports two independent uses of a static role, they don't need to point at the same role, and each has its own username setting:
 
 - `vault_path`: verify the password a client sends to PgDog against Vault's current password for the role, instead of a statically configured password.
 - `server_auth = "vault_static"` with `server_vault_path`: use the role's Vault-managed password for PgDog-to-PostgreSQL connections. Unlike `vault_dynamic`, PgDog doesn't take the username from Vault, it connects as `server_user` or `name`, if `server_user` isn't set.
--- a/README.md
+++ b/README.md
@@ -291,7 +291,7 @@ When any user has `server_auth = "vault_dynamic"` or `"vault_static"`, the follo
 
 #### HashiCorp Vault static role authentication
 
-Unlike dynamic credentials, a Vault static database role has a fixed username and only its password rotates, on a schedule Vault manages. PgDog supports two independent uses of a static role — they don't need to point at the same role, and each has its own username setting:
+Unlike dynamic credentials, a Vault static database role has a fixed username and only its password rotates, on a schedule Vault manages. PgDog supports two independent uses of a static role,they don't need to point at the same role, and each has its own username setting:
 
 - `vault_path`: verify the password a client sends to PgDog against Vault's current password for the role, instead of a statically configured password.
 - `server_auth = "vault_static"` with `server_vault_path`: use the role's Vault-managed password for PgDog-to-PostgreSQL connections. Unlike `vault_dynamic`, PgDog doesn't take the username from Vault, it connects as `server_user` or `name`, if `server_user` isn't set.
--- a/README.md
+++ b/README.md
@@ -198,12 +198,13 @@ role = "auto"
 
 &#128216; **[Authentication](https://docs.pgdog.dev/features/authentication/)**
 
-PgDog supports four authentication methods:
+PgDog supports five authentication methods:
 
 1. Password-based
 2. AWS RDS IAM
 3. Azure Workload Identity
 4. HashiCorp Vault dynamic credentials
+5. HashiCorp Vault static role credentials
 
 #### Password-based authentication
 
@@ -253,7 +254,7 @@ When any user has `server_auth = "azure_workload_identity"`, the following setti
 - `tls_verify` must **not** be `"disabled"`.
 - `passthrough_auth` must be `"disabled"`.
 
-#### HashiCorp Vault authentication
+#### HashiCorp Vault dynamic role authentication
 
 PgDog can fetch dynamic database credentials (username and password) from HashiCorp Vault's database secrets engine, while keeping client-to-PgDog authentication unchanged. Credentials are cached and rotated automatically after a configured percentage of the Vault lease has elapsed.
 
@@ -287,6 +288,32 @@ When any user has `server_auth = "vault_dynamic"` or `"vault_static"`, the follo
 
 - `tls_verify` must **not** be `"disabled"`.
 - `passthrough_auth` must be `"disabled"`.
+
+#### HashiCorp Vault static role authentication
+
+Unlike dynamic credentials, a Vault static database role has a fixed username; only its password rotates, on a schedule Vault manages. PgDog supports two independent uses of a static role, and they can be combined on the same user:
+
+- `client_vault_path`: verify the password a client sends to PgDog against Vault's current password for the role, instead of a statically configured password.
+- `server_auth = "vault_static"` with `backend_vault_path`: use the role's Vault-managed username and password for PgDog-to-PostgreSQL connections, refreshed like dynamic credentials.
+
+**Example**
+
+In `users.toml`:
+
+```toml
+[[users]]
+name = "alice"
+database = "pgdog"
+client_vault_path = "database/static-creds/pgdog-static-role"
+server_auth = "vault_static"
+backend_vault_path = "database/static-creds/pgdog-static-role"
+# Refresh the backend password after 80% of the rotation TTL (default).
+# vault_refresh_percent = 80
+```
+
+In `pgdog.toml`, the same `[vault]` section used for dynamic credentials applies.
+
+Both settings are optional and independent: set only `client_vault_path` to verify client passwords while keeping any other backend authentication method, or only `server_auth = "vault_static"` to use the static role for backend connections while clients authenticate with a regular password.
 
 ### Sharding
 
--- a/README.md
+++ b/README.md
@@ -294,7 +294,7 @@ When any user has `server_auth = "vault_dynamic"` or `"vault_static"`, the follo
 Unlike dynamic credentials, a Vault static database role has a fixed username and only its password rotates, on a schedule Vault manages. PgDog supports two independent uses of a static role — they don't need to point at the same role, and each has its own username setting:
 
 - `client_vault_path`: verify the password a client sends to PgDog against Vault's current password for the role, instead of a statically configured password.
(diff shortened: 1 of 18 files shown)


pgdog-config/src/users.rs
```diff
@@ -179,15 +189,22 @@ pub enum ServerAuth {
     RdsIam,
     /// Generate an Azure Workload Identity auth token per connection attempt.
     AzureWorkloadIdentity,
-    /// Fetch dynamic credentials from HashiCorp Vault.
-    Vault,
+    /// Fetch dynamic credentials from HashiCorp Vault (database secrets engine).
+    /// Vault generates a new username and password on each lease.
+    ///
+    /// **Note:** `"vault"` is accepted as a deprecated alias for backward compatibility.
```
levkk: Don't think anyone used this yet so I wouldn't worry about it.

pgdog-config/src/users.rs
```diff
@@ -282,13 +307,22 @@ pub struct User {
...
-    /// Vault path to fetch dynamic database credentials from, e.g. `database/creds/my-role`.
-    /// Required when `server_auth` is set to `vault`.
-    pub vault_path: Option<String>,
+    /// Vault path used to fetch backend (server-side) database credentials,
+    /// e.g. `database/creds/my-role` for `server_auth = "vault_dynamic"` or
+    /// `database/static-creds/my-role` for `server_auth = "vault_static"`.
+    ///
+    /// **Note:** `"vault_path"` is accepted as a deprecated alias for backward compatibility.
+    #[serde(alias = "vault_path")]
+    pub backend_vault_path: Option<String>,
```
levkk: Let's keep naming consistent with other settings, e.g. `server_vault_path` for server settings and `vault_path` for client settings perhaps?

pgdog/src/auth/vault.rs
```diff
@@ -0,0 +1,630 @@
...
+    }
+
+    let auth: AuthResponse = response
+        .json()
+        .await
+        .map_err(|err| error(format!("invalid Vault login response: {}", err)))?;
+
+    Ok(VaultToken {
+        token: auth.auth.client_token,
+        expires_at: SystemTime::now() + Duration::from_secs(auth.auth.lease_duration),
```
levkk: You kind of want to use `Instant` here because its guaranteed to never be affected but clock skew.

levkk:
Is the static vs dynamic credentials a known Vault thing that we can point people to, e.g., in their docs? Just trying to understand if this is something you guys developed internally or is more of an "industry standard" (quotes for dramatic effect, I don't think Vault is that ubiquitous).

larrywax:
> Is the static vs dynamic credentials a known Vault thing that we can point people to, e.g., in their docs? Just trying to understand if this is something you guys developed internally or is more of an "industry standard" (quotes for dramatic effect, I don't think Vault is that ubiquitous).

Both are native Vault features and are documented here:
- https://developer.hashicorp.com/vault/docs/secrets/databases
- https://developer.hashicorp.com/vault/docs/secrets/databases/postgresql

levkk:
Sounds good. Will review asap

pgdog/src/auth/vault.rs
larrywax: Are you sure you meant here for the vault auth token, or did you actually mean in the token cache?

levkk: Good question. So, if the expiration date is external - agreed, let's use system time or parsed time. If it's our own, e.g., we pull a duration from somewhere and add it to `now()`, let's use `Instant` because that will guarantee, in the unlikely scenario someone updates the clock, that the token cache won't blow up.

pgdog-config/src/users.rs
```diff
@@ -88,18 +88,28 @@ impl Users {
...
+                        r#"user "{}" (database "{}") uses "server_auth" = "{:?}" but "backend_vault_path" is not set"#,
+                        user.name, user.database, user.server_auth
                     );
                 }
 
                 if config.vault.is_none() {
                     warn!(
-                        r#"user "{}" (database "{}") uses "server_auth" = "vault" but the [vault] section is missing from pgdog.toml"#,
-                        user.name, user.database
+                        r#"user "{}" (database "{}") uses "server_auth" = "{:?}" but the [vault] section is missing from pgdog.toml"#,
```
levkk: Let's use `Display` here, looks better.

pgdog-config/src/users.rs
```diff
@@ -88,18 +88,28 @@ impl Users {
...
+
+            if matches!(
+                user.server_auth,
+                ServerAuth::VaultDynamic | ServerAuth::VaultStatic
+            ) {
+                if user.backend_vault_path.is_none() {
                     warn!(
-                        r#"user "{}" (database "{}") uses "server_auth" = "vault" but "vault_path" is not set"#,
-                        user.name, user.database
+                        r#"user "{}" (database "{}") uses "server_auth" = "{:?}" but "backend_vault_path" is not set"#,
```
levkk: Same as below.

pgdog/src/frontend/client/mod.rs
```diff
@@ -160,6 +160,30 @@ impl Client {
             return Ok(AuthResult::NoPasswordConfig);
         }
 
+        // Resolve any Vault static role entries to plaintext before starting
+        // the auth exchange.  MD5, SCRAM, and plain all need the actual
+        // password, so this must happen before the first message is sent.
+        let resolved;
```
levkk: I understood what happens here, but maybe there is a cleaner way to write this? e.g., if vault, do x, else do y, or maybe move this into its own testable method?

README.md
```diff
@@ -283,11 +284,36 @@ kubernetes_role = "pgdog"
...
 
-When any user has `server_auth = "vault"`, the following settings must be configured as well:
+When any user has `server_auth = "vault_dynamic"` or `"vault_static"`, the following settings must be configured as well:
 
 - `tls_verify` must **not** be `"disabled"`.
 - `passthrough_auth` must be `"disabled"`.
 
+#### HashiCorp Vault static role authentication
+
+Unlike dynamic credentials, a Vault static database role has a fixed username and only its password rotates, on a schedule Vault manages. PgDog supports two independent uses of a static role — they don't need to point at the same role, and each has its own username setting:
```
levkk: m-dashes are an AI code smell these days - ok to have them in rustdoc, but in the readme, let's try to avoid them. I know...times we live in.

levkk (commented):
Looks good, just a few naming convention updates and a bit of code cleanup. Happy to merge this afterwards.

We should also start running tests for this in CI (i.e. integration tests with actual vault server) if we can. Maybe as a follow up PR, just to make sure we don't break anything here going forward. This is definitely more complex than RDS IAM and friends.

levkk:
Also, I just found this: https://docs.rs/vaultrs/latest/vaultrs/

Any reason to not use an existing vault client and implement our own? I get that it's just an HTTP API with JWT, so pretty simple, but still worth considering.

pgdog/src/frontend/client/mod.rs
larrywax: Moved the logic in the `auth::vault` crate and moved the call before `check_password`
Should look way cleaner now

larrywax:
> Also, I just found this: https://docs.rs/vaultrs/latest/vaultrs/
> 
> Any reason to not use an existing vault client and implement our own? I get that it's just an HTTP API with JWT, so pretty simple, but still worth considering.

Didn't want to import an entire library for just two endpoint calls, but happy to switch to it if you prefer 👍

pgdog/src/auth/vault.rs
larrywax: Changed the auth token and `refresh_at` cache field to `Instant` 👍 
I made the `expire_at` field, used by RDS and Azure, optional to avoid this cumbersome logic in the vault code
```
expire_at: SystemTime::now() + ttl
refresh_at: Instant:now() + ttl
```
Let me know what you think

levkk:
> > Also, I just found this: https://docs.rs/vaultrs/latest/vaultrs/
> > Any reason to not use an existing vault client and implement our own? I get that it's just an HTTP API with JWT, so pretty simple, but still worth considering.
> 
> Didn't want to import an entire library for just two endpoint calls, but happy to switch to it if you prefer 👍

All good. I just wanted to confirm

pgdog/src/auth/vault.rs
```diff
@@ -0,0 +1,795 @@
...
+/// passed through unchanged.  Failed Vault fetches are logged and skipped so
+/// that any remaining plain/hashed passwords can still authenticate the client.
+pub(crate) async fn resolve_passwords(passwords: &[PasswordKind]) -> Vec<PasswordKind> {
+    let mut buf = Vec::with_capacity(passwords.len());
+    for p in passwords {
+        match p {
+            PasswordKind::VaultStaticRole(path) => match static_client_password(path).await {
+                Ok(pw) => buf.push(PasswordKind::Plain(pw)),
+                Err(err) => {
+                    warn!(vault_path = path, %err, "failed to resolve Vault password, skipping")
```
levkk, line 283: nit: if you want this to be more debuggable, you should send this error to the client that way it appears in the logs / APM instead of just in pgdog logs

README.md
```diff
@@ -283,11 +284,36 @@ kubernetes_role = "pgdog"
...
 
-When any user has `server_auth = "vault"`, the following settings must be configured as well:
+When any user has `server_auth = "vault_dynamic"` or `"vault_static"`, the following settings must be configured as well:
 
 - `tls_verify` must **not** be `"disabled"`.
 - `passthrough_auth` must be `"disabled"`.
 
+#### HashiCorp Vault static role authentication
+
+Unlike dynamic credentials, a Vault static database role has a fixed username and only its password rotates, on a schedule Vault manages. PgDog supports two independent uses of a static role,they don't need to point at the same role, and each has its own username setting:
```
levkk:
```suggestion
Unlike dynamic credentials, a Vault static database role has a fixed username and only its password rotates, on a schedule Vault manages. PgDog supports two independent uses of a static role, they don't need to point at the same role, and each has its own username setting:
```

levkk (approved):
Awesome, thanks!

pgdog/src/auth/vault.rs
levkk: Feel free to send a separate PR for this - it's okay for now, but you will want this in the future I suspect once you start integration testing on your end.

larrywax:
Thank you so much! I'll send a PR for integration testing your way shortly

PR #1131, opened 2026-07-02
fix(replication): replace custom AbortToken with tokio::CancellationToken

meskill:
Replace custom AbortSignal with CancellationToken to make the cancellation explicit and not relied on futures drop (that holds rx part of channel). Pass cancellation token for graceful cancellation. Early cancellation on errors


pgdog/src/backend/replication/logical/publisher/parallel_sync.rs
```diff
@@ -202,20 +188,24 @@ impl ParallelSyncManager {
...
+    pub async fn run(self, cancel: CancellationToken) -> Result<Vec<Table>, Error> {
         info!(
             "starting parallel table copy using {} replicas and {} parallel copies",
             self.replicas.len(),
             self.permit.available_permits() / self.replicas.len(),
         );
 
+        // Create a child cancel token with the guard to cancel the handles below
+        // in case any of it fails without affecting the parent task.
+        // If every handle succeed the guard token will just cancel already finished work
```
sgrif:
```suggestion
        // If every handle succeeds the guard token will just cancel already finished work
```

pgdog/src/backend/replication/logical/publisher/parallel_sync.rs
```diff
@@ -202,20 +188,24 @@ impl ParallelSyncManager {
...
+        let cancel = cancel.child_token();
```
sgrif, line 201: I think we should require the caller to call `.child_token()` if they don't want to be cancelled.
```suggestion
```

pgdog/src/backend/replication/logical/publisher/parallel_sync.rs
```diff
@@ -202,20 +188,24 @@ impl ParallelSyncManager {
...
+        let _guard = cancel.clone().drop_guard();
```
sgrif:
```suggestion
        let _guard = cancel.drop_guard_ref();
```

pgdog/src/backend/replication/logical/publisher/publisher_impl.rs
```diff
@@ -431,17 +439,16 @@ impl Publisher {
             };
 
             let dest = dest.clone();
-            set.spawn(async move {
+            let cancel = cancel.clone();
+            syncs.push(async move {
                 let manager = ParallelSyncManager::new(tables, replicas, dest)?;
-                let tables = manager.run().await?;
+                let tables = manager.run(cancel).await?;
```
sgrif, line 447:
```suggestion
            syncs.push(async move {
let manager = ParallelSyncManager::new(tables, replicas, dest)?;
                let tables = manager.run(cancel.child_token()).await?;
```

pgdog/src/backend/replication/logical/publisher/publisher_impl.rs
```diff
@@ -431,17 +439,16 @@ impl Publisher {
             };
 
             let dest = dest.clone();
-            set.spawn(async move {
+            let cancel = cancel.clone();
```
sgrif, line 444: This is the first place we actually do anything with a cancellation token other than pass it through. Should we remove everything up the chain from here and the `CancellationToken::new()`s, and create it at this point?

pgdog/src/backend/replication/logical/publisher/parallel_sync.rs
meskill: I think it's better to leave it here, in the `ParallelSyncManager`. In the `run` method we start multiple `ParallelSync` instances to copy tables in parallel. So, basically ParallelSyncManager manager the parallel work here and the cancel token is here to help manage it: we create a child_token with guard to not affect parent or sibling work, but still be affected if the parent is cancelled. The guard is needed to cancel all the other running syncs in case any of them fails: in that case we cancel this execution subtree and propagate the error to parent where the parent will react on error and finish with the error eventually and not cancellation.

pgdog/src/backend/replication/logical/publisher/publisher_impl.rs
meskill: The idea here, that the initial CancellationToken comes from the https://github.com/pgdogdev/pgdog/blob/02c7a327c50e1fab0ed9683b908dd8c9ce9df0ce/pgdog/src/api/async_task.rs#L358-L361
so the whole task becomes manageable from there. We create the token in the root and pass it down to any child task. That's why it passed here and just cloned or child_token is created.

sgrif (approved):

PR #1132, opened 2026-07-02
Port StatementRewrite::sharding_key_update_check to the new parser

sgrif:
There's quite a bit of churn in this one, as I needed to change StatementRewrite to always have the new AST so every test that was previously passing `None` needed to be updated. One file in particular had a *bunch* of repeated code in its test, which I extracted to a helper function.

The other bit of funkiness here is that since the return value of this function is actually passed to later functions, and I am still trying to break this port up into pieces, we deparse and reparse to convert to the equivalent node in the old parser

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3299,7 +3299,7 @@ dependencies = [
 [[package]]
 name = "pg_raw_parse"
 version = "0.1.0"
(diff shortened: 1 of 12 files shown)


levkk (approved):
kewl

PR #1136, opened 2026-07-03
Port StatementRewrite::sharding_key_update to the new parser

sgrif:
More specifically, this ports over the glue code as well as the last subroutine of that function, `create_stmts`. This should be the beefiest function we have to port (dear god please let this be the beefiest one).

The new implementation diverges from the old in a couple of ways:

- We deparse things lazily instead of eagerly
  - The old code was only deparsing eagerly due to the slowness of deparsing. This should no longer be slow enough to justify that. If needed we can memoize the result, but I'd rather have it be a memoized method than something upstream code cares about
- We do not rewrite the bind parameters, instead making every query use all of them
  - The old code did a recursive mutable AST walk to look for any bind parameters in the query, rewrite/compress the numbers to be a contiguous list starting from 1, and then mapepd that to the original. I *really* don't want to have a mutable AST walk for the new parser, so instead I just shove a no-op CTE that uses all the params in the queries and copy the original bind over.
- ShardingKeyUpdate holds the entire original UpdateStmt rather than smaller pieces
  - I actually tried having this hold onto references of the original query, but ultimately this struct ends up in a field that can't is a sibling of the original AST, so it can't hold a lifetime like that. I probably will revist that at some point, but for now copying the whole thing made more sense than copying each smaller piece into its own allocation

The unit tests look a little funky now to allow the new code to pass. What they're testing is technically a little bit weaker than before (it just checks that the bind parameters are all in the right places, it gives zero shits about the numbers). However, once the port is finished, I'll likely delete all but one of those unit tests. The complex logic they were testing simply no longer exists, and we really just need one test to assert that the where clause gets stuck on all 3 subqueries, and that all 3 queries use the same bind parameters as the original update

The CTE hack for the bind params really is funky, but I couldn't come up with any less hacky way to do it. I would very much prefer to stick `SELECT $1, $2, ...` somewhere instead, but that won't work. We can't rely on the client having sent us OIDs for the bind parameters in their `Parse` message, and any binds that appear in a top level select will get inferred to `text`. Since we can't assume any random blob of bytes is valid UTF-8, that would cause the query to fail. Putting the original update with `false AND` in the where clause was the simplest way I could think of to ensure the query uses every bind parameter, and keep the same type inference that we had before.

Other than the caveats mentioned above, the actual logic is mostly a line for line port. The syntax is certainly very different for constructing these trees, but it's a clean mapping each way.

### DIFF
--- a/pgdog/src/frontend/router/parser/rewrite/statement/update.rs
+++ b/pgdog/src/frontend/router/parser/rewrite/statement/update.rs
@@ -1564,6 +1564,7 @@ mod test {
             Field::bigint("id"),
             Field::text("email"),
             Field::text("other_col"),
+            #[cfg(feature = "new_parser")]
             Field::text("other_other_col"),
         ]);
 
@@ -1572,6 +1573,7 @@ mod test {
         data_row.add("1"); // id - will be overwritten by mapping
         data_row.add("old@example.com"); // email - will be overwritten by mapping
         data_row.add("other_value"); // other_col - from existing row
+        #[cfg(feature = "new_parser")]
         data_row.add("other_other_value"); // other_other_col - from existing row
 
         // Create a simple query request (not prepared statement)
--- a/pgdog/src/frontend/router/parser/rewrite/statement/update.rs
+++ b/pgdog/src/frontend/router/parser/rewrite/statement/update.rs
@@ -14,7 +14,9 @@ use pg_query::{
     },
 };
 #[cfg(feature = "new_parser")]
-use pg_raw_parse::{DeparseResult, Node, NodeMut, Owned, deparse, make::owned, nodes, walk};
+use pg_raw_parse::make::{owned, try_owned};
+#[cfg(feature = "new_parser")]
+use pg_raw_parse::{DeparseResult, Node, NodeMut, Owned, deparse, nodes, walk};
 #[cfg(not(feature = "new_parser"))]
 use pgdog_config::QueryParserEngine;
 use pgdog_config::RewriteMode;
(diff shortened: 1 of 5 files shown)


sgrif:
Once again going to assume `wait_for_replication_finishes_with_unrelated_writes` failing is unrelated to my change.

pgdog/src/frontend/router/parser/rewrite/statement/update.rs
```diff
@@ -140,13 +166,213 @@ pub(crate) struct Inner {
...
+
+        let mut bind = Bind::new_statement("");
+        let mut columns = Vec::new();
+        let mut values = Vec::new();
+        let mut columns_str = Vec::new();
+        let mut values_str = Vec::new();
+
+        let mut bind_idx = 1;
+        for (row_idx, field) in row_description.iter().enumerate() {
+            columns_str.push(format!(r#""{}""#, field.name.replace("\"", "\"\""))); // Escape "
```
levkk: We have an `escape_identifier` function lying around somewhere you could re-use.

pgdog/src/frontend/router/parser/rewrite/statement/update.rs
```diff
@@ -140,13 +166,213 @@ pub(crate) struct Inner {
...
+                .set_valuesLists(mem.make_List(&[mem.make_List(&values)]));
+            insert.as_mut().set_selectStmt(select.uncast());
+            insert
+                .as_mut()
+                .set_returningList(mem.make_unique(returning_list));
+            mem.make_List(&[mem.make_RawStmt(insert.uncast())])
+        });
+
+        let stmt = format!(
+            "INSERT INTO {} ({}) VALUES ({}){}",
```
levkk: Ah yeah, using the parser to build this would be better, but we can live with it while we port stuff over. Just less likely to make a mistake, especially around encoding weird types.

pgdog/src/frontend/router/parser/rewrite/statement/update.rs
```diff
@@ -685,6 +894,132 @@ fn deparse_list(
...
+    let has_bind_params = walk::walk_manual(stmt.into(), |node| match node {
+        Node::ParamRef(_) => std::ops::ControlFlow::Break(()),
+        _ => Recurse::yes(),
+    })
+    .is_some();
+
+    // Stick the original UPDATE in a CTE with `AND false` at the end of the
+    // WHERE clause to ensure all bind parameters are used, and have their
+    // types inferred the same way they would have in the original query
+    let use_all_params = if has_bind_params {
```
levkk: I get it, but it definitely screams like a hack around not having the ability to mutate an AST. Like, there is no reason to send this to Postgres whatsoever, so that's why it kinda stands out. Anyway, something to think about maybe - mutable ASTs are important to have - we are gonna do way more complicated things down the line for sure, so we're gonna need the ability to at least construct new ones using some kind of rewrite rules.

levkk (approved):
Provisionally agreed, but this will need cleanup - we got some more features coming in here soon and they will conflict - our current implementation is pretty easy to mod, this one might not be, I'm not sure. N.B.: I only had one coffee so far so I may be hallucinating issues.

pgdog/src/frontend/router/parser/rewrite/statement/update.rs
sgrif: I agree, but for the most part I just took what the old version did line for line

pgdog/src/frontend/router/parser/rewrite/statement/update.rs
sgrif: Construct new ones, yes. Iterating a tree while mutating it is a footgun though for all the same reasons as any other data structure.

sgrif:
I'll for sure look at adding a transforming AST walk (which is different than mutating, it will return a new ast) It'll probably after the port is done unless some other part of the code desperately needs it. I'll revisit this once I have that 

Though I'm gonna be honest, despite how hacky it is I don't agree that it's obviously worse than what the old code was doing. The sheer number of tests it has for various edge cases that just aren't necessary here says a lot.

sgrif:
I'm gonna spike on both transforming and (limited) mutating AST walks on Monday. The more I think about it the more I realize at least one of those two is inevitable. My main concern is ensuring that they can be done safely, I don't want to have to have every ast walk be in an unsafe block. It's possible I'm being overly cautious.

Anyway the more I think about it the more I realize it doesn't make sense to merge this until I've fully explored that problem.

levkk (approved):
I assume the PR description is not accurate anymore since you're creating params now, and also that all of this is fully covered by tests!

sgrif:
Yup, planning on rewriting it on merge

PR #1138, opened 2026-07-03
Remove dead `OwnedColumn`, mark `Column` as `pub(crate)`

sgrif:
THIS IS WHAT HAPPENS WHEN YOU MARK EVERYTHING AS PUB, ROBOTS


sgrif:
(Asked for review because there were some columns that were never read in an enum that is in parts of the code I've never touched)

levkk (approved):
Yeah I think so, I know all of this should be fully covered

sgrif:
Oh it's covered by the compiler. Just wanted to make sure these dead fields weren't intentionally there for something you're working on

PR #1146, opened 2026-07-06
Remove unneccessarily complex expression identification

sgrif:
I think the original intent here was to be able to have something like `SELECT COUNT(foo::bigint), AVG(foo::bigint)` use the existing count column based on any arbitrary expression. However, if this ever did fire it would have been wrong for anything other than `COUNT`, since it assumed things like typecasts were noops which is very much not the case for any other aggregate function, and none of this matters anyway since I deleted the helper deduplication a while ago. So instead we can just use the index of the output column as the identifier for each helper.


levkk (approved):
I believe you but the context on this is now long gone from my mind.

PR #1147, opened 2026-07-06
Port `Aggregate::parse` to the new parser

sgrif:
This is more or less a line for line rewrite, though with some minor refactoring to make the code flow a bit more clearly. A few tests got a bit funky because they're assuming the AST passed to `Aggregate::parse` is the same AST that gets rewritten later, which is not currently the case. That funkiness should go away as soon as I port the aggregate rewrite code to the new parser


levkk (approved):
very nice

PR #1148, opened 2026-07-06
Port `LimitClause` to the new parser

sgrif:
This is mostly a line-for-line port with a few minor refactors to make the code flow more smoothly. The only behavior change is that we now error if we see `LIMIT $1` but have no bind parameters, instead of silently treating it as NULL.

The main structural difference in the refactor is that I've removed any `return Ok(None)` that appeared other than cases where the node we're looking at really does mean NULL. Things like `A_Const` with a value other than an integer fall through to the `_` branch of the top level match, so that if/when we change this to treat unrecognized nodes as an error, it all happens in one place.


levkk:
Careful with erroring out - the following flow is real:

```
Parse(SELECT * FROM users LIMIT $1)
Describe
Flush
```

We use round robin and send this to only one shard.

sgrif:
> Careful with erroring out - the following flow is real:
> 
> ```
> Parse(SELECT * FROM users LIMIT $1)
> Describe
> Flush
> ```
> 
> We use round robin and send this to only one shard.

If only we had test coverage for this... (We do I just didn't run that test :rofl:)

PR #1151, opened 2026-07-06
Port StatementRewrite::rewrite_aggregates to the new parser

sgrif:
This is mostly a line for line port, with some minor refactoring to make the code flow better. There is a minor behavioral change in that we generate `SUM(arg::numeric * arg::numeric)` for sum of squares instead of `SUM(POWER(arg::numeric, 2))`. This was slightly more convenient to write, and in most languages `x * x` tends to be more performant than `pow(x, 2)` since generic power functions have to handle fractional exponents

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3299,7 +3299,7 @@ dependencies = [
 [[package]]
 name = "pg_raw_parse"
 version = "0.1.0"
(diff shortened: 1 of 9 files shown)


levkk (approved):
neat

PR #1154, opened 2026-07-07
Use the new parser to count params

sgrif:
We only do this if we have a CRUD statement, as this function gets called with all the obscure top level statements (transaction, create, etc) that PG errors with if passed to its walker


levkk (approved):
can't wait for the red PR

sgrif:
🔜

Issue #912, opened 2026-04-17
[Load balancer] Sticky reads


levkk:
Stick to the primary for a read from a table that just received a write from the same connection. This can help with the following:

```ruby
user = User.create email: "test@test.com"
user.reload
```

ats1999:
Curious about how it can be achieved? Read your own write consistency is scoped to user session, pgdog will have no idea about user.

mediprtl:
We hit this on the replica side rather than the primary. Our consumers read right after a write whose commit LSN they already know (it rides on an async message), but PgDog routes reads per transaction with no replica affinity, so a read can land on a replica that hasn't replayed the write yet. (#823 is the same problem from a real user, missing ~30% of reads; the only fix today is routing whole tables to the primary, which gives up replica offload.)

We prototyped a per-read freshness floor that keeps reads on replicas. Full implementation + tests here: https://github.com/mediprtl/pgdog/pull/4 (design notes in https://github.com/mediprtl/pgdog/issues/3).

- `pgdog.min_lsn` param (startup `options=-c` or `SET pgdog.min_lsn = 'X/Y'`), intercepted like `pgdog.role`. The client passes the LSN it must observe.
- The load balancer routes the read only to a replica whose replayed LSN (already tracked by the LSN monitor) is >= `min_lsn`; unmet errors recoverably (client retries), with opt-in primary fallback.

Would you welcome this upstream?

levkk:
Have you guys tested your patch in prod yet and did it have a meaningful impact? A few concerns:

- This leaks the LSN abstraction to the client: clients that don't know how to poll for LSN can't use it
- Read after write will require a primary read 9 times out of 10. Replication delay in production systems is typically sub-second, but even then, that's enough to cause a race. So, while you're polling for the LSN (1-2ms roundtrip), the replicas have already caught up...usually at least. So you should be able to just get away with `/* pgdog_role: primary */ SELECT [...]` for those specific read after write queries
- The lsn monitor can run very frequently, but the typical cadence is once per second, to avoid spamming the databases. We could run it more frequently (it's just a config chage), but either way, it won't update the LSN often enough for a significant amount of query traffic to hit the replica where it matters (the <1s time it takes for the replica to catch up). The clients that read after write might as well just put a `sleep(1s)` in their code and that might have the same effect

[...]

PR #1156, opened 2026-07-07
feat: pgdog.min_lsn read-your-writes replica routing floor

mediprtl:
## Sticky reads via a per-read LSN floor - #912

Part of #912 (@levkk asked for a PR there). Opt-in read-your-writes that keeps
reads on replicas instead of pinning tables to the primary.

A client sets `pgdog.min_lsn` (startup `options=-c` or `SET pgdog.min_lsn='X/Y'`),
intercepted like `pgdog.role`. The load balancer routes the read only to a
replica whose replayed LSN (already tracked by the LSN monitor) is >= that floor.
If none qualify it errors recoverably with `NoReplicaCaughtUp` (SQLSTATE 58000)
carrying a catch-up ETA from the replica's replication lag, or — opt-in —
falls back to the primary.

### Addressing the concerns from #912
- **No client-side LSN polling.** Our consumers already hold the commit LSN — it
  rides on the async message that triggers the read — so there's no round-trip to
  race. (PG 19's `WAIT FOR LSN` makes client-side LSN a first-class primitive too.)
- **Opt-in.** No param → unchanged behavior. It doesn't replace the row-change /
  per-driver-plugin direction you described — it's a small primitive those can
  build on.
- **Conservative.** A stale monitor view only costs an offload opportunity; it
  never serves a stale read.

### Testing

`pgdog.min_lsn` has run continuously in our production and sandbox for 3+ weeks,
built on the LSN monitor PgDog already ships (it reads the monitor's tracked
replayed LSN — no new poller).

**Production — 2 read replicas**
- On the order of **~8M reads** carrying `pgdog.min_lsn` over the window.
- **>99.9% were served from a caught-up replica.** The remaining <0.1% never
  caught up before the client stopped retrying and fell back on the client side.
  **No stale reads** — routing is conservative (a read only goes to a replica
  whose replayed LSN is ≥ the floor).
- Reaching a replica took **~4.6M `NoReplicaCaughtUp` responses** (clients
  retried): most reads were served immediately or after one short wait; the floor
  only holds a read back until the replica catches up.
- **Real lag, not sub-second:** replica lag averaged **~2 min** and peaked at
  **~50 min** over the window. The floor adapts — the reported ETA averaged ~1 min
  and tracked actual lag (max ~26 min).
- **Stable** — no PgDog crashes or restarts attributable to the feature.

**Sandbox — 1 read replica**
- ~1M `NoReplicaCaughtUp` responses over ~4 weeks; lower lag / lighter load.

**Coverage** — unit tests (replica-filter routing, `NoReplicaCaughtUp` +
ETA-from-lag) and a pgx integration test in `integration/load_balancer/pgx/`.

### Notes / happy to adjust
- `min_lsn_primary_fallback` is currently a bool in `[general]` — glad to fold it
  into an existing routing enum if you'd prefer.
- `replay_lag_seconds` lives on the pool's `LsnStats` wrapper; can move it into
  `pgdog_stats::LsnStats` if that's the better home.

### DIFF
--- a/pgdog/src/backend/pool/lsn_monitor.rs
+++ b/pgdog/src/backend/pool/lsn_monitor.rs
@@ -48,7 +48,16 @@ SELECT
             COALESCE(pg_last_xact_replay_timestamp(), now())
         ELSE
             now()
-    END AS timestamp
+    END AS timestamp,
+    CASE
+        WHEN pg_is_in_recovery() THEN
+            COALESCE(
+                EXTRACT(EPOCH FROM (now() - pg_last_xact_replay_timestamp()))::bigint,
+                0
+            )
+        ELSE
+            0
+    END AS replay_lag_seconds
 ";
 
 static AURORA_LSN_QUERY: &str = "
@@ -56,13 +65,23 @@ SELECT
     pg_is_in_recovery() AS replica,
     '0/0'::pg_lsn AS lsn,
     0::bigint AS offset_bytes,
-    now() AS timestamp
+    now() AS timestamp,
+    0::bigint AS replay_lag_seconds
 ";
+
+/// Cap a replication-lag ETA at one day so a pathological lag value can't
+/// overflow `Duration`; clients clamp further.
+const MAX_REPLAY_LAG_SECONDS: i64 = 86_400;
 
 /// LSN information.
 #[derive(Debug, Clone, Copy, Default)]
 pub struct LsnStats {
     inner: StatsLsnStats,
+    /// Replica replication lag in whole seconds: `now() -
+    /// pg_last_xact_replay_timestamp()`, computed DB-side (one clock, no
+    /// pgdog<->DB skew or timezone handling). `0` on the primary, on Aurora, or
+    /// when the replica has not replayed any transaction yet.
+    replay_lag_seconds: i64,
 }
 
 impl Deref for LsnStats {
@@ -81,7 +100,10 @@ impl DerefMut for LsnStats {
 
 impl From<StatsLsnStats> for LsnStats {
     fn from(value: StatsLsnStats) -> Self {
-        Self { inner: value }
+        Self {
+            inner: value,
+            replay_lag_seconds: 0,
+        }
     }
 }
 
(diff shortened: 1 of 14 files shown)


pgdog/src/backend/pool/lsn_monitor.rs
```diff
@@ -48,21 +48,40 @@ SELECT
             COALESCE(pg_last_xact_replay_timestamp(), now())
         ELSE
             now()
-    END AS timestamp
+    END AS timestamp,
```
levkk, line 51: This is not the correct formula for replica lag. This only works if and only if the primary and replicas do not have the same LSN, i.e. replica is behind. When they do, i.e., replica is caught up to the primary and the primary hasn't sent any writes yet, the replica lag should be zero, but this will give it a non-zero lag.

We handle this in here: https://github.com/pgdogdev/pgdog/blob/main/pgdog/src/backend/pool/shard/monitor.rs#L139

It's an edge case and it's not "terminal", i.e., it won't break the app but will cause pgdog to route the query to a primary unnecessarily. Since we have a function to calculate replica lag already, maybe you could re-use it here?

pgdog/src/backend/pool/lsn_monitor.rs
```diff
@@ -109,19 +131,40 @@ impl LsnStats {
...
         }
     }
+
+    /// Estimated time for this replica to reach `min_lsn`: its replication lag in
+    /// time (`now() - pg_last_xact_replay_timestamp()`, measured DB-side), used by
+    /// the client to size its read-your-writes defer. A stable single-sample
+    /// reading — a bytes/apply-rate derivative gets fooled by bursty replay into a
+    /// runaway ETA. Safe-biased (true wait <= lag, since `min_lsn` committed after
+    /// the frontier). `None` if stats are invalid; `ZERO` once at/past the floor.
+    pub fn eta_to(&self, min_lsn: i64) -> Option<Duration> {
```
levkk, line 141: See comment above about re-using existing replica lag calculator.

pgdog/src/frontend/client/query_engine/connect.rs
```diff
@@ -30,7 +33,17 @@ impl QueryEngine {
 
         let connect_route = connect_route.unwrap_or(context.client_request.route());
 
-        let request = Request::new(context.id, connect_route.is_read());
+        // Read-your-writes floor: a client sets `pgdog.min_lsn` (startup param or
+        // `SET`) to the commit LSN it must observe; the LB routes the read only to
+        // a replica that has replayed at least this far.
+        let min_lsn = context
+            .params
```
levkk, line 40: For code quality reasons, maybe we could handle this here: https://github.com/pgdogdev/pgdog/blob/main/pgdog/src/frontend/router/parameter_hints.rs

Caveat is you have to store this on `Route` and pass it back up, and that the parser has to be "on", but that's the case with your deployment since you're using the load balancer, so I think that should work.

levkk:
Looks great actually - I was wrong about pushing back on this. A few code quality comments to go through and we should be good to merge.

PR #1157, opened 2026-07-07
fix: skip multi-statement SET validation in session mode

### LINKED ISSUE #1087: Session mode incorrectly rejects multi-statement queries mixing SET with other commands

zefixlluja:
## Summary

In session mode, PgDog should forward multi-statement queries to the server verbatim without validating whether they mix SET with other commands. The server owns the session and can handle these naturally.

Previously, `try_multi_set` unconditionally rejected any multi-statement batch mixing SET with non-SET statements (e.g. SHOW), returning `Error::MultiStatementMixedSet`. This broke psqlODBC connections, which send exactly such a batch during connection startup:

```sql
SET DateStyle='ISO';SET extra_float_digits = 2;show transaction_isolation
```

## Changes

- `QueryParserContext::is_session_mode()` — checks the cluster's pooler mode
- `try_multi_set` returns `Ok(None)` early in session mode, letting the query fall through to the server unchanged
- `cargo fmt` fix in `context.rs` (method-chain line break)
- Unit tests: `test_mixed_set_passthrough_in_session_mode` and `test_mixed_set_rejected_in_transaction_mode`
- `Cluster::new_test_session_mode` / `QueryParserTest::new_session_mode` added as test helpers

## Test plan

- [ ] `cargo nextest run` — new unit tests cover both session-mode passthrough and transaction-mode rejection
- [ ] Manual: connect via psqlODBC with `pooler_mode = session` and verify the connection startup batch succeeds

Closes #1087

### DIFF
--- a/pgdog/src/backend/pool/cluster.rs
+++ b/pgdog/src/backend/pool/cluster.rs
@@ -737,7 +737,9 @@ impl Cluster {
 mod test {
     use std::{sync::Arc, time::Duration};
 
-    use pgdog_config::{ConfigAndUsers, OmnishardedTable, QueryParserLevel, ShardedSchema};
+    use pgdog_config::{
+        ConfigAndUsers, OmnishardedTable, PoolerMode, QueryParserLevel, ShardedSchema,
+    };
 
     use crate::frontend::router::sharding::ShardedTable;
     use crate::{
@@ -876,6 +878,12 @@ mod test {
             let mut cluster = Self::new_test(config);
             cluster.shards.pop();
 
+            cluster
+        }
+
+        pub fn new_test_session_mode(config: &ConfigAndUsers) -> Cluster {
+            let mut cluster = Self::new_test(config);
+            cluster.pooler_mode = PoolerMode::Session;
             cluster
         }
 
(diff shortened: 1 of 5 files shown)


levkk (commented):
Sorry for the delay. Busy week. LGTM!

levkk (approved):

PR #1158, opened 2026-07-07
Port `StatementRewrite::rewrite_simple_prepared` to the new parser

sgrif:
I'm trying something new with this function, I'm intentionally using a silly form of cfg'ing the old code out to force those lines to have an indentation change, as this causes the diff to interpret the new code as changed lines and the old code as an addition. This will result in less churn in the long run (since the old code is getting deleted), and gives way nicer diffs for review

### DIFF
--- a/pgdog/src/frontend/router/parser/rewrite/statement/mod.rs
+++ b/pgdog/src/frontend/router/parser/rewrite/statement/mod.rs
@@ -124,17 +124,22 @@ impl<'a> StatementRewrite<'a> {
                 _ => (),
             }
 
+            let mut node = mem.make_unique(node);
+
             // Handle top-level PREPARE/EXECUTE statements.
-            let prepared_result = self.rewrite_simple_prepared()?;
+            let prepared_result = self.rewrite_simple_prepared(node.as_mut(), mem)?;
             if prepared_result.rewritten {
                 self.rewritten = true;
                 plan.prepares = prepared_result.prepares;
+                self.stmt.stmts = pg_query::parse(pg_raw_parse::deparse(node.as_ref())?.as_str())?
+                    .protobuf
+                    .stmts;
             }
 
             // Inject pgdog.unique_id() for missing BIGINT primary keys.
             // This must run BEFORE the unique_id rewriter so the injected
             // function calls get processed.
-            self.inject_auto_id(node, &mut plan)?;
+            self.inject_auto_id(node.as_ref(), &mut plan)?;
 
             // Track the next parameter number to use
             let mut next_param = plan.params as i32 + 1;
--- a/pgdog/src/frontend/router/parser/rewrite/statement/simple_prepared.rs
+++ b/pgdog/src/frontend/router/parser/rewrite/statement/simple_prepared.rs
@@ -1,4 +1,8 @@
+#[cfg(not(feature = "new_parser"))]
 use pg_query::{Error as PgQueryError, NodeEnum};
+#[cfg(feature = "new_parser")]
+use pg_raw_parse::{NodeMut, make::MemoryToken};
+#[cfg(not(feature = "new_parser"))]
 use pgdog_config::QueryParserEngine;
 
 use crate::net::Parse;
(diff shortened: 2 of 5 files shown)


pgdog/src/frontend/router/parser/rewrite/statement/mod.rs
```diff
@@ -124,17 +124,22 @@ impl<'a> StatementRewrite<'a> {
...
+
             // Handle top-level PREPARE/EXECUTE statements.
-            let prepared_result = self.rewrite_simple_prepared()?;
+            let prepared_result = self.rewrite_simple_prepared(node.as_mut(), mem)?;
             if prepared_result.rewritten {
                 self.rewritten = true;
                 plan.prepares = prepared_result.prepares;
+                self.stmt.stmts = pg_query::parse(pg_raw_parse::deparse(node.as_ref())?.as_str())?
+                    .protobuf
+                    .stmts;
```
meskill, line 136: can you help understand this? do we need to update stmt just to push the modified version to self and provide the compatibility with old parser types? and for old parser it's not required because we do the in-place mutation of stmt?

pgdog/src/frontend/router/parser/rewrite/statement/simple_prepared.rs
```diff
@@ -39,84 +43,161 @@ impl StatementRewrite<'_> {
...
-        NodeEnum::ExecuteStmt(stmt) => {
-            let parse = prepared_statements.parse(&stmt.name);
+        NodeMut::ExecuteStmt(mut stmt) => {
+            let stmt_name = stmt.name().expect("EXECUTE always has name");
+            let parse = prepared_statements.parse(stmt_name);
             if let Some(parse) = parse {
                 let global_name = parse.name().to_string();
                 let statement = parse.query().to_string();
-                stmt.name = global_name.clone();
+                stmt.set_name(Some(mem.copy_string(&global_name)));
```
meskill, line 132: I'm sorry for asking probably a dumb question, but I'm still at the start of my journey to understand the new parser. But could the `set_name` be simplified to not juggle with explicit mem operation to copy the data? I mean could the `set_name` handle it's by itself in its internals? I think the mem token is not yet stored on NodeMut, but probably we can store it or get the proper mem context somehow?

pgdog/src/frontend/router/parser/rewrite/statement/mod.rs
sgrif: Correct. I'm parsing these functions in relatively small pieces, so some of the other functions called here modify or read from `self.stmt` instead of `node`. So any time a function modifies `self.stmt` we deparse and reparse into `node`, and any time we modify `node`, we need to deparse and reparse into `self.stmt`. Once everything in `StatementRewrite` is ported, this will no longer be necessary and I will cfg out the `stmt` field.

pgdog/src/frontend/router/parser/rewrite/statement/simple_prepared.rs
sgrif: You never need to apologize for asking questions. Part of the reason for code review is to spread knowledge 😄 

> But could the set_name be simplified to not juggle with explicit mem operation to copy the data? I mean could the set_name handle it's by itself in its internals?

We do this in constructor functions, but not setters. This is because the node doesn't have a reference to the memory context.

> I think the mem token is not yet stored on NodeMut, but probably we can store it or get the proper mem context somehow?

We intentionally avoid keeping anything on `Node`/`NodeMut` other than a single pointer or zero-sized types. We want the node structs to be layout compatible with the C versions, and we want `Node`/`NodeMut`/`&SomeNode`/`SomeNodeMut` to all be layout identical to a `*mut raw::Node`, so that the conversion cost is nothing more than a pointer cast. (There's some caveats with `Node`/`NodeMut` that I'm happy to go into if you'd like but aren't relevant to this question)

sgrif: Shorter version of the answer: I don't see having to write `mem.copy_string(s)` as any different than having to write `s.clone()`. The only difference here is that we're using PG's allocator instead of the global one.

meskill: > Shorter version of the answer: I don't see having to write mem.copy_string(s) as any different than having to write s.clone(). The only difference here is that we're using PG's allocator instead of the global one.

Not a big deal for sure. Just wondering if it can be simpler without need to pass both node and mem through functions.

> We want the node structs to be layout compatible with the C version

Yea, got it. I think I was looking into the NodeMut variant types like this

image

That already are rust wrappers over repr(C) struct. So maybe...

Or, what about [GetMemoryChunkContext](https://github.com/postgres/postgres/blob/master/src/backend/utils/mmgr/mcxt.c#L753-L762)? can we get the memory context with it and copy_string into it? I think it's just more unsafe calls internally outside of safe wrappers, but just curios.

meskill (approved):

pgdog/src/frontend/router/parser/rewrite/statement/simple_prepared.rs
sgrif: I hadn't seen `GetMemoryChunkContext` before. Yes, in theory we could use that.

sgrif: Actually looking at it, it's not clear to me if we could use that or not. It reads to me like it's getting whatever the active memory context of a given type is, but not any arbitrary memory context

meskill: isn't that we need for this? i.e. the context where the current node is located so we can copy data into the same context as with strings? ofc if we consider the NodeMut that should have a pointer to the newly created context

sgrif: "Active" in this case means left in a thread local, which we never leave for longer than necessary. I may be misreading that code though. I'm not a C programmer and they're doing some weird vtable shit so maybe it does get what we'd need.

PR #1161, opened 2026-07-07
Ensure we handle pipelined cross-shard queries

sgrif:
When mulitple queries are received before a Sync message, we need to treat them as inside an implicit transaction. Before this commit, we'd see that the first statement routes to a single shard, open a direct-to-shard connection to that shard, and forward the message. This would result in the implicit transaction on the shard's end, purely because it's receiving the messages in the same order we do. But we were completely unaware of the transaction-like behavior on our end, unless the client was explicitly starting a transaction.

With this change, we now correctly recognize that we're in a transaction any time we receive multiple executable messages prior to a sync, and behave the same as we would if the first query we received in the pipeline were `BEGIN`

### DIFF
--- a/integration/go/go_pgx/batch_sharded_test.go
+++ b/integration/go/go_pgx/batch_sharded_test.go
@@ -1,0 +1,86 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/jackc/pgx/v5"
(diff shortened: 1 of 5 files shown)


levkk (approved):
Very elegant and _correct_ fix. Thanks!

sgrif:
> Very elegant and _correct_ fix. Thanks!

That's my style 💜

PR #1165, opened 2026-07-08
feat: add support for having clause

### LINKED ISSUE #695: [Query engine] Support HAVING in aggregates
### Description

Add support for `HAVING` clause in `GROUP BY` cross-shard aggregates.

murex971:
## Summary

- Add support for `HAVING` in cross-shard `GROUP BY` queries.
- Introduce `HavingRewritePlan` to inject hidden HAVING dependencies (when not in SELECT).
- Update multi-shard processing order to run: aggregate -> having -> sort/distinct/limit, then drop hidden columns before returning rows/row description to clients.

## Testing
Added unit and integrated tests

## Related Issues
 closes #695

### DIFF
--- a/integration/rust/tests/integration/having.rs
+++ b/integration/rust/tests/integration/having.rs
@@ -1,0 +1,91 @@
+use crate::setup::{admin_sqlx, connections_sqlx};
+use sqlx::{Executor, Row, postgres::PgPool};
+
+#[tokio::test]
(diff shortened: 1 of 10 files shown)


levkk:
This will require a thorough review. Will try to circle back next week due to https://github.com/pgdogdev/pgdog/pull/1114#issuecomment-4921291575

murex971:
Note: I have added hidden column helper logic in this having clause which is a current limitation of `ORDER BY`

PR #1166, opened 2026-07-08
perf(replication): implement async processing during replication

### LINKED ISSUE #1036: Add async streaming for replication to improve replication speed
The current replication process is strictly sequential with parallelization only between multiple shards. Every event runs the cycle `send -> wait for response` that can slow the whole process significantly especially with high latency on read.

Rework the process to push the updates without waiting for response immediately and read the responses asynchronously in separate task to react on possible errors.

ygxio:
fixes #1036

### DIFF
--- a/pgdog/src/backend/replication/logical/subscriber/pipeline.rs
+++ b/pgdog/src/backend/replication/logical/subscriber/pipeline.rs
@@ -324,10 +324,10 @@ impl Listener {
                 } else {
                     false
                 };
-                if resolved {
-                    if let Some((_, done)) = self.parse_ack_sync_points.pop_front() {
-                        let _ = done.send(());
-                    }
+                if resolved
+                    && let Some((_, done)) = self.parse_ack_sync_points.pop_front()
+                {
+                    let _ = done.send(());
                 }
             }
             // BindComplete: nothing to account for.
@@ -335,12 +335,11 @@ impl Listener {
             // CommandComplete: match to the DML op and count missed rows.
             'C' => {
                 let is_direct = self.direct.pop_front().unwrap_or(false);
-                if is_direct {
-                    if let Ok(complete) = CommandComplete::try_from(message) {
-                        if matches!(complete.rows(), Ok(Some(0))) {
-                            self.shared.lock().missed.record(complete.tag());
-                        }
-                    }
+                if is_direct
+                    && let Ok(complete) = CommandComplete::try_from(message)
+                    && matches!(complete.rows(), Ok(Some(0)))
+                {
+                    self.shared.lock().missed.record(complete.tag());
                 }
                 if self.direct.is_empty() {
                     self.wake_drain();
--- a/pgdog/src/backend/replication/logical/subscriber/pipeline.rs
+++ b/pgdog/src/backend/replication/logical/subscriber/pipeline.rs
@@ -27,9 +27,6 @@ struct Shared {
     error: Option<Error>,
     // Rows a direct-to-shard DML expected to touch but didn't (0 rows affected).
     missed: MissedRows,
-    // Test-only: total ParseComplete acks consumed by parse-ack sync points.
-    #[cfg(test)]
-    parse_acks_consumed: usize,
 }
 
 // How a sync point completes.
@@ -38,6 +35,16 @@ enum SyncPointKind {
     ParseAcks(usize),
     // Wait for a single ReadyForQuery (`Sync`: commit, or out-of-transaction prepare).
     ReadyForQuery,
+}
+
+// One entry per command sent to Postgres, in send order. Popped as acks arrive.
+enum OpSyncPoint {
+    // Bind/Execute/Flush: resolved by CommandComplete ('C').
+    DirectDml { is_direct: bool },
+    // In-transaction prepare (Flush): resolved after `remaining` ParseComplete ('1') acks.
+    ParseAcks { remaining: usize, done: oneshot::Sender<()> },
+    // Commit or out-of-transaction prepare (Sync): resolved by ReadyForQuery ('Z').
+    ReadyForQuery { done: oneshot::Sender<()> },
 }
 
 // Work sent from the handle to the listener task.
@@ -79,9 +86,7 @@ impl PipelinedConnection {
             rx,
             server,
             shared: shared.clone(),
-            direct: VecDeque::new(),
-            parse_ack_sync_points: VecDeque::new(),
-            ready_for_query_sync_points: VecDeque::new(),
+            queue: VecDeque::new(),
             drain_waiter: None,
         };
 
@@ -129,13 +134,13 @@ impl PipelinedConnection {
             messages.push(Sync.into());
             SyncPointKind::ReadyForQuery
         };
-        self.sync_point(messages, kind).await
+        self.send_command_and_wait_for_sync_point(messages, kind).await
     }
 
     /// Send `Sync` and wait for `ReadyForQuery` (commits the open implicit
     /// transaction on this shard).
     pub async fn sync_and_drain(&self) -> Result<(), Error> {
-        self.sync_point(vec![Sync.into()], SyncPointKind::ReadyForQuery)
+        self.send_command_and_wait_for_sync_point(vec![Sync.into()], SyncPointKind::ReadyForQuery)
             .await
     }
 
@@ -147,7 +152,7 @@ impl PipelinedConnection {
             .send(Command::DrainAcks { done })
             .await
             .map_err(|_| Error::PipelineClosed)?;
-        self.await_done(rx).await
+        self.resolve_sync_point(rx).await
     }
 
     /// Non-blocking peek + take of the latched error. `Some` means the shard
@@ -161,14 +166,8 @@ impl PipelinedConnection {
         std::mem::take(&mut self.shared.lock().missed)
     }
 
-    /// Test-only: number of ParseComplete acks the listener has attributed to
-    /// parse-ack sync points so far.
-    #[cfg(test)]
-    fn parse_acks_consumed(&self) -> usize {
-        self.shared.lock().parse_acks_consumed
-    }
-
-    async fn sync_point(
+    // This function send the prepared command along with the type of sync point we are waiting for.
+    async fn send_command_and_wait_for_sync_point(
         &self,
         messages: Vec<ProtocolMessage>,
         kind: SyncPointKind,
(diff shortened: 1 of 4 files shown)


ygxio:
I am still working on the design for it, just a peak if the direction seems well @levkk

levkk:
Cool! Would be good to see some benchmark numbers, just to get a sense of how much faster this is.

ygxio:
Do you have any such setup so that i can benchmark it?

levkk:
> Do you have any such setup so that i can benchmark it?

This might work:

```
benches/bench.sh
```

Specifically, see `resharding` folder.

ygxio:
image


PGDOG_BIN=/tmp/pgdog-new bash benches/resharding/replication/run.sh --baseline main

ygxio:
For BENCH_SCALE = 3000, DEST_LATENCY_MS=10 we have more performance boosts


image

ygxio:
RUNS = 2
WARMUP = 1
BENCH_SCALE = 5000
DEST_LATENCY_MS = 10
USE_TOXI = 1


image

@levkk results seems good but also too good at a point

levkk:
Async should be a big improvement, so not _that_ surprising. This is very cool. Let me know when you're ready for a review, I'll tag Kiryl.

ygxio:
@levkk ready for review

pgdog/src/backend/replication/logical/subscriber/pipeline.rs
```diff
@@ -0,0 +1,423 @@
...
+        }
+    }
+
+    fn address(&self) -> &Address {
+        self.server.addr()
+    }
+}
+
+#[cfg(test)]
+mod test {
```
meskill, line 410: Do you think we should add more unit tests here? I think we should have some for testing the whole flow - prepare -> execute -> await. And also some tests for possible errors

ygxio: Sure @meskill would add more tests

ygxio:
@meskill i have added a few more could you take a look?

meskill:
Thanks @ygxio I'm still checking the code more deeply, but anyway looking great so far.

Putting my bench results to confirm the improvements:

### Against main

With defaults:
image

With toxi and latency for source/destination for 1ms:
image

### Separately

**_And the next without comparison with main (compare against prev run) since since it is executing too long to get the result._**

With latency 10ms:
image

With latency 100ms:
image

ygxio:
i see from 1 to 10 to 100 ms of latency the jump has been pretty high,  need to see the cause maybe

pgdog/src/backend/replication/logical/subscriber/pipeline.rs
```diff
@@ -0,0 +1,530 @@
...
+    // One entry per outstanding DML op, in send order: `is_direct`. Popped on
+    // each CommandComplete.
+    direct: VecDeque<bool>,
+    // In-transaction prepare sync points: (remaining ParseComplete acks, waiter).
+    parse_ack_sync_points: VecDeque<(usize, oneshot::Sender<()>)>,
+    // Commit / out-of-transaction prepare sync points, resolved on ReadyForQuery.
+    ready_for_query_sync_points: VecDeque<oneshot::Sender<()>>,
+    // Waiter that resolves once every DML ack has been read. At most one is
+    // outstanding: `commit()` issues drain_acks sequentially, one per connection.
+    drain_waiter: Option<oneshot::Sender<()>>,
```
meskill: The multiple separate VecDeque like this works, but I see this more error-prone and harder to maintain. The issue is this multiple queues partially maps to the underlaying FIFO queue executed by the backend. Due to current design of the replication it works just fine, but in case we add something new to a processing or try to extend async handling that's could become a break point.

I'd suggest to use a single queue of enums to control the state of the processing and the expected responses, to make sure we get everything we expect and we don't process entries in the wrong order for some reason. 
I think the ProtocolState from `pgdog/src/backend/protocol/state.rs` does use the single queue for something like this.

pgdog/src/backend/replication/logical/subscriber/pipeline.rs
```diff
@@ -0,0 +1,530 @@
...
+}
+
+impl Listener {
+    async fn run(mut self) {
+        loop {
+            select! {
```
meskill, line 221: I can't prove it but I believe here could be a theoretical issue due to disproportion of read/writes events - the usual execute command from recv expected to generate 2 server.read events. Since the select! is fair by default the buffer for server.read could grow fast that could lead to tcp stream blocking. 

I'm thinking if we need to add `biased` to select! and move the read first, so we'll read the messages first if any to avoid this.

pgdog/src/backend/replication/logical/subscriber/stream.rs
```diff
@@ -223,23 +223,34 @@ impl StreamSubscriber {
...
         Ok(())
     }
 
     // Dispatch a pre-built bind to the matching shard(s).
     async fn send(&mut self, val: &Shard, bind: &Bind) -> Result<(), Error> {
+        // Fail-fast: the replicated transaction spans every shard, so an error
+        // latched on any connection means whole transaction is aborted.
+        for conn in &self.connections {
+            if let Some(err) = conn.take_error() {
+                return Err(err);
```
meskill, line 245: could you please check how does effect on one connection affect other connections and streams - are they cancelled in any way?
I think they stop after dropping the channel, but I want to be sure we are not doing useless work here.

pgdog/src/backend/replication/logical/subscriber/pipeline.rs
```diff
@@ -0,0 +1,530 @@
...
+
+// State shared between the handle and its background listener task.
+#[derive(Debug, Default)]
+struct Shared {
+    // First error observed on this connection. Sticky until taken.
+    error: Option<Error>,
+    // Rows a direct-to-shard DML expected to touch but didn't (0 rows affected).
+    missed: MissedRows,
+    // Test-only: total ParseComplete acks consumed by parse-ack sync points.
+    #[cfg(test)]
```
meskill: can we try to avoid this is the main code? 

let's better test that everything works as expected after some commands than to rely on some implementation details. If we have prepare statements - we can validate that this statement work after parsing and that listener doesn't block for example.

pgdog/src/backend/replication/logical/subscriber/pipeline.rs
meskill: We still miss some important tests:
- error path - something that will trigger an error response, validate that we get the error properly and we won't block the execution for future calls
- missed rows calculation - just some calls to generate missed rows and make sure we get proper stats for it

pgdog/src/backend/replication/logical/subscriber/pipeline.rs
```diff
@@ -0,0 +1,530 @@
...
+    /// Prepare `parses` and wait for the acknowledgments. Inside a transaction
+    /// uses `Flush` (must not commit the open implicit transaction); otherwise
+    /// uses `Sync`.
+    pub async fn prepare(&self, parses: &[Parse], in_transaction: bool) -> Result<(), Error> {
+        let mut messages: Vec<ProtocolMessage> = parses.iter().map(|p| p.clone().into()).collect();
+        let kind = if in_transaction {
+            // If in transaction we send flush and wait for the acknowledgements
+            // since we donot want to commit the open transaction.
+            messages.push(Flush.into());
+            SyncPointKind::ParseAcks(parses.len())
```
meskill, line 140: let's add some validation that this should not be 0, since we may overflow on this later

pgdog/src/backend/replication/logical/subscriber/stream.rs
```diff
@@ -458,37 +432,11 @@ impl StreamSubscriber {
...
-                        return Err(Error::PgError(Box::new(ErrorResponse::from_bytes(
-                            msg.to_bytes(),
-                        )?)));
-                    }
-                    'Z' => break,
-                    '1' => continue,
-                    c => return Err(Error::RelationOutOfSync(c)),
-                }
-            }
+            server.prepare(parses, in_txn).await?;
```
meskill: also can speed up this a very little if we use try_join to run prepare in parallel

meskill (changes requested):

pgdog/src/backend/replication/logical/subscriber/pipeline.rs
ygxio: I have pushed changes for it

pgdog/src/backend/replication/logical/subscriber/pipeline.rs
ygxio: Yes, this can be problem i am trying to come up with some way to prove this hypothesis

meskill: I'd say don't spend much time on it, we can just put biased to prioritize the read and for the write we have more explicit backpressure control thanks to channels

ygxio: i was able to get some result, so i shrinked my RCV and SND socket buffer and sent execute statements `select!` kept sending commands which made the pg -> pgdog receive buffer full hence on next send commands the system entered a deadlock since pgdog -> pg buffer is full and pg -> pgdog buffer is also full and there is no way pgdog can read the receive buffer, so read bais is definately important

pgdog/src/backend/replication/logical/subscriber/stream.rs
ygxio: right, it is a redundant task as i on Error i am already dropping the Sender channels hence should already work

pgdog/src/backend/replication/logical/subscriber/pipeline.rs
ygxio: handled this at prepare itself, to not allow empty parses to go for sync points

ygxio:
DEFAULT
image

Src/dest latecy = 1 ms

image

Src/dest latecy = 10 ms

image

Src/dest latecy = 100 ms

image


Looks like almost same performance

ygxio:
have fixed most of the issues, need a re review @levkk @meskill

pgdog/src/backend/replication/logical/subscriber/pipeline.rs
```diff
@@ -478,53 +479,222 @@ mod test {
...
+        let err = conn
+            .prepare(&[Parse::named("__pipe_bad", "NOT VALID SQL")], true)
+            .await
+            .unwrap_err();
+        assert!(
+            matches!(err, Error::PgError(_)),
+            "unexpected error: {err:?}"
+        );
+        assert!(conn.take_error().is_none());
+    }
```
meskill, line 517: those are identical tests, please verify

meskill (changes requested):
Looks good, only the small test fix and we'll merge

pgdog/src/backend/replication/logical/subscriber/pipeline.rs
ygxio: @meskill one it for out of transaction and other is for in transaction, should i merge them into one test with a loop over in_transaction value either false or true

meskill: oh, nvm, `true/false` option I didn't notice.

meskill (approved):
Thank you!

PR #1169, opened 2026-07-09
Port `StatementRewrite::inject_auto_id`

sgrif:
This was a very quick and easy port. I definitely did not lose an entire day to a lifetime issue because I wanted mutable lists to make this function look nicer when it was done being ported.


levkk (approved):
Neat

PR #1173, opened 2026-07-10
feat(tls): present client certificate to upstream servers for mTLS

### LINKED ISSUE #890: mTLS between pgdog and db server
Hi, I've read through https://docs.pgdog.dev/features/tls/ and https://docs.pgdog.dev/configuration/pgdog.toml/general/#tls_certificate but the settings there don't seem to have configuration for setting up mTLS between pddog and db server. There's no setting to present pgdog cert to the server and it seems to be server TLS only at this point connecting to DB. Additionally, there's no CA bundle field to verify the client cert presented to pgdog when tls_client_required is on.

Am I right in this assumption?

andy-paine:
## Summary

PgDog can validate an upstream Postgres server's certificate (`tls_verify` + `tls_server_ca_certificate`) but has no way to present its *own* client certificate on that connection. This means it can't connect to servers that require mutual TLS — e.g. Cloud SQL with "require trusted client certificate", or any Postgres configured with `clientcert=verify-full`.

This adds the missing upstream leg. Two new settings, named to mirror the existing `tls_server_ca_certificate`:

- **`tls_server_certificate`** — the client certificate PgDog presents to the server
- **`tls_server_private_key`** — its private key

(also available as `PGDOG_TLS_SERVER_CERTIFICATE` / `PGDOG_TLS_SERVER_PRIVATE_KEY`.)

Closes #890

## Implementation

- The certificate + key are loaded exactly the way `build_acceptor` loads PgDog's own server certificate (`CertificateDer::from_pem_file` / `PrivateKeyDer::from_pem_file`), then attached via rustls `with_client_auth_cert` across every `tls_verify` mode (`disabled` / `prefer` / `verify_ca` / `verify_full`).
- Setting only one of the two is rejected at load time, so misconfigurations surface on reload rather than at connection time.
- The upstream connector cache key now includes the client cert/key paths, so a cached no-client-auth connector isn't reused after a cert is configured.

## Tests

- `connector_requires_both_client_cert_and_key` — both-or-neither validation.
- `client_cert_is_part_of_connector_cache_key` — adding a client cert rebuilds the connector; identical inputs reuse the cache.
- `test_upstream_client_cert_config_parses` (pgdog-config) — the new keys parse and are optional.

### DIFF
--- a/.schema/pgdog.schema.json
+++ b/.schema/pgdog.schema.json
@@ -114,6 +114,8 @@
         "tls_client_required": false,
         "tls_private_key": null,
         "tls_server_ca_certificate": null,
+        "tls_server_certificate": null,
+        "tls_server_private_key": null,
         "tls_verify": "prefer",
         "two_phase_commit": false,
         "two_phase_commit_auto": null,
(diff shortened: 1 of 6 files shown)


andy-paine:
👋 Hi team, is there anything more you need me to do to move this along please? @levkk @sgrif

levkk (approved):
Looks good. Sorry for the delay!

If you ever get a chance to come back to this, an integration test would be awesome, maybe setting up the certs on the Postgres DB in CI and making sure Postgres accepts the ones PgDog passes through, but I think the current code works as intended, so only if you have some time.

PR #1086, opened 2026-06-18
feat: protect against omni writes inside direct-to-shard transactions

levkk:
### Breaking change

Using `SET pgdog.shard` and `pgdog.set_sharding_key` inside a transaction (`BEGIN`) binds the
entire transaction to that one shard only.

Added the following checks to prevent data consistency errors:

1. Omnisharded tables cannot be written to inside direct-to-shard transactions anymore
2. Caller cannot change shards mid-transaction anymore, e.g. by calling `SET pgdog.shard` or by running a direct-to-shard or cross-shard query

Technically, this not a breaking change per-se but a but a bug fix, since this was causing silent data corruption errors previously, but still worth calling out.

fix #1085

### Other changes

- refactor: rename really poorly named methods in `Route`
- refactor: connection binding enum handling was nonsense, now it's less so
- refactor: added lots of comments in places where I touched something in this PR
- refactor: removed some dead code

PR #1179, opened 2026-07-10
fix: don't sum row counts across shards for EXECUTE on omnisharded tables

mscrivo:
## Problem

`EXECUTE` of a server-side prepared statement targeting an omnisharded table returned the **sum** of row counts from all shards instead of the count from one shard:

```sql
PREPARE upd AS UPDATE sharded_omni SET value = $1;
EXECUTE upd('x');  -- UPDATE 6 on a 2-shard cluster with 3 rows (should be UPDATE 3)
```

Plain `UPDATE`/`DELETE`/`INSERT` on omnisharded tables already report the correct count (#752), but `ExecuteStmt` fell through to the DDL catch-all, which broadcasts with `Route::write(Shard::All)` and no omnisharded flag — so `MultiShard` aggregated the `CommandComplete` counts.

## Fix

- Route `PREPARE` explicitly: still broadcast to all shards, but extract the statement behind the name and store it in the regular prepared statements cache (`Parse::named(name, statement)` + `PreparedStatements::insert()`), the same way the `prepared_statements = "full"` rewriter does. First PREPARE wins, matching the server, where a duplicate PREPARE fails.
- Route `EXECUTE` explicitly: resolve the name via `PreparedStatements::parse()` and inspect the statement through the statement cache, so each unique statement is parsed once, not per execution. If it's a **write** that only touches omnisharded tables, the route carries the omnisharded flag and `MultiShard` deduplicates results across shards instead of aggregating them.
- Reads behind `EXECUTE` are not flagged: `EXECUTE` always routes as a write (the statement only exists on the connections `PREPARE` reached), and an omnisharded write requires full shard coverage — flagging reads would reject shard directives on statements that can't diverge the shards.
- The router reaches the client's prepared statements via a new optional `RouterContext::with_prepared_statements()`.

This also extends the omni-write-in-direct-to-shard-transaction guard (#1086) and the omni-write-with-directive guard to `EXECUTE` of omnisharded writes, since both key off the same route flag.

## Testing

- Routing tests in `parser/query/test/test_execute.rs` (fail without the fix): omnisharded UPDATE/DELETE/INSERT behind `EXECUTE` carry the flag; SELECT/VALUES/sharded/unknown names don't; shard directives are allowed on reads and rejected on omnisharded writes.
- Verified end-to-end against a 2-shard cluster: `EXECUTE upd('x')` returns `UPDATE 3` instead of `UPDATE 6`, and the write still reaches every shard. Same for `DELETE`.

### DIFF
--- a/pgdog/src/frontend/prepared_statements/mod.rs
+++ b/pgdog/src/frontend/prepared_statements/mod.rs
@@ -32,6 +32,10 @@ fn str_mem(s: &str) -> usize {
 pub struct PreparedStatements {
     pub(super) global: Arc<RwLock<GlobalCache>>,
     pub(super) local: HashMap<String, String>,
+    /// SQL-level `PREPARE` statements sent over the simple protocol,
+    /// name -> statement text. Tracked so `EXECUTE` can be routed
+    /// based on the statement behind the name.
+    pub(super) simple: HashMap<String, String>,
     pub(super) level: PreparedStatementsLevel,
     pub(super) memory_used: usize,
 }
@@ -41,6 +45,7 @@ impl Default for PreparedStatements {
         Self {
             global: Arc::new(RwLock::new(GlobalCache::default())),
             local: HashMap::default(),
+            simple: HashMap::default(),
             level: PreparedStatementsLevel::Extended,
             memory_used: 0,
         }
@@ -110,6 +115,21 @@ impl PreparedStatements {
         self.local.get(name)
     }
 
+    /// Record a SQL-level `PREPARE` statement.
+    pub fn insert_simple(&mut self, name: &str, query: &str) {
+        if let Some(old_query) = self.simple.insert(name.to_owned(), query.to_owned()) {
+            self.memory_used = self.memory_used.saturating_sub(str_mem(&old_query));
+            self.memory_used += str_mem(query);
+        } else {
+            self.memory_used += str_mem(name) + str_mem(query);
+        }
+    }
+
+    /// Get the query behind a SQL-level `PREPARE` statement name.
+    pub fn simple_query(&self, name: &str) -> Option<&str> {
+        self.simple.get(name).map(|query| query.as_str())
+    }
+
     /// Get globally-prepared statement by local name.
     pub fn parse(&self, name: &str) -> Option<Parse> {
         self.local
@@ -154,6 +174,7 @@ impl PreparedStatements {
         }
 
         self.local.clear();
+        self.simple.clear();
         self.memory_used = 0;
     }
 
--- a/pgdog/src/frontend/prepared_statements/mod.rs
+++ b/pgdog/src/frontend/prepared_statements/mod.rs
@@ -252,6 +252,28 @@ mod test {
     }
 
     #[test]
+    fn test_insert_simple_overwrite() {
+        let mut statements = PreparedStatements::default();
+
+        statements.insert_simple("upd", "UPDATE one SET x = 1");
+        let memory_used = statements.memory_used();
+
+        statements.insert_simple("upd", "UPDATE two SET x = 22");
+        assert_eq!(
+            statements.simple_query("upd"),
+            Some("UPDATE two SET x = 22")
+        );
+        assert_eq!(
+            statements.memory_used(),
+            memory_used + "UPDATE two SET x = 22".len() - "UPDATE one SET x = 1".len()
+        );
+
+        statements.close_all();
+        assert_eq!(statements.simple_query("upd"), None);
+        assert_eq!(statements.memory_used(), 0);
+    }
+
+    #[test]
     fn test_counted_only_once_per_client() {
         let mut statements = PreparedStatements::default();
 
--- a/pgdog/src/frontend/router/context.rs
+++ b/pgdog/src/frontend/router/context.rs
@@ -2,7 +2,7 @@ use super::{Error, ParameterHints};
 use crate::{
     backend::{Cluster, Schema},
     frontend::{
-        BufferedQuery, ClientRequest,
+        BufferedQuery, ClientRequest, PreparedStatements,
         client::{Sticky, TransactionType},
         router::Ast,
     },
@@ -37,6 +37,9 @@ pub struct RouterContext<'a> {
     pub schema: Schema,
     /// Original client request.
     pub client_request: &'a ClientRequest,
+    /// Client's prepared statements, used to route `EXECUTE`
+    /// based on the statement behind the name.
+    pub prepared_statements: Option<&'a mut PreparedStatements>,
 }
 
 impl<'a> RouterContext<'a> {
@@ -65,7 +68,17 @@ impl<'a> RouterContext<'a> {
             ast: buffer.ast.clone(),
             schema: cluster.schema(),
             client_request: buffer,
+            prepared_statements: None,
         })
+    }
+
+    /// Give the router access to the client's prepared statements.
+    pub fn with_prepared_statements(
+        mut self,
+        prepared_statements: &'a mut PreparedStatements,
+    ) -> Self {
+        self.prepared_statements = Some(prepared_statements);
+        self
     }
 
     pub fn in_transaction(&self) -> bool {
--- a/pgdog/src/frontend/router/context.rs
+++ b/pgdog/src/frontend/router/context.rs
@@ -2,7 +2,7 @@ use super::{Error, ParameterHints};
 use crate::{
     backend::{Cluster, Schema},
     frontend::{
-        BufferedQuery, ClientRequest,
+        BufferedQuery, ClientRequest, PreparedStatements,
         client::{Sticky, TransactionType},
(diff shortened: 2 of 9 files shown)


sgrif:
This touches the parser, which we are in the middle of a major rewrite of. This isn't likely to get reviewed until after that is done (we are hoping next week, but take that with a huge grain of salt)

pgdog/src/frontend/prepared_statements/mod.rs
```diff
@@ -32,6 +32,10 @@ fn str_mem(s: &str) -> usize {
 pub struct PreparedStatements {
     pub(super) global: Arc<RwLock<GlobalCache>>,
     pub(super) local: HashMap<String, String>,
+    /// SQL-level `PREPARE` statements sent over the simple protocol,
+    /// name -> statement text. Tracked so `EXECUTE` can be routed
+    /// based on the statement behind the name.
+    pub(super) simple: HashMap<String, String>,
```
levkk: I'm not sure this needs its own dedicated cache. You could extract the statement from `PREPARE` and store it in the global cache, e.g.:

```sql
PREPARE __sqlx_1 AS SELECT $1;
```

when extracted becomes:

| Name | Statement |
|-|-|
| `__sqlx_1` | `SELECT $1`|

You can create a `Parse` from this and insert it into the regular cache, e.g.:

```rust
let parse = Parse::named("__sqlx_1", "SELECT $1");
prepared_statements.insert(&mut parse);
```

When capturing `EXECUTE`, you can now fetch the original statement by calling:

```rust
let name = /* get name from EXECUTE using parser */;
prepared_statements.parse(name);
```

We do this already in the rewriter here: https://github.com/pgdogdev/pgdog/blob/bafec81f1898e96d4a419d4dfe711b7d090b5566/pgdog/src/frontend/router/parser/rewrite/statement/simple_prepared.rs#L77

pgdog/src/frontend/router/parser/query/execute.rs
```diff
@@ -0,0 +1,107 @@
...
+                Some(query) => query.to_string(),
+                None => return false,
+            }
+        };
+
+        Self::statement_omnisharded(&query, schema)
+    }
+
+    fn statement_omnisharded(query: &str, schema: &ShardingSchema) -> bool {
+        let Ok(ast) = pg_raw_parse::parse(query) else {
```
levkk: This should be using the statement cache, so we don't have to parse the statement on each execution.

pgdog/src/frontend/prepared_statements/mod.rs
mscrivo: thanks Lev! I'm reworking this now since I found some other issues as well. Sorry for the churn

levkk:
It's a good start, but I think the correct solution here is to handle `EXECUTE` in sharded databases overall. While this works for omni tables, we should handle this for sharded tables, too.

pgdog/src/frontend/prepared_statements/mod.rs
mscrivo: Done

pgdog/src/frontend/router/parser/query/execute.rs
mscrivo: Done

mscrivo:
> It's a good start, but I think the correct solution here is to handle `EXECUTE` in sharded databases overall. While this works for omni tables, we should handle this for sharded tables, too.

@levkk I'll leave it with you and @sgrif at this point. I don't have enough context with the codebase or experience with rust, so I'm in over my head at this point :)  It does demonstrate the outcome I'd like to see, so if you all close this PR and rework it entirely, that'd be fine with me as well.

levkk:
> > It's a good start, but I think the correct solution here is to handle `EXECUTE` in sharded databases overall. While this works for omni tables, we should handle this for sharded tables, too.
> 
> @levkk I'll leave it with you and @sgrif at this point. I don't have enough context with the codebase or experience with rust, so I'm in over my head at this point :) It does demonstrate the outcome I'd like to see, so if you all close this PR and rework it entirely, that'd be fine with me as well.

I'm not opposed to merging this (with maybe some minor changes) to unblock you guys! Let me know how soon you need this and we'll prioritize. We can always add support for sharded `EXECUTE` as a follow up.

mscrivo:
> > > It's a good start, but I think the correct solution here is to handle `EXECUTE` in sharded databases overall. While this works for omni tables, we should handle this for sharded tables, too.
> > 
> > 
> > @levkk I'll leave it with you and @sgrif at this point. I don't have enough context with the codebase or experience with rust, so I'm in over my head at this point :) It does demonstrate the outcome I'd like to see, so if you all close this PR and rework it entirely, that'd be fine with me as well.
> 
> I'm not opposed to merging this (with maybe some minor changes) to unblock you guys! Let me know how soon you need this and we'll prioritize. We can always add support for sharded `EXECUTE` as a follow up.

as soon as you're able I guess. We're not in any particular rush, but the problem on our end is that there's still a rather large bucket of test failures we're dealing with and so it makes it difficult to triage and categorize them. ie are they from this issue, another fixed or still open issue, etc. So the more we knock out, the easier that process becomes.

pgdog/src/frontend/router/parser/query/execute.rs
```diff
@@ -0,0 +1,127 @@
...
+            user: context.router_context.cluster.user(),
+            search_path: context.router_context.parameter_hints.search_path,
+        };
+        let Ok(ast) = Cache::get().query(
+            &BufferedQuery::Prepared(parse),
+            &ast_context,
+            prepared_statements,
+        ) else {
+            return false;
+        };
```
sgrif, line 113: Should this be returning the error? Should we at least log it?

pgdog/src/frontend/router/parser/query/execute.rs
```diff
@@ -0,0 +1,127 @@
...
+    /// left out conservatively; its row counts keep aggregating.
+    fn executed_statement_omnisharded(
+        stmt: &nodes::ExecuteStmt,
+        context: &mut QueryParserContext,
+    ) -> bool {
+        if context.sharding_schema.tables.omnishards().is_empty() {
+            return false;
+        }
+
+        let Some(name) = stmt.name() else {
```
sgrif, line 88: This can be `.expect`, execute statements always have a name.

pgdog/src/frontend/router/parser/query/execute.rs
```diff
@@ -0,0 +1,127 @@
...
+            return false;
+        };
+        let Some(prepared_statements) = context.router_context.prepared_statements.as_deref_mut()
```
sgrif, line 91: Would this returning `None` indicate a bug on our end? Should this be `.expect` instead?

pgdog/src/frontend/router/parser/query/execute.rs
```diff
@@ -0,0 +1,127 @@
...
+        };
+        let Ok(ast) = Cache::get().query(
+            &BufferedQuery::Prepared(parse),
+            &ast_context,
+            prepared_statements,
+        ) else {
+            return false;
+        };
+
+        let Some(root) = ast.ast.stmts().next() else {
```
sgrif, line 115: This can also be `.expect`, we should never have an empty query here.

pgdog/src/frontend/router/parser/query/execute.rs
```diff
@@ -0,0 +1,127 @@
...
+            return false;
+        };
+        if !matches!(
+            root,
+            Node::InsertStmt(_) | Node::UpdateStmt(_) | Node::DeleteStmt(_)
+        ) {
+            return false;
+        }
+
+        StatementParser::new(root, None, &context.sharding_schema, None).is_all_omnisharded()
```
sgrif, line 125:
```suggestion
        match root {
            Node::InsertStmt(_) | Node::UpdateStmt(_) | Node::DeleteStmt(_) => return false,
            _ => StatementParser::new(root, None, &context.sharding_schema, None).is_all_omnisharded(),
        }
```

pgdog/src/frontend/router/context.rs
```diff
@@ -42,6 +42,9 @@ pub struct RouterContext<'a> {
     /// reads these before the lookup cache, so a second routing pass
     /// after resolving lookups can't miss.
     pub resolved_lookups: ResolvedLookups,
+    /// Client's prepared statements, used to route `EXECUTE`
+    /// based on the statement behind the name.
+    pub prepared_statements: Option<&'a mut PreparedStatements>,
```
sgrif, line 47: What do you think about setting this to `Default::default()` in `new` and removing the `Option` so callers can rely on this always being set?

sgrif (changes requested):
[thing this asked for is already present elsewhere in the codebase]

sgrif (approved):
...Disregard my previous review, everything I asked for is already handled outside of the code this touches.

levkk:
This PR is being taken over by #1368

PR #1180, opened 2026-07-10
Port StatementRewrite::split_insert_stmt to the new parser

sgrif:
This is mostly a straight port, with some minor control flow changes to make it read better. The only notable behavior change is that the params it stores are one indexed instead of zero indexed, and that it doesn't duplicate parameters that are used multiple times. This makes it behave nearly identically to the statement struct we use for multi-step updates, which I intend to unify this with in the long term.


levkk (approved):
really cool

meskill (approved):

PR #1181, opened 2026-07-13
Don't panic on missing binds in insert_split

sgrif:
The code on the old parser, surprisingly, treated this operation as infallible, and just sent an incorrect number of parameters to the shards (which given that we don't behave transactionally for split inserts, could lead to a partial insert). This doesn't fix the lack of transactional behavior, but at least in the new port we won't send any rows at all if we don't have enough bind parameters

### DIFF
--- a/pgdog/src/frontend/router/parser/rewrite/statement/insert.rs
+++ b/pgdog/src/frontend/router/parser/rewrite/statement/insert.rs
@@ -589,13 +589,13 @@ mod tests {
         );
 
         // First split uses params 0 and 1
-        let extracted = splits[0].extract_bind_params(&bind);
+        let extracted = splits[0].extract_bind_params(&bind).unwrap();
         assert_eq!(extracted.params_raw().len(), 2);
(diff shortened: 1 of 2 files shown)


pgdog/src/frontend/router/parser/rewrite/statement/insert.rs
```diff
@@ -103,28 +103,26 @@ impl InsertSplit {
...
+    fn extract_bind_params(&self, bind: &Bind) -> Result<Bind, Error> {
         let mut new = Bind::new_statement(self.statement_name().unwrap_or_default());
         for param in &self.params {
             let param = bind
-                .parameter(*param as usize - 1)
-                .expect("Remove when port done")
-                .ok_or(Error::MissingParameter(*param))
-                .expect("Remove when port done");
+                .parameter(*param as usize - 1)?
+                .ok_or(Error::MissingParameter(*param))?;
```
meskill, line 116: Would you mind add a small unit test for it? just to make sure

PR #1182, opened 2026-07-13
Remove the old parser's AST from `StatementRewrite`

sgrif:
It is no longer used, so we can get rid of it. I opted to change the signature of `maybe_rewrite` to take a mutable reference to the statement rather than copy it and return a new one. This resulted in a significant indentation change, but no actual code changes (beyond replacing a few `as_ref` and `as_mut`s with `stmt` and `stmt_mut`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3313,7 +3313,7 @@ dependencies = [
 [[package]]
 name = "pg_raw_parse"
 version = "0.1.0"
-source = "git+https://github.com/pgdogdev/pg_raw_parse.git?rev=6421337#6421337b411e91b4b0ca553b8146108c2a273667"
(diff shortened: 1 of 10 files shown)


levkk (approved):
This is because we couldn't mutate ASTs before, right? Nice. Bye bye pg_query, say hello to sage_query. No qualms with you naming the library after yourself.

sgrif:
> This is because we couldn't mutate ASTs before, right?

No, we could. It's purely a question of where `make::owned` is called. The reason I opted to move it up a level instead of returning the copied/mutated node from this function is multi-statement ASTs. This function only operates on a single node, but we still need to keep the unmanipulated nodes around, meaning it's not just a question of creating a new single statement list of the transformed node. We'd need to assign it to index 0 in the old list, which requires a mutable reference to it, which requires a `make::owned` block at that level.

PR #1187, opened 2026-07-13
Port QueryParser::set to the new parser

sgrif:
There's a few significant changes here. First, we're more explicitly checking for `SET TRANSACTION` by looking at the `kind` flag, rather than the name that happens to be set as an implementation detail. Second, we're much more intentional about where we expect the parsing functions to be called. We only expect to parse values for `VAR_SET_VALUE`. We explicitly look at the kind before calling that now, so that an unexpected condition is properly elevated as the error that it is, and not just silently passed through to the shard to leave things in a bad state.

For posterity, here is the expected (and implemented in the ported code) behavior for each type of set statement:

- `VAR_RESET | VAR_SET_DEFAULT`: These are different spellings of the same thing, which we treat as a `Set` command
- `VAR_SET_VALUE`: We expect either a constant or a type cast constant expression. We treat this as a `Set` command.
- `VAR_SET_MULTI`: This is `SET TRANSACTION`, and is passed through as `Query`
- `VAR_RESET_ALL`: `Command::ResetAll`
- `VAR_SET_CURRENT`: Unreachable, this is part of `ALTER DATABASE` DDL statements

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3313,7 +3313,7 @@ dependencies = [
 [[package]]
 name = "pg_raw_parse"
 version = "0.1.0"
-source = "git+https://github.com/pgdogdev/pg_raw_parse.git?rev=6b424c4#6b424c4510ef4e9c482ef8a7fd7246b66f15de0a"
+source = "git+https://github.com/pgdogdev/pg_raw_parse.git?rev=576226d#576226dd9d9e86539bc0b9d9baeb16c3601b346d"
 dependencies = [
  "bindgen 0.72.1",
  "cc",
--- a/pgdog/Cargo.toml
+++ b/pgdog/Cargo.toml
@@ -80,7 +80,7 @@ smallvec = "1"
 reqwest.workspace = true
 hex = "0.4"
 x509-parser = "0.18"
-pg_raw_parse = { git = "https://github.com/pgdogdev/pg_raw_parse.git", rev = "6b424c4", optional = true }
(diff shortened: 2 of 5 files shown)


sgrif:
(failure is just that I need to revert a line I had changed in the old parser code when I was planning on cfging individual lines. I ended up replacing every function wholesale so it's just a one line revert when I'm back at my computer)

levkk (approved):
As long as you checked that you covered all enum variants. Our arch kind of opts into things it cares about and lets everything else flow through as a query. In the case of SET, this kind of works I think, because we only care about the happy path and erroring out for the sad path doesn't need to happen afaik (I don't even know what the sad path is and it's not clear from the PR what bad state you discovered).

sgrif:
I don't believe there's currently a bad state that I've uncovered, just the code being written in such a way that a bug could cause things to slip through unnoticed in the future. The consequences would be the bug we observed before handling multiple set_configs in a single query, where it would pass through to the shard as a session config but not pin the connection

> Our arch kind of opts into things it cares about and lets everything else flow through as a query.

That's what this does at the top level set function. It's after that where we are very strict about what conditions should be possible. Basically what I mean is that we've opted into what we support, but now we assert that we actually handle all cases for the things we support, rather than silently skipping

levkk:
Awesome possum

sgrif:
Here's a better way of phrasing my concern with the old code. It's not the letting things we don't support fall through, it's the functions returning `Option` where `None` is only cases that should never happen, and then calling code uses `filter_map` resulting in us silently doing the wrong thing. It's a consistent pattern I see across the code base that I'm trying to improve

PR #1200, opened 2026-07-14
Port QueryParser::ddl to the new parser

sgrif:
This contains a minor behavioral change for `DO` statements, ensuring we operate on the actual code block and not the language declaration. The behavior for this type of node is somewhat nonsensical though, as we're parsing individual lines which may not be a complete statement, and the body is always going to be a language we can't actually parse. Even if the language is plpgsql, correctly finding a line that is a valid DDL statement is unlikely to succeed

Other than that, the only other change is some refactoring of tests.

### DIFF
--- a/pgdog/src/frontend/router/parser/query/ddl.rs
+++ b/pgdog/src/frontend/router/parser/query/ddl.rs
@@ -1,14 +1,11 @@
-use pg_query::parse;
-#[cfg(feature = "new_parser")]
-use pg_query::protobuf::{ObjectType, RangeVar};
-
 use super::*;
 
 impl QueryParser {
     /// Handle DDL, e.g. CREATE, DROP, ALTER, etc.
     pub(super) fn ddl(
         &mut self,
-        node: &Option<NodeEnum>,
+        #[cfg(feature = "new_parser")] node: Node<'_>,
+        #[cfg(not(feature = "new_parser"))] node: &Option<NodeEnum>,
         context: &mut QueryParserContext<'_>,
     ) -> Result<Command, Error> {
         let command = Self::shard_ddl(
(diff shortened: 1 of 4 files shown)


pgdog/src/frontend/router/parser/query/ddl.rs
```diff
@@ -139,46 +136,37 @@ impl QueryParser {
...
-            Some(NodeEnum::VacuumStmt(stmt)) => {
-                for rel in &stmt.rels {
-                    if let Some(NodeEnum::VacuumRelation(ref stmt)) = rel.node {
-                        shard =
-                            Self::shard_ddl_table(&stmt.relation, schema)?.unwrap_or(Shard::All);
-                    }
+            Node::VacuumStmt(stmt) => {
+                for rel in stmt.rels() {
+                    // FIXME: This almost certainly needs to be combining
+                    // shards, not setting it to the target of the last
```
levkk, line 142: This check is for schema-based sharding only really, e.g., `VACUUM customer_b.users` should be routed to shard that contains `customer_b` schema: https://docs.pgdog.dev/configuration/pgdog.toml/sharded_schemas/

IIRC, postgres doesn't allow more than one "command" per statement. Maybe that's wrong?

levkk (approved):

pgdog/src/frontend/router/parser/query/ddl.rs
sgrif: What should we do for `VACUUM customer_a.users, customer_b.users`? Or `VACUUM customer_b.users, unsharded_table`? That's what this comment is referring to

levkk: Oh yeah, good point. I don't think that statement will actually work, since the tables are located on different shards. Probably just send it as-is to all shards and let Postgres return an error.

levkk: Edge case being both schemas on the same shard...which would "just work" in the current design, so maybe we just leave it as-is?

sgrif: I think the best solution short term would be to send to a single shard if we calculate the same shard for all relations, and fall back to Shard::All to let PG error if any differ. Long term we probably just need to split this into multiple statements

levkk: Agreed. The rewriter should actually probably do this for all SQL, incl the ones  separated by a semicolon.

sgrif: Separated by a semicolon would mean it's a separate `RawStmt` at the top level, which is dictated by [this](https://github.com/pgdogdev/pgdog/blob/7108ac7327acbfe18c716c0106e79fa966169367/pgdog/src/frontend/router/parser/query/mod.rs#L298-L303). I agree we should handle it, but it's a different issue than something like VACUUM which is a single statement operating on multiple tables

levkk: Correct

PR #1201, opened 2026-07-14
fix(replication): fix wait_for_replication logic for unrelated changes

meskill:
Send the updates to the walsender also when we advance the LSN track. 

That updates the walsender pointer more often and that makes our check for replication lag more relevant. The issue with the test was that it were sent only when walsender was asking for the reply, but with default settings it happens like around once in 30sec, just like the test initial waiting time - so it passed sometimes. Too bad on my local machine I had the timeout like 1sec (thanks clunker I guess) and I haven't caught the issue earlier.

### DIFF
--- a/pgdog/src/backend/replication/logical/orchestrator.rs
+++ b/pgdog/src/backend/replication/logical/orchestrator.rs
@@ -910,7 +910,7 @@ mod tests {
 
         // 30s margin: keepalive cadence (wal_sender_timeout) is not bounded by
         // the 1s sleep, so give confirmed_flush_lsn room to advance.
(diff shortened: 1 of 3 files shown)


levkk (approved):
😆

meskill:
> 😆

don't look, it's a highly experimental and explosive!

meskill:
I'm not sure what is going on, but the unit tests started to run really long for like 20minutes...

levkk:
> I'm not sure what is going on, but the unit tests started to run really long for like 20minutes...

I think we are getting throttled by Blacksmith. I've told them about this in the past, but haven't heard back. I think they are just overscheduling jobs.

levkk (approved):
perfect!

PR #1202, opened 2026-07-14
Fix test that was mostly testing pg_query, not our code

sgrif:


levkk (approved):
I wrote a blog post about this: https://levkk.bearblog.dev/not-tdd/

sgrif:
I don't think we should have tests in our codebase that are purely testing whether a library works. If a library were to being behaving differently in a future version, that's something our integration tests should catch. And generally speaking we'd expect a compiler error, not just a silent runtime behavior change

levkk:
> I don't think we should have tests in our codebase that are purely testing whether a library works. If a library were to being behaving differently in a future version, that's something our integration tests should catch. And generally speaking we'd expect a compiler error, not just a silent runtime behavior change

Aight! Saving this comment for posterity. 🤣

sgrif:
I know I wrote a blog post that goes into more details like 12 years ago
but I'm having trouble finding it lol

levkk:
Everyone needs a bear blog. Rawr

PR #1204, opened 2026-07-14
Port MultiTenantCheck to new parser

sgrif:
Note that a good bit of this is actually dead code, as our code currently appears to incorrectly only ever be running this check for queries with a single table in the from clause. We have no test coverage for this, and I'm not familiar enough with this part of the code to be able to quickly fill that test coverage in, so I've left the bug in place for the time being with the intention to come back to it in the near future.

I left the dead code in place because I don't want to get bogged down by a major refactor while in the middle of a port


levkk (approved):
yeah dont worry about this one; it's obscure, not well documented and not really that useful.

sgrif:
In that case are you cool with me killing all the dead code and making it clear that it only operates on single tables?

PR #1207, opened 2026-07-15
Use a less horrendously unsafe FFI interface for plugins

sgrif:
The previous interface was a bit of a mess. All the types were defined in C for no clear reason, everything was stuck behind `*void` for no clear reason, pointers were cast into things like `Vec` in blatant disregard of its documented invariants, and all lifetime information was erased making it impossible to expose a safe interface around any of it. The old code also seemed to think that you can pass repr-Rust structs around as long as they were compiled with the same compiler version, which is absolutely not true and results in instant UB, regardless of whether you're casting to a void* and back or not.

This new interface is simpler, looks more like ideomatic Rust, and is completely safe with a single caveat.

Plugins are expected to implement the `Plugin` trait, and we have a dead simple macro that just shoves a vtable into a global static, which is where we load things from. We mostly pass types as-is, except for strings and slices which have no defined repr. This unfortunately requires an intermediate allocation for the bind parameters, as there's no FFI-safe way to represent `&[&[u8]]`. (The old code also had this problem, it was just invoking undefined behavior).

The one big caveat here is that the parse tree for the old parser is not FFI safe and is still being passed as-is. This is UB, and we are continuing to rely on the implementation detail of our compiler invocations not currently resulting in any changes in layout (at least as far as we know. It's UB, it absolutely could just be happening without us realizing it).

The only way to make this part safe would be to go through a full protobuf ser/de step at the FFI boundary. That would suck, we're getting ready to throw out the old parser anyway, and the new one is fully FFI safe, so I've chosen to just pretend this problem doesn't exist.

This commit updates docs for the individual structures/functions it touches, but there are likely higher level docs that are now out of date and need to be updated in a follow-up PR


levkk (approved):
Amaze amaze amaze

meskill:
Have you considered any libraries to implement this. For example https://crates.io/crates/stabby

I had no experience with it or similar, but it looks like that may help to validate layout mismatch and other checks. Also, they have "ABI stable equivalent" of basic types like str https://docs.rs/stabby/36.2.2/stabby/str/struct.Str.html

pgdog-plugin/src/string.rs
```diff
@@ -15,50 +15,49 @@
...
-        PdStr {
-            data: value.as_ptr() as *mut c_void,
-            len: value.len(),
-        }
-    }
+#[repr(C)]
+/// A wrapper around a rust `&str`
+#[derive(Clone, Copy)]
+pub struct PdStr<'a> {
+    data: Option<&'a u8>,
```
meskill, line 24: huh? should it be some kind of pointer and not &u8? I guess you are using this as pointer anyway, but maybe?

sgrif:
> Have you considered any libraries to implement this. For example https://crates.io/crates/stabby

I haven't, but I'd definitely love to have something that validates only ABI-stable types get passed. I thought clippy had a lint for this, but I must have dreamed it.

That's something I'll look at in a follow-up PR though, as it's not something that's worthwhile while we still have the old parser in place, as we know it's not ABI-stable

pgdog-plugin/src/string.rs
```diff
@@ -15,50 +15,49 @@
...
     type Target = str;
 
     fn deref(&self) -> &Self::Target {
+        // SAFETY: The returned lifetime is shorter than 'a by definition.
+        // The only way to construct this is from a valid string
         unsafe {
-            let slice = from_raw_parts::<u8>(self.data as *mut u8, self.len);
-            from_utf8_unchecked(slice)
+            str::from_utf8_unchecked(slice::from_raw_parts(
+                self.data.map(ptr::from_ref).unwrap_or_default(),
```
meskill: what if the data == None? What'd happen here? The docs says "data must be non-null, valid for reads for len * size_of::<T>() many bytes,
and it must be properly aligned. ", but we get the null pointer here cuz of `unwrap_or_default()` and we get None if we have empty string, right?

pgdog/src/backend/pool/connection/mirror/mod.rs
```diff
@@ -234,7 +234,7 @@ mod test {
 
     // This test relies on data written to multiple databases.
     // It's flaky. Not sure why yet.
-    #[pgdog_plugin::macros::flaky]
+    #[pgdog_macros::flaky]
```
meskill, line 237: TIL we have flaky macro for tests

plugins/pgdog-example-plugin/src/lib.rs
```diff
@@ -8,35 +8,38 @@
...
-///
-/// N.B. This is sync and will prevent PgDog from exiting if it gets stuck.
-///
-#[macros::fini]
-fn shutdown() {}
+    /// If defined, this function is called on every query going through PgDog.
+    ///
+    /// It's provided with the AST generated by pg_query and context on how many databases
+    /// PgDog is proxying.
+    fn route(context: Context<'_>) -> Route {
```
meskill, line 28: Why do we not add the extern here for route? isn't rust fns are not ABI compatible?

pgdog-plugin/src/string.rs
sgrif: `&u8` is a pointer :)

pgdog-plugin/src/string.rs
sgrif: Good catch. We need to use `ptr::dangling` not default

plugins/pgdog-example-plugin/src/lib.rs
sgrif: I didn't use extern here because it's not strictly necessary, since this isn't a function pointer passed across an ABI boundary. I could for consistency if this bothers folks

sgrif: (`route_raw` is the ABI stable version, and all it does is immediately call this function)

pgdog-plugin/src/string.rs
sgrif: And this is why code review is important 💜

pgdog/src/backend/pool/connection/mirror/mod.rs
sgrif: Yeah, seems bad 🤣

pgdog-plugin/src/parameters.rs
```diff
@@ -48,18 +35,23 @@ pub enum ParameterValue<'a> {
...
+    len: usize,
+    /// The pointer to the data
+    data: Option<&'a u8>,
 }
 
-impl Parameter {
+impl<'a> Parameter<'a> {
+    /// Construct this from the given slice
+    pub fn new(data: Option<&'a [u8]>) -> Self {
+        let len = data.map(|d| d.len()).unwrap_or(usize::MAX);
```
meskill, line 50: I'd say the Option<usize> would be better to represent additional states for the null state instead of usize::MAX, it seems like checked properly anyway

pgdog-plugin/src/parameters.rs
```diff
@@ -70,176 +62,134 @@ impl Parameter {
...
-impl Clone for Parameters {
-    fn clone(&self) -> Self {
+impl<'a> From<Option<&'a [u8]>> for Parameter<'a> {
+    fn from(value: Option<&'a [u8]>) -> Self {
         Self {
-            params: self.params.clone(),
-            format_codes: self.format_codes.clone(),
-            borrowed: false,
+            len: value.map(|s| s.len()).unwrap_or(usize::MAX),
+            data: value.and_then(|s| s.first()),
```
meskill, line 117:
```
Self::new(value)
```

pgdog-plugin/src/string.rs
meskill: yeah... but it's confusing for me, since it feels easier to mess up with the reference to u8.

is it to bypass the unused lifetime otherwise? I guess PhantomData is usually used for this?

pgdog/src/plugin/mod.rs
```diff
@@ -38,68 +37,50 @@ pub fn load(config: &Config) -> Result<(), libloading::Error> {
...
-            } else {
+            if !same_major_minor(&plugin_lib.pgdog_plugin_api_version(), pgdog_plugin_api_version) {
                 warn!(
-                    "plugin {} doesn't expose its plugin API version, please update version of pgdog-plugin crate in your plugin",
-                    plugin_lib.name()
+                    "skipping plugin \"{}\" because it was compiled with different plugin API version ({})",
+                    plugin.name,
+                    &*plugin_lib.pgdog_plugin_api_version()
                 );
+                return None;
```
meskill, line 72: ah, my first contribution to the project. I'm glad it survived the refactor 😄

plugins/pgdog-example-plugin/src/lib.rs
meskill: ah, I see now. Could you elaborate why do we have separate route and route_raw?

pgdog-plugin/src/parameters.rs
sgrif: I chose `usize::MAX` as a sentinel because the type we're mapping from uses `-1_i32`, and this is the value that'd map to that. This isn't the same as no data, this is "a parameter that was passed with the value of SQL NULL". For data-types like `text`, `bytea`, you can still have a valid value with no data

plugins/pgdog-example-plugin/src/lib.rs
sgrif: Because `Context` isn't ABI stable, we need a place that we convert the len/ptr pairs into slices. `route_raw` is where we do that.

pgdog-plugin/src/string.rs
sgrif: It's not to bypass the unused lifetime, it's to make it clear that *this pointer specifically* is a valid, non-null, aligned pointer with a given lifetime. It's retaining as much information about the type as humanly possible in an abi-stable way

plugins/pgdog-example-plugin/src/lib.rs
meskill: oh, right, and the repr(C) on `Context` does not mean it's ABI stable if the inner fields are not, right?

sgrif: Correct. In fact, I should just remove that

pgdog-plugin/src/parameters.rs
meskill: you mean we have 3 options: null, empty, some data?

We infer this from the input Option<&[u8]> that basically encodes this and I my suggestion was to use the `len: Option<usize>` to encode the null state instead of MAX

pgdog-plugin/src/string.rs
meskill: interesting

pgdog-plugin/src/parameters.rs
sgrif: I wanted to keep this struct the same size as a `&[]`, and that would increase the size of this struct from 2 words to 3. I guess it's not that important

PR #1209, opened 2026-07-15
Port the plugin system to the new parser

sgrif:
This ports both the plugin system, and the in-tree plugins to use the new parser. Since `pgdog-plugin` is small enough to port the whole crate at once, I've ensured that if `new_parser` is enabled, `pg_query` isn't linked at all (at least not by that crate)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3313,7 +3313,7 @@ dependencies = [
 [[package]]
 name = "pg_raw_parse"
 version = "0.1.0"
-source = "git+https://github.com/pgdogdev/pg_raw_parse.git?rev=576226d#576226dd9d9e86539bc0b9d9baeb16c3601b346d"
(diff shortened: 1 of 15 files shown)


levkk (approved):
brilliant

levkk:
we'll need to build the libs in docker with the feature flag on for both plugins and pgdog...I think we just run `cargo build` for everything so it'll be straightforward. Just a headsup.

sgrif:
Yup, for the beta any `cargo build` will need to be run with `cargo build --no-default-features --features new_parser`

sgrif:
> we'll need to build the libs in docker with the feature flag on for both plugins and pgdog

Yeah I opted to not bother with checking that the flags match since as you said, everything is built in-tree, and this is going away in a week anyway

PR #1210, opened 2026-07-15
Port rewrite_ast_limit_offset to the new parser

sgrif:
This has a minor behavioral change of rewriting to `LIMIT x` instead of `LIMIT x OFFSET 0`


levkk (approved):
this is correct

sgrif:
> this is correct

I always am 🙃

PR #1216, opened 2026-07-15
Remove fingerprinting from our codebase

sgrif:
We agreed that we hate it, it's not currently actually used anywhere, and if we want something like this in the future we can probably implement it as a plugin.

There's likely more dead code in Cache as a result of this but I'll come back and depub that another time


levkk (approved):
first breaking change, but im happy we're removing this. manual queries are a silly feature that stops working immediately after attempting it in the real world.

it's also not documented, so if someone is using it...well...

levkk:
> There's likely more dead code in Cache as a result of this but I'll come back and depub that another time

yes, and in metrics (finterprint timing, count is tracked separately from query parsing time because its so much slower)

sgrif:
> first breaking change

Tbh this further reinforces my belief that we should do stuff like this as plugins. Much easier to call something an experiment and/or have its stability be independent of the main thing

sgrif:
> first breaking change, but im happy we're removing this. manual queries are a silly feature that stops working immediately after attempting it in the real world.
> 
> it's also not documented, so if someone is using it...well...

Yeah and given the only thing it does is force round robin routing I would be shocked if anyone was trying to use it

meskill:
🫡

meskill (approved):

PR #1218, opened 2026-07-15
CFG/remove some dead code

sgrif:


levkk (approved):
do the cache fingerprint metrics too

PR #1224, opened 2026-07-16
Generate a valid default node-id in UniqueId

sgrif:
This code clearly intended to have a fallback to a random number if the environment variable `NODE_ID` wasn't set. This was fundamentally broken as what it was generating was a hex string and it was trying to parse it as decimal. And even if it was parsing it correctly, the max node ID is
1023. So we generate a random number between 0 and 1023

### DIFF
--- a/pgdog/src/util.rs
+++ b/pgdog/src/util.rs
@@ -114,18 +114,13 @@ pub fn random_string(n: usize) -> String {
         .collect()
 }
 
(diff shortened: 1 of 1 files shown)


levkk (commented):
That's not what we want though. The NODE_ID must come from the environment and be correct. If it's not, we must error our when attempting to generate unique id in the rewriter. Rewriter that doesn't need unique id must not call any of this code at all

PR #1225, opened 2026-07-16
Don't attempt to construct a unique_id generator eagerly

sgrif:
We had lifted the construction of the generator out of the `transform` block because that function needs to be infallible. Unfortunately, that is where the error from the lack of a NODE_ID variable being set occurs, and moving that to the unique_id generation would have the same problem but require changing more code.

So we construct the generator at the last moment, as the old code did, rather than injecting it.

### DIFF
--- a/pgdog/src/frontend/router/parser/rewrite/statement/mod.rs
+++ b/pgdog/src/frontend/router/parser/rewrite/statement/mod.rs
@@ -150,26 +150,28 @@ impl<'a> StatementRewrite<'a> {
 
         // Track the next parameter number to use
         let mut next_param = plan.params as i32 + 1;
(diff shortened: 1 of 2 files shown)


pgdog/src/frontend/router/parser/rewrite/statement/mod.rs
```diff
@@ -150,26 +150,28 @@ impl<'a> StatementRewrite<'a> {
...
+                    }
+                    Err(e) => {
+                        err = Some(e);
+                        None
+                    }
+                    Ok(None) => Some(node),
                 }
             }),
         );
+        if let Some(err) = err {
```
sgrif, line 172: We have go at home.

Go at home:

levkk (approved):
yup

PR #1231, opened 2026-07-17
fix: don't drop the Describe response on deferred transactions

### LINKED ISSUE #1196: Deferred BEGIN over extended protocol drops the Describe response → client gets "unexpected message from server"
**PgDog version**

  `v0.1.47`

  **Description**

  When `query_parser` intercepts a `BEGIN` (levels `on`, `auto`, or `session_control*`) and defers it, PgDog synthesizes the reply to the client instead of forwarding to Postgres. For clients using the **extended query
  protocol** (e.g. Rust `tokio-postgres`), the synthetic reply omits the response to the `Describe` message.

  `extended_transaction_reply` in `pgdog/src/frontend/client/query_engine/start_transaction.rs` walks the client's extended-protocol messages but the `Describe` arm is a no-op:

  ```rust
  match message.code() {
      'P' => reply.push(ParseComplete.message()?),   // Parse    -> ParseComplete
      'B' => reply.push(BindComplete.message()?),    // Bind     -> BindComplete
      'D' | 'H' => (),                               // Describe -> NOTHING  (bug)
      'E' => reply.push(/* CommandComplete */),      // Execute  -> CommandComplete
      'S' => { /* ... ReadyForQuery ... */ }         // Sync     -> ReadyForQuery
      c => return Err(Error::UnexpectedMessage(c)),
  }
  ```

  Postgres answers a `Describe` with `RowDescription('T')` or `NoData('n')`; PgDog sends neither.

  **What happens:** a `tokio-postgres` client issuing `BEGIN` over the extended protocol sends `Parse, Bind, Describe, Execute, Sync` and expects `ParseComplete, BindComplete, NoData, CommandComplete, ReadyForQuery`. PgDog
  instead sends `ParseComplete, BindComplete, <nothing>, CommandComplete, ReadyForQuery`. After its `Describe` the client is waiting for `NoData`/`RowDescription`, reads `CommandComplete`, and fails immediately with
  `unexpected message from server`.

  **What should happen:** `BEGIN` over the extended protocol should behave as if sent directly to Postgres, then the client receives `NoData` in response to its `Describe`. Direct-to-Postgres works, and `query_parser = "off"`
  works (pure passthrough).

  **Reproduce:**
  1. PgDog 0.1.47, `pooler_mode = "transaction"`, `query_parser = "auto"` (or `"on"`), single unsharded primary.
  2. From `tokio-postgres`, issue `BEGIN` over the extended protocol (e.g. `client.query("BEGIN", &[]).await`).
  3. The call fails immediately with `unexpected message from server`.

  **Proposed fix:** the `'D'` arm should emit `NoData` (transaction-control statements return no rows):

  ```rust
  'D' => reply.push(NoData.message()?),
  'H' => (),  // Flush needs no synthetic reply
  ```

  **Related:** #999 also involves extended-protocol transactions but presents as a ~10s hang / `bad connection`; this one is an immediate `unexpected message from server` traced to the `'D'` arm dropping the Describe
  response. Possibly the same area, flagging in case it's separately fixable.

**Configuration**

  Minimal repro slice (the Windmill metadata DB). `pgdog.toml`:

  ```toml
  [general]
  query_parser = "auto"   # default — bug triggers here (and on "on"); "off" is the workaround we applied

  [[databases]]
  name = "db_windmill"
  host = "<redacted-host>"
  port = 5432
  # no pooler_mode override -> transaction pooling (default)
  ```

  `users.toml`:

  ```toml
  [[users]]
  name = "app_user"
  database = "db_windmill"
  pool_size = 100
  min_pool_size = 100
  pooler_mode = "transaction"   # default; not overridden for this database
  # password redacted
  ```

kajuyerim:
Fixes #1196.

## Problem

When the query parser defers a transaction-control statement (`BEGIN`/`COMMIT`/`ROLLBACK`) because no server is attached yet, `extended_transaction_reply` builds the reply but ignores the client's `Describe`. Extended-protocol clients (tokio-postgres and friends) get no `ParameterDescription`/`NoData` and abort with `unexpected message from server`.

## Fix

Answer the `Describe`. Always sending `NoData` isn't enough: a **statement** describe (`'S'`) must be answered with `ParameterDescription` first. tokio-postgres sends one while preparing and expects `ParameterDescription` before anything else. 
So:

- statement `Describe` → `ParameterDescription` + `NoData`
- portal `Describe` → `NoData`

matching the [extended-query protocol][pg-ext] for utility statements (no params, no rows).

[pg-ext]: https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-EXT-QUERY

## Tests

The existing `tokio_postgres` tests only run `SELECT`/`INSERT`/`COPY`, so nothing exercised transaction control over the extended protocol. Added `tokio_postgres/transaction.rs`, plus protocol-level unit tests in `query_engine/test/extended_transaction.rs`.

### DIFF
--- a/integration/rust/tests/tokio_postgres/mod.rs
+++ b/integration/rust/tests/tokio_postgres/mod.rs
@@ -2,3 +2,4 @@ pub mod copy;
 pub mod nulls;
 pub mod prepared;
 pub mod select;
+pub mod transaction;
--- a/integration/rust/tests/tokio_postgres/transaction.rs
+++ b/integration/rust/tests/tokio_postgres/transaction.rs
@@ -1,0 +1,19 @@
+use crate::setup::connections_tokio;
+
+#[tokio::test]
+async fn test_transaction_control() {
+    let conns = connections_tokio().await;
+    for conn in conns {
+        let begin = conn.prepare("BEGIN").await.unwrap();
+        assert!(begin.params().is_empty());
+        assert!(begin.columns().is_empty());
+
+        for _ in 0..25 {
+            conn.execute("BEGIN", &[]).await.unwrap();
+            conn.execute("COMMIT", &[]).await.unwrap();
+
+            conn.execute("BEGIN", &[]).await.unwrap();
+            conn.execute("ROLLBACK", &[]).await.unwrap();
+        }
+    }
+}
--- a/pgdog/src/frontend/client/query_engine/start_transaction.rs
+++ b/pgdog/src/frontend/client/query_engine/start_transaction.rs
@@ -1,6 +1,9 @@
 use crate::{
     frontend::client::TransactionType,
-    net::{BindComplete, CommandComplete, NoticeResponse, ParseComplete, Protocol, ReadyForQuery},
+    net::{
+        BindComplete, CommandComplete, NoData, NoticeResponse, ParameterDescription, ParseComplete,
+        Protocol, ProtocolMessage, ReadyForQuery,
+    },
 };
 
 use super::*;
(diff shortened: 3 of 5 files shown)


kajuyerim:
Rebased onto latest main. 
Pre-rebase failure for reference (orphaned by the rebase now): [Actions/runs/29613579718]( https://github.com/pgdogdev/pgdog/actions/runs/29613579718 )

Turns out my branch was a commit behind and missing #1232 (hot mutex in comms), which lines up with that checkout timeout in test_schema_sharding. Pretty sure it was unrelated to this change, so hopefully the rebase sorts it out. Hope it's green now

levkk:
Good stuff, thanks!

PR #1233, opened 2026-07-18
feat: extend passthrough auth to mirror destinations

### LINKED ISSUE #1189: [Mirroring] Dynamic pool creation with passthrough auth
Automatically create the corresponding user/database connection pool on mirrors when using passthrough authentication with mirroring.

lalvania:
Closes #1189 — Support Mirrors with pass through auth.

`from_config` only provisions a mirror pool when the source and destination databases have matching `[[users]]` name sets. Passthrough only writes a `[[users]]` entry for the source side, so mirroring stays disabled. This patch also writes a matching entry for each mirror destination when passthrough runs, unlocking the name-match check. Any pre-existing destination entry is left alone.

## Rollout notes

For operators upgrading with a valid `users.toml`:

- No schema change — drop-in binary replacement, no `pgdog.toml` or `users.toml` edits required.
- Login for any `(name, database)` already in `users.toml` with a matching password is a zero-write, zero-reload no-op. Indistinguishable from pre-patch.
- The new code path only fires for a passthrough login whose `(name, database)` isn't in `users.toml`. If you have no `[[mirroring]]` blocks it's dormant regardless.
- The existing `mirroring disabled ... users don't match` warning should decrease (never increase) after upgrade. Increase = regression signal.
- Under `enabled_allow_change`, source-side rotation is unchanged. The mirror record is no longer rotated in lockstep — matters only if your mirror Postgres password is being rotated in sync with the source externally.

## Open questions

Three things worth calling out that I didn't want to decide unilaterally:

1. Should `[[mirroring]]` only speak in `[[databases]]`? Should it care about user / authentication, or only that the source / target exists within `[[databases]]`? Or is this for establishing authentication?

2. Should `[[users]]` be server-side authentication only, while `[[databases]]` handles client-side auth only? Today `[[users]]` does both — `name`/`password` gate the client, and the same fields (or `server_*` overrides) get used for `pgdog → Postgres`. A cleaner split might make this feature — and mirroring in general — less bolted-on.

3. Should the ~1350-line inline `mod tests` in `databases.rs` be extracted to a sibling file? The new tests join it in the same style. Thoughts on extracting this to a sibling crate like [diesel's `diesel_tests/`](https://github.com/diesel-rs/diesel/tree/master/diesel_tests)?

### DIFF
--- a/pgdog/src/backend/databases.rs
+++ b/pgdog/src/backend/databases.rs
@@ -180,13 +180,13 @@ pub(crate) fn add(user: ConfigUser) -> Result<AuthResult, Error> {
     let existing = config.users.find(&user);
 
     // User already exists in users.toml.
-    if let Some(mut existing) = existing {
+    let result = if let Some(mut existing) = existing {
         // Password hasn't been set yet.
         if existing.password.is_none() {
             existing.password = user.password.clone();
             add_user(existing)?;
             reload_from_existing()?;
-            Ok(AuthResult::Ok)
+            AuthResult::Ok
         } else if existing
             .password
             .as_deref()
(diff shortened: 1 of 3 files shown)


pgdog/src/backend/databases.rs
```diff
@@ -196,21 +196,73 @@ pub(crate) fn add(user: ConfigUser) -> Result<AuthResult, Error> {
...
+    }
+
+    debug!(
+        r#"propagating passthrough user "{}" to {} mirror destination(s)"#,
+        source.name,
+        writes.len(),
+    );
+
+    {
+        let _lock = lock();
```
levkk: I think you may want to take that lock a little earlier - when you get the config on line 232.

lalvania: Good catch — done. The lock now covers the config read as well, so the snapshot/decision/write is atomic (same discipline as add_user). It's dropped before reload_from_existing(), which re-acquires it.

levkk (approved):
Cool!

I think this is good to go. Quick question re: integration config, did you try to run it locally manually? If so, would you want to add it to our integration suite to protect against regressions?

Issue #623, opened 2025-11-23
[Query engine] Load balancing supports configuring write_function_list and read_only_function_list


cobolbaby:
### Improved Feature Request Description  

**Is your feature request related to a problem? Please describe.**  
I'm frustrated by the lack of fine-grained traffic control when using load balancing with UDFs in my database setup. Specifically, there's no way to differentiate between read-only and write operations for user-defined functions (UDFs) when distributing traffic across database nodes. This limitation makes it challenging to optimize performance and avoid unnecessary conflicts, particularly in scenarios where read and write operations need to be handled by different nodes. Tools like pgpool provide this functionality, and it would be valuable to have similar capabilities.

**Describe the solution you'd like**  
I would like the load balancer to support the configuration of `write_function_list` and `read_only_function_list`. These configurations should allow specifying which UDFs are considered write-intensive or read-only, enabling the load balancer to intelligently route traffic. This feature would help achieve more effective traffic splitting and improve overall system performance and reliability.

**Describe alternatives you've considered**  
An alternative could involve implementing custom middleware to handle traffic splitting at the application level. However, this approach introduces additional complexity and maintenance overhead. Another option might be to modify the application code to explicitly specify UDF behavior, but this reduces flexibility and increases the risk of errors.  

**Additional context**  
This feature is inspired by the functionality provided by tools like pgpool, which allow traffic splitting based on UDF operation types. 

https://github.com/cobolbaby/dockerize-and-ansible/blob/master/pgpool/build/config/confd/templates/pgpool.tmpl#L230-L257

PR #1236, opened 2026-07-19
feat: add configurable write_functions

murex971:
## Description
- Implemented configurable write_functions routing in the new parser path.
- For configured entries, matching ollows PostgreSQL identifier behavior (unquoted -> lowercase, quoted -> case-sensitive).
- Kept existing hardcoded write functions.

## Testing
 Added unit and integration coverage for routing, schema qualification, and PG identifier semantics.
 
## Related Issues
#623


murex971:
@levkk should new features like these be compatible with legacy parser path too ?
currently added to new parser path only.

pgdog-config/src/sharding.rs
```diff
@@ -565,6 +565,17 @@ pub struct QueryParser {
...
 
+/// Per-database write function classification.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Hash, Default, JsonSchema)]
+#[serde(rename_all = "snake_case", deny_unknown_fields)]
+pub struct WriteFunctions {
+    /// Database name.
+    pub database: String,
+    /// Functions that should be treated as write-only.
+    #[serde(default)]
+    pub functions: Vec<String>,
```
levkk: Functions need to be fully qualifiable, e.g.: `partman.create_partition` is a write function, while `create_partition(array)` may not be.

levkk:
> @levkk should new features like these be compatible with legacy parser path too ? currently added to new parser path only.

That's fine, we'll be removing the legacy parser shortly (tm).

example.pgdog.toml
```diff
@@ -276,6 +276,19 @@ cross_shard_disabled = false
...
+# Per-database list of functions that should always route to the primary.
+#
+# Function matching follows PostgreSQL identifier semantics:
+# - unquoted identifiers are folded to lowercase
+# - quoted identifiers preserve case
+#
+# [[write_functions]]
+# database = "pgdog"
+# schema = "partman" # optional
+# functions = ["my_write_fn", "\"Set_Customer_Balance\""]
```
murex971: @levkk I noticed a behavior gap around identifier normalization from pgdog.toml vs PostgreSQL semantics.

Right now, for write_functions, we normalize config identifiers with PG-like rules (unquoted -> lowercase, quoted -> case-preserving), but this is not consistently applied across other config-driven identifier matching paths in PgDog.

pgdog-config/src/sharding.rs
murex971: added optional schema, not sure if you were referring to this ?

pgdog-config/src/sharding.rs
```diff
@@ -565,6 +565,19 @@ pub struct QueryParser {
...
 }
 
+/// Per-database write function classification.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Hash, Default, JsonSchema)]
+#[serde(rename_all = "snake_case", deny_unknown_fields)]
+pub struct WriteFunctions {
+    /// Database name.
+    pub database: String,
+    /// Optional schema qualifier for all functions in this entry.
+    pub schema: Option<String>,
```
levkk: You want to be able to qualify each function separately, e.g.:

```
partman.create_partition
another_schema.write_row
```

pgdog-config/src/sharding.rs
```diff
@@ -565,6 +565,18 @@ pub struct QueryParser {
...
+/// Per-database write function classification.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Hash, Default, JsonSchema)]
+#[serde(rename_all = "snake_case", deny_unknown_fields)]
+pub struct WriteFunctions {
+    /// Database name.
+    pub database: String,
+    /// Functions that should be treated as write-only.
+    /// Each entry can be schema-qualified.
+    #[serde(default)]
+    pub functions: Vec<String>,
```
levkk: You want something like this:

```toml
[[write_functions]]
database = "prod"
name = "create_partition"
schema = "partman"
```

This allows you to disambiguiate between the same function created in different schemas. It's not a common edge case, but it can happen, and we want to be able to scope all Postgres primitives we scan for using a fully qualified identifier. I know that functions can be overloaded, so we should consider capturing parameter data types as well, but for read/write separation it seems unlikely that one version of the function reads and the other one writes, so maybe that's not necessary for this.

murex971: addressed.

PR #1239, opened 2026-07-20
Compute SHA1 shard key from raw digest bytes

crodas:
The SHA1 sharding hash formatted the full 20-byte digest as a hex string, then parsed the last 8 hex characters back into an integer on every call. That allocated a string and ran a radix parse to recover bytes already sitting in the digest.

Read the last four bytes directly and assemble the key with u32::from_be_bytes. The last 8 hex characters map to bytes 16 through 19, so the shard assignment is unchanged; only the work to extract it goes away. This also drops the unwrap on the parse.


crodas:
This is already merged as part of #1240

PR #1240, opened 2026-07-20
Avoid heap allocations when encoding integer and float parameters

crodas:
The sharding hasher and the query-rewrite paths formatted integers and floats with to_string() only to read the resulting bytes and immediately drop the String. On the per-query and per-message paths this is a wasted heap allocation on every call, since Parameter::new and BytesMut copy the bytes into an owned buffer right after.

Format into stack itoa/ryu buffers instead. The bytes fed to the hasher and the wire encoder are unchanged for integers (decimal, no padding), so behavior is identical.

While in the sharding hasher, compute the SHA1 shard key directly from the digest bytes instead of formatting the digest to a hex string, slicing the last eight characters, and parsing them back with from_str_radix. The result is the same value, without the allocation or the unwrap.


crodas:
Related to #1239

levkk (approved):
Nice!

crodas:
I just force-pushed; I forgot to sign my commit.

PR #1241, opened 2026-07-20
fix(config): idempotent admin password generation

meskill:
Prevent regenerating new random password every time on config load. The generated password stays the same after initial load.

The easy way to make the config drift comparison without normalization for the password if it's random.


levkk (approved):
Thank you! This has been driving me nuts. We should also remove the hacky diff logic in our EE code that "handled" this.

meskill:
> Thank you! This has been driving me nuts. We should also remove the hacky diff logic in our EE code that "handled" this.

That's the big change I was talking, I just need this change on open-source to pass the test suites 😄

PR #1243, opened 2026-07-20
Make pgdog-config doc comments pass rustdoc lints

crodas:
cargo doc with -D warnings rejects the crate because doc comments carry bare URLs, which rustdoc::bare_urls flags as non-clickable, and one comment contains an unescaped Option<Self> that rustdoc reads as an unclosed HTML tag. That breaks any workflow gated on a clean doc build.

Wrap every documentation URL in angle brackets so rustdoc renders it as a link, and backtick the Option<Self> reference. The doc text still reaches schemars as the JSON Schema description, now with the URL in the angle-bracket form. Update the CONTRIBUTING.md examples to the same form so new fields follow the convention that builds cleanly.


levkk (approved):
Thanks! To fix CI:

```
cargo run --bin pgdog-jsonschema
```

and commit changes.

levkk (approved):

PR #1244, opened 2026-07-20
Drop client queries abandoned during maintenance mode

### LINKED ISSUE #1242: [Query engine] maintenance mode should not allow queries from disconnected clients
**PgDog version**
Latest main

**Description**
Maintenance  mode checks runs too late, once we already loaded the query into memory/query engine. This will execute the query anyway, even if client disconnects, once maintenance mode is disabled.

crodas:
Attempt to fix #1242

Maintenance mode parked a fully-buffered request on a bare await that ignored the client socket. If the client gave up and disconnected while the database was paused, its query was still dispatched to the backend once maintenance lifted, running work for a client that was already gone.

Wait for maintenance to lift while watching the client socket, so a disconnect aborts the request before it reaches the query engine. Add a cancel-safe MessageBuffer::read_more that detects the disconnect without consuming any pipelined data, and cover the behavior with an integration test that asserts a write from a disconnected client never lands.

### DIFF
--- a/integration/rust/tests/integration/maintenance_mode.rs
+++ b/integration/rust/tests/integration/maintenance_mode.rs
@@ -83,9 +83,13 @@ async fn test_maintenance_mode_drops_disconnected_client_query() {
     });
 
     // Fire the write. It blocks inside PgDog, parked on maintenance mode.
+    // Use simple_query so the INSERT travels as a single Query message and is
+    // buffered as one complete request. execute() would use the extended
+    // protocol and park on the statement-prepare round-trip instead, leaving
+    // no INSERT buffered to dispatch, so it can't tell the fix from the bug.
     let query = tokio::spawn(async move {
         let _ = client
-            .execute("INSERT INTO maintenance_probe VALUES (1)", &[])
+            .simple_query("INSERT INTO maintenance_probe VALUES (1)")
             .await;
         drop(client);
     });
--- a/integration/rust/tests/integration/maintenance_mode.rs
+++ b/integration/rust/tests/integration/maintenance_mode.rs
@@ -2,8 +2,9 @@ use std::time::Duration;
 
 use crate::setup::{admin_sqlx, admin_tokio, connection_failover, connections_sqlx};
 use serial_test::serial;
-use sqlx::Executor;
+use sqlx::{Executor, Row};
 use tokio::time::sleep;
+use tokio_postgres::NoTls;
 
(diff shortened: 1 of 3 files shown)


levkk:
I _think_ a simpler solution could just be to put the maintenance check at the top of `async fn buffer`, wdyt?

crodas:
I think 21764a1f2d2cf4d08362cf8cd1cc9a960000c820 is simpler

integration/rust/tests/integration/maintenance_mode.rs
```diff
@@ -50,6 +51,69 @@ async fn test_maintenance_mode_client_queries() {
     conn.close().await;
 }
 
+#[tokio::test]
+#[serial]
+async fn test_maintenance_mode_drops_disconnected_client_query() {
```
meskill, line 56: this test succeeds on main without fix. 

I think it's because the client.execute will execute parse/describe/sync first and this blocked by maintenance before next execution round-trip anyway. So, try simple_query I guess

crodas: You were right @meskill. Would it be useful if I shared a couple of scripts that I wrote to run the tests locally? If it would be, I would open another PR.

```
 integration/rust/Dockerfile                            | 51 +++++++++++++++++++++++++++++++++++++++++++++++++++
 integration/rust/docker-run.sh                         | 30 ++++++++++++++++++++++++++++++
````

File: pgdog/src/frontend/client/mod.rs [hunk 2]
```diff
@@ -656,6 +646,38 @@ impl Client {
...
+            && let Some(waiter) = maintenance_mode::waiter(&self.database)
+        {
+            // `Pin<Box<_>>` is `Unpin`, so `&mut waiter` polls the same
+            // future across iterations.
+            let mut waiter = waiter.into_future();
+            loop {
+                select! {
+                    _ = &mut waiter => break,
+
+                    // A client waiting for its response shouldn't be sending
```
levkk, line 666: Not necessarily, the client can be pipelined, i.e., send multiple queries in a row. That's allowed by the protocol and is commonly done by clients, e.g., `pgx.Batch` (go/pgx).

integration/rust/tests/integration/maintenance_mode.rs
meskill: if we can make it as integration test that would be great as the part of this pr

PR #1248, opened 2026-07-21
feat: add process name to 2pc transaction ID

levkk:
### Globally unique transaction ID

Without hopefully going over the 200 bytes limit.

Example:

```
__pgdog_2pc_pgdog-node-1_1231424234
```

The last part (the number) can be up to 20 chars long (`usize::MAX`), so I'm optimistic people don't name their nodes something 150 characters long, but I may be proven wrong.

### Misc

- Refactor cluster readiness into its own module.
- Maybe? add code coverage to CI for integration tests? Still not sure if its working.
- Add explicit `SIGTERM` handler which calls `std::process:exit`.


pgdog/src/frontend/client/query_engine/two_pc/manager.rs
```diff
@@ -379,12 +382,71 @@ impl Manager {
...
 
         Ok(())
     }
 
+    /// Drop abandoned two-phase commit transactions.
+    ///
+    /// WARNING: This only happens if durability for 2pc is off. Running this
+    /// will cause data loss.
+    ///
+    pub async fn cleanup_abandoned(&self) -> Result<(), Error> {
```
sgrif:
```suggestion
    pub(crate) async fn cleanup_abandoned(&self) -> Result<(), Error> {
```

pgdog/src/frontend/client/query_engine/two_pc/statement.rs
```diff
@@ -1,15 +1,17 @@
 //! Per-shard two-phase commit transaction names and control statements.
 
+use crate::frontend::client::query_engine::two_pc::TwoPcTransaction;
+
 use super::TwoPcPhase;
 
 /// Prepared transaction name for a coordinator transaction on one shard.
-pub fn shard_name(transaction: &str, shard: usize) -> String {
+pub fn shard_name(transaction: TwoPcTransaction, shard: usize) -> String {
```
sgrif:
```suggestion
pub(crate) fn shard_name(transaction: TwoPcTransaction, shard: usize) -> String {
```

pgdog/src/frontend/client/query_engine/two_pc/statement.rs
```diff
@@ -1,15 +1,17 @@
...
     format!("{transaction}_{shard}")
 }
 
 /// Build `PREPARE TRANSACTION`, `COMMIT PREPARED`, or `ROLLBACK PREPARED`
 /// for a shard participant.
-pub fn phase_control(transaction: &str, shard: usize, phase: TwoPcPhase) -> String {
+pub fn phase_control(transaction: TwoPcTransaction, shard: usize, phase: TwoPcPhase) -> String {
```
sgrif:
```suggestion
pub(crate) fn phase_control(transaction: TwoPcTransaction, shard: usize, phase: TwoPcPhase) -> String {
```

pgdog/src/util.rs
```diff
@@ -138,11 +138,27 @@ pub fn instance_id() -> &'static str {
...
 }
 
+static DEPLOYMENT_ID: Lazy<Option<String>> = Lazy::new(|| env::var("DEPLOYMENT_ID").ok());
+
+/// Get the ID of this PgDog deployment.
+///
+/// This should be _globally_ unique
+/// and is used to differentiate 2pc transactions.
+///
+pub fn deployment_id() -> Option<&'static str> {
```
sgrif:
```suggestion
pub(crate) fn deployment_id() -> Option<&'static str> {
```

pgdog/src/backend/pool/connection/binding.rs
```diff
@@ -394,9 +397,15 @@ impl Binding {
     }
 
     /// Execute two-phase commit transaction control statements.
-    pub async fn two_pc(&mut self, name: &str, phase: TwoPcPhase) -> Result<(), Error> {
+    pub async fn two_pc(
```
sgrif:
```suggestion
    pub(crate) async fn two_pc(
```

pgdog/src/frontend/client/query_engine/two_pc/transaction.rs
```diff
@@ -13,19 +15,38 @@ impl TwoPcTransaction {
...
     }
+
+    /// This transaction was created by this process.
+    pub(crate) fn is_mine(&self) -> bool {
+        self.to_string().starts_with(&Self::global_prefix())
+    }
+
+    /// A prefix to identify two-phase commit transactions generated
+    /// by this PgDog process.
+    pub(crate) fn global_prefix() -> String {
```
sgrif:
```suggestion
    fn global_prefix() -> String {
```

pgdog/src/frontend/client/query_engine/two_pc/transaction.rs
```diff
@@ -13,19 +15,38 @@ impl TwoPcTransaction {
         // so multiple instances of PgDog don't create an identical transaction.
         Self(rng().random_range(0..usize::MAX))
     }
+
+    /// This transaction was created by this process.
+    pub(crate) fn is_mine(&self) -> bool {
```
sgrif: This function name confused me, I think we can make it clearer.

```suggestion
    pub(crate) fn created_by_this_deployment(&self) -> bool {
```

pgdog/src/frontend/client/query_engine/two_pc/transaction.rs
```diff
@@ -46,4 +69,24 @@ mod test {
         let reverse = TwoPcTransaction::from_str(transaction.to_string().as_str()).unwrap();
         assert_eq!(reverse.0, transaction.0);
     }
+
+    #[test]
+    fn test_instance_id() {
+        for id in [1024, 11111111, usize::MAX, usize::MIN] {
+            let transaction = TwoPcTransaction(id);
+            let instance_id = instance_id(); // Generate it, it's a singleton.
```
sgrif: It feels really gross having `TwoPcTransaction` rely on global state for its `Display` impl. Can we separate that out into something that allows the state to be injected?

pgdog/src/util.rs
```diff
@@ -138,11 +138,27 @@ pub fn instance_id() -> &'static str {
...
+/// This assumes the NODE ID follows the following format:
+///
+/// <something we don't care about>-<number between 0 and 1023 inclusively>
+///
 pub fn node_id() -> Result<u64, ParseIntError> {
     // split always returns at least one element.
     instance_id().split("-").last().unwrap().parse()
 }
 
+static DEPLOYMENT_ID: Lazy<Option<String>> = Lazy::new(|| env::var("DEPLOYMENT_ID").ok());
```
sgrif, line 151: `DEPLOYMENT_ID` is a really generic name. Should we prefix this?

sgrif (changes requested):

pgdog/src/util.rs
levkk: Like `PGDOG_DEPLOYMENT_ID` or something?

pgdog/src/frontend/client/query_engine/two_pc/transaction.rs
levkk: It does, yeah. We could call it more explicitly for sure, but we also want to avoid ever rendering a "non-unique" ID by accident.

pgdog/src/frontend/client/query_engine/two_pc/manager.rs
levkk: This one I'm calling from main.rs actually so needs to be pub

pgdog/src/util.rs
sgrif: Yes

pgdog/src/frontend/client/query_engine/two_pc/manager.rs
sgrif: Ah. We should fix that so main doesn't have different visibility rules than lib but that's a separate change

pgdog/src/frontend/client/query_engine/two_pc/transaction.rs
sgrif: Right, I'm thinking we remove the `Display` impl and have `fn display_name(&self, deployment_id: Option<&str>, instance_id: &str)`, pushing the state up to the caller and making this easier to test

pgdog-config/src/general.rs
```diff
@@ -622,6 +622,18 @@ pub struct General {
...
+    ///
+    /// WARNING: Data loss will occur, enable this only if you don't care about consistency
+    /// and are not using the 2pc WAL.
+    #[serde(default = "General::two_phase_commit_rollback_abandoned")]
+    pub two_phase_commit_rollback_abandoned: bool,
+
+    /// Maximum amount of time to block startup in order to rollback abandoned
+    /// transactions.
+    #[serde(default = "General::two_phase_commit_rollback_abandoned_timeout")]
+    pub two_phase_commit_rollback_abandoned_timeout: u64,
```
meskill: *raising hands to the sky*: when would we split the configuration into the subsections and not put everything in general

pgdog-config/src/general.rs
```diff
@@ -622,6 +622,18 @@ pub struct General {
     #[serde(default = "General::two_phase_commit_wal_checkpoint_interval")]
     pub two_phase_commit_wal_checkpoint_interval: u64,
 
+    /// Rollback abandoned transactions.
+    ///
+    /// WARNING: Data loss will occur, enable this only if you don't care about consistency
+    /// and are not using the 2pc WAL.
```
meskill: should we ignore this if wal is enabled then?

pgdog/src/frontend/client/query_engine/two_pc/statement.rs
```diff
@@ -1,15 +1,17 @@
 //! Per-shard two-phase commit transaction names and control statements.
 
+use crate::frontend::client::query_engine::two_pc::TwoPcTransaction;
+
 use super::TwoPcPhase;
 
 /// Prepared transaction name for a coordinator transaction on one shard.
-pub fn shard_name(transaction: &str, shard: usize) -> String {
+pub(crate) fn shard_name(transaction: TwoPcTransaction, shard: usize) -> String {
```
meskill: let's make a reverse function for this to parse the TwoPcTransaction and not embed the parsing logic inside `TwoPcTransactions::load`

pgdog/src/frontend/client/query_engine/two_pc/manager.rs
```diff
@@ -379,12 +383,87 @@ impl Manager {
...
+        info!("[2pc] rolling back abandoned transactions");
+
+        let mut cleaned_up = 0;
+        for cluster in databases().all().values() {
+            cleaned_up += self.cleanup_abandoned_for_cluster(cluster).await?;
+        }
```
meskill: try_join_all everywhere otherwise we can hit timeout limit executing it sequentially

pgdog/src/frontend/client/query_engine/two_pc/manager.rs
```diff
@@ -379,12 +383,87 @@ impl Manager {
...
+
+    async fn cleanup_abandoned_for_cluster(&self, cluster: &Cluster) -> Result<usize, Error> {
+        let cleaned_up = try_join_all(
+            cluster
+                .shards()
+                .iter()
+                .enumerate()
+                .map(|(number, shard)| Self::cleanup_abandoned_for_shard(number, shard)),
+        )
+        .await?;
```
meskill: should we propagate all the errors and stop on the first one, or we should try to cleanup as many as possible even if some failed?

meskill (changes requested):

pgdog-config/src/general.rs
levkk: LOL. Yeah....might be time, huh?

pgdog-config/src/general.rs
levkk: The WAL is super experimental at the moment. We need to put a lot more work into this subsystem.

pgdog/src/frontend/client/query_engine/two_pc/manager.rs
levkk: Good question...it's one of those doomed if you do, doomed if you don't situations I think

pgdog/src/frontend/client/query_engine/two_pc/transaction.rs
levkk: That'll be a lot of places, which I guess...doesn't really help, since we can fat-finger the calling site. And those are very much invariants/identifiers  which we never ever want to confuse.

pgdog/src/util.rs
levkk: Maybe. We already use other similar-named env variables there, e.g. `NODE_ID`. I don't expect ppl to run pgdog in an env shared by another app, and if they do, they certainly wouldn't be sharding, and if they did, I would be like...damn. But yeah, gotta be careful with breaking changes, like I'd want to change all env variables to match the new PGDOG_ format, but I cant :(

pgdog/src/frontend/client/query_engine/two_pc/manager.rs
levkk: Refactored this to be done at cluster startup, so fully async and parallel

pgdog/src/frontend/client/query_engine/two_pc/transaction.rs
```diff
@@ -13,19 +15,38 @@ impl TwoPcTransaction {
         // so multiple instances of PgDog don't create an identical transaction.
         Self(rng().random_range(0..usize::MAX))
     }
+
+    /// This transaction was created by this process.
+    pub(crate) fn is_created_by_this_process(&self) -> bool {
+        self.to_string().starts_with(&Self::global_prefix())
+    }
```
meskill: this is always true - we use `to_string` that uses global_prefix to build the string and then we check if we have the global_prefix here. The self it's just a usize, there is no way we can check it actually. And no unit tests for this

pgdog/src/frontend/client/query_engine/two_pc/manager.rs
meskill: good, but as a reminder - try_join_all will cancel all other futures on the first error raised

pgdog/src/backend/pool/cluster_launch.rs
```diff
@@ -0,0 +1,168 @@
...
+        self.launch_readiness_monitor();
+    }
+
+    /// Shutdown the connection pools.
+    pub(crate) fn shutdown(&self) {
+        for shard in self.shards() {
+            shard.shutdown();
+        }
+
+        self.readiness.set_online(false);
```
meskill: shouldn't we disable readiness before calling actual shutdown, to prevent partial usage?

pgdog/src/backend/pool/cluster_launch.rs
```diff
@@ -0,0 +1,168 @@
...
+
+        // Release readiness waiters, the cluster is going away.
+        self.mark_ready();
```
meskill, line 70: marking ready on shutdown looks odd - is this expected?

pgdog/src/frontend/client/query_engine/two_pc/transaction.rs
meskill: oh, ok, there were some wrong function when the cleanup was there

pgdog/src/backend/pool/cluster_launch.rs
levkk: Yeah, we want to kick the waiters if any, so nobody waits forever on startup. They will try to use the cluster and hit the "cluster is offline" error.

pgdog/src/backend/pool/cluster_launch.rs
levkk: Good call!

pgdog/src/frontend/client/query_engine/two_pc/transaction.rs
levkk: Deleted the whole thing

pgdog-config/src/general.rs
levkk: Deleted this feature entirely, was a bad idea.

meskill (approved):

pgdog/src/backend/pool/cluster_launch.rs
sgrif: Maybe give this a clearer name? Tbh `CancellationToken` feels like the wrong type to be using to wait for a server to be ready.

pgdog/src/backend/pool/cluster_launch.rs
```diff
@@ -0,0 +1,167 @@
...
+
+    /// Wait for the schema to load on all shards.
+    async fn wait_schema_loaded(&self) {
+        if !self.load_schema() {
+            return;
+        }
+
+        for shard in self.shards() {
+            shard.wait_schema_loaded().await;
+        }
```
sgrif, line 115: These should be able to run in parallel.

sgrif (approved):

levkk:
Still trying to figure out how to get coverage on integration tests -- will use this PR as guinea pig until I give up / it works.

pgdog/src/backend/pool/cluster_launch.rs
levkk: Ah yeah. Those are just waiters, the actual schema fetching runs in parallel, so these will resolve quickly.

levkk:
Pretty sure its working guys, let me know if you spot anything:

image

levkk:
@sgrif can you tell me if this is hallucinated: https://github.com/pgdogdev/pgdog/blob/ae77c993cf057dbe89cc6056bfd0167794c20b0e/pgdog/src/main.rs#L243

sgrif:
It's accurate, but if this is purely for code coverage purposes I'd be sending SIGINT instead of SITGERM to the server (presumably we already gracefully shutdown on sigint)

levkk:
> It's accurate, but if this is purely for code coverage purposes I'd be sending SIGINT instead of SITGERM to the server

Hmm, I wonder who's sending  SIGTERM...nevermind, the call is coming from inside the house, it's our CI script.

levkk:
> (presumably we already gracefully shutdown on sigint)

Correct

levkk:
Aight, well...maybe I'll just keep the handler? It seems like a nice to have feature, plus we get a logline, so that's cool.

PR #1258, opened 2026-07-23
feat: schema cache to prevent thundering herd

levkk:
Sharded deployments with dozens / hundreds of users will stampede the servers to load the schema. With large schemas, e.g., thousands of tables, and multiple replicas, this is pretty slow and will cause issues.

The cache is using the database name and shard number, which uniquely identifies a host in `pgdog.toml`.

### DIFF
--- a/pgdog/src/backend/pool/cluster.rs
+++ b/pgdog/src/backend/pool/cluster.rs
@@ -8,6 +8,7 @@ use pgdog_config::{
 };
 use std::{sync::Arc, time::Duration};
 
+use crate::backend::schema::SchemaCache;
 use crate::frontend::router::sharding::ShardedTable;
 use crate::{
     backend::{
@@ -126,47 +127,52 @@ impl ClusterShardConfig {
 /// Cluster creation config.
 #[derive(Debug)]
 pub struct ClusterConfig<'a> {
-    pub name: &'a str,
-    pub shards: &'a [ClusterShardConfig],
-    pub lb_strategy: LoadBalancingStrategy,
-    pub user: &'a str,
-    pub passwords: Vec<PasswordKind>,
-    pub pooler_mode: PoolerMode,
-    pub sharded_tables: ShardedTables,
-    pub replication_sharding: Option<String>,
-    pub multi_tenant: &'a Option<MultiTenant>,
-    pub rw_strategy: ReadWriteStrategy,
-    pub rw_split: ReadWriteSplit,
-    pub schema_admin: bool,
-    pub cross_shard_disabled: bool,
-    pub two_pc: bool,
-    pub two_pc_auto: bool,
-    pub sharded_schemas: ShardedSchemas,
-    pub rewrite: &'a Rewrite,
-    pub prepared_statements: &'a PreparedStatements,
-    pub dry_run: bool,
-    pub expanded_explain: bool,
-    pub pub_sub_channel_size: usize,
-    pub query_parser: QueryParserLevel,
-    pub query_parser_engine: QueryParserEngine,
-    pub log_min_duration_parse: Option<Duration>,
-    pub log_query_sample_length: usize,
-    pub connection_recovery: ConnectionRecovery,
-    pub client_connection_recovery: ConnectionRecovery,
-    pub lsn_check_interval: Duration,
-    pub reload_schema_on_ddl: bool,
-    pub load_schema: LoadSchema,
-    pub resharding_parallel_copies: usize,
-    pub resharding_copy_retry_max_attempts: usize,
-    pub resharding_copy_retry_min_delay: u64,
-    pub resharding_replication_retry_max_attempts: usize,
-    pub resharding_replication_retry_min_delay: u64,
-    pub regex_parser_limit: usize,
-    pub pub_sub_enabled: bool,
-    pub identity: &'a Option<String>,
+    name: &'a str,
+    shards: &'a [ClusterShardConfig],
+    lb_strategy: LoadBalancingStrategy,
+    user: &'a str,
+    passwords: Vec<PasswordKind>,
+    pooler_mode: PoolerMode,
+    sharded_tables: ShardedTables,
+    replication_sharding: Option<String>,
+    multi_tenant: &'a Option<MultiTenant>,
+    rw_strategy: ReadWriteStrategy,
+    rw_split: ReadWriteSplit,
+    schema_admin: bool,
+    cross_shard_disabled: bool,
+    two_pc: bool,
+    two_pc_auto: bool,
+    sharded_schemas: ShardedSchemas,
+    rewrite: &'a Rewrite,
+    prepared_statements: &'a PreparedStatements,
+    dry_run: bool,
+    expanded_explain: bool,
+    pub_sub_channel_size: usize,
+    query_parser: QueryParserLevel,
+    query_parser_engine: QueryParserEngine,
+    log_min_duration_parse: Option<Duration>,
+    log_query_sample_length: usize,
+    connection_recovery: ConnectionRecovery,
+    client_connection_recovery: ConnectionRecovery,
+    lsn_check_interval: Duration,
+    reload_schema_on_ddl: bool,
+    load_schema: LoadSchema,
+    resharding_parallel_copies: usize,
+    resharding_copy_retry_max_attempts: usize,
+    resharding_copy_retry_min_delay: u64,
+    resharding_replication_retry_max_attempts: usize,
+    resharding_replication_retry_min_delay: u64,
+    regex_parser_limit: usize,
+    pub_sub_enabled: bool,
+    identity: &'a Option<String>,
+    schema_cache: SchemaCache,
 }
 
 impl<'a> ClusterConfig<'a> {
+    /// Dependencies for creating a Cluster.
+    ///
+    /// TODO(lev): This is getting unruly. We may need a struct for a struct :)
+    ///
     pub(crate) fn new(
         config: &'a crate::config::Config,
         user: &'a User,
@@ -174,6 +180,7 @@ impl<'a> ClusterConfig<'a> {
         sharded_tables: ShardedTables,
         sharded_schemas: ShardedSchemas,
         query_parser: QueryParser,
+        schema_cache: SchemaCache,
     ) -> Self {
         let general = &config.general;
         let multi_tenant = config.multi_tenant();
@@ -228,6 +235,7 @@ impl<'a> ClusterConfig<'a> {
             regex_parser_limit: general.regex_parser_limit,
             pub_sub_enabled: general.pub_sub_enabled(),
             identity: &user.identity,
+            schema_cache,
         }
     }
 }
@@ -274,6 +282,7 @@ impl Cluster {
             regex_parser_limit,
             pub_sub_enabled,
             identity,
+            schema_cache,
         } = config;
 
         let identifier = Arc::new(DatabaseUser {
@@ -296,6 +305,7 @@ impl Cluster {
                         identifier: identifier.clone(),
                         lsn_check_interval,
                         pub_sub_enabled,
+                        schema_cache: schema_cache.clone(),
                     })
                 })
                 .collect(),
@@ -661,6 +671,7 @@ mod test {
         ConfigAndUsers, OmnishardedTable, PoolerMode, QueryParserLevel, ShardedSchema,
     };
 
+    use crate::backend::schema::SchemaCache;
     use crate::frontend::router::sharding::ShardedTable;
     use crate::{
         backend::{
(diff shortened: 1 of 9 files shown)


pgdog/src/backend/schema/cache/mod.rs
```diff
@@ -0,0 +1,63 @@
...
+        // so only one shard/user can fetch the schema at a time.
+        let mut guard = entry.lock().await;
+
+        if guard.is_loaded() {
+            return Ok(guard.clone());
+        }
+
+        let schema = shard.fetch_schema().await?;
+
+        *guard = schema.clone();
```
meskill, line 45: in theory here could be a tricky data race with `SchemaCache::global().clear();`:

on `replace_databases` the following happens: `clear -> new_databases.launch -> old_databases.shutdown`

if for some reason we trigger `replace_databases` while schemas are still loading for old (like with retries due to errors) we may try to write into the cache that is basically associated with the new databases already. 

The shutdown token would've be helpful, but this is called after the clear and launch, leaving this race window.

There is another race that is not important I think: lock -> fetch_schema -> reload_databases -> set to old guard. But it's not a big issue since the guard will not point to the cache map entry anymore and that'd be a separate mutex.

levkk: I think I know what to do here. I'll add a global "config is reloading" signal that can cancel all of these long-running loops. I've been meaning to add something like this; making everything more deterministic.

levkk: Might also have to be versioned too. Each config reload increments the config version so each `Cluster` knows which "epoch" it belongs to.

pgdog/src/backend/schema/cache/mod.rs
```diff
@@ -0,0 +1,63 @@
...
+
+type Entry = Arc<Mutex<Schema>>;
+
+static CACHE: Lazy<SchemaCache> = Lazy::new(SchemaCache::default);
+
+/// Schema cache.
+#[derive(Debug, Default)]
+pub(crate) struct SchemaCache {
+    // Database => shard => Schema
+    cache: DashMap<String, DashMap<usize, Entry>>,
```
sgrif: Let's newtype these so we don't need a comment to know what they are

pgdog/src/backend/schema/cache/mod.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+impl SchemaCache {
+    /// Get a schema entry from the cache or load it from
+    /// the server and store it in the cache.
+    ///
+    /// The loading is synchronized with a mutex, so only one user
+    /// can load a schema at a time, preventing a thundering herd situtation.
+    ///
```
sgrif:
```suggestion
```

pgdog/src/backend/schema/cache/mod.rs
```diff
@@ -0,0 +1,63 @@
...
+    pub(crate) async fn get(&self, shard: &Shard) -> Result<Schema, super::Error> {
+        // This is synchronized.
```
sgrif, line 26: The compiler would have prevented it if it wasn't.


```suggestion
```

pgdog/src/backend/schema/cache/mod.rs
```diff
@@ -0,0 +1,63 @@
...
+//! it for each [`crate::backend::pool::Cluster`].
+
+use dashmap::DashMap;
+use once_cell::sync::Lazy;
+use std::sync::Arc;
+use tokio::sync::Mutex;
+
+use crate::backend::{Schema, Shard};
+
+type Entry = Arc<Mutex<Schema>>;
```
sgrif, line 10: This should definitely at least be a `RwLock` not a `Mutex`, but given that the only mutation we're doing is setting the value, `OnceCell` seems much more appropriate.

pgdog/src/backend/schema/cache/mod.rs
sgrif, line 45: If we use `OnceCell` for this

```suggestion
        entry.get_or_try_init(|| shard.fetch_schema())
```

pgdog/src/backend/schema/cache/mod.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    /// Remove all entries from the schema cache.
+    pub(crate) fn clear(&self) {
+        self.cache.clear();
+    }
+
+    /// Get a reference the the global schema cache.
+    pub(crate) fn global() -> &'static SchemaCache {
+        &CACHE
+    }
```
sgrif: This feels pointless over just making the static be public

sgrif (dismissed):

pgdog/src/backend/schema/cache/mod.rs
sgrif: *race condition, not a data race.

pgdog/src/backend/schema/cache/mod.rs
levkk: We are preventing a thundering herd, so the contention is intentional. We are also calling an async primitive to load the schema, so we want an async mutex to prevent others from doing so.

pgdog/src/backend/schema/cache/mod.rs
levkk: That feels obscure to me. What I have here is easier to read, at least for me and to understand its purpose.

pgdog/src/backend/schema/cache/mod.rs
sgrif: We don't want to have contention on reads after loading. `OnceCell` has an async primitive, but it shouldn't actually matter if we use the tokio or std version for this

pgdog/src/backend/schema/cache/mod.rs
levkk: I don't like global statics. They look ugly and make me thing I'm writing C.

pgdog/src/backend/schema/cache/mod.rs
levkk: Either it will have automatic conversion from string, which will do nothing to prevent type issues, or it won't and we have to refactor `String` to `Database` in like 100 files. Pick your poison. Comment is fine for now.

pgdog/src/backend/schema/cache/mod.rs
levkk: Ok :D

pgdog/src/backend/schema/cache/mod.rs
sgrif: And to be clear, `OnceCell` will prevent the thundering herd issue

levkk: There is no contention on reads. This will be loaded _once_ by maybe 100 `Cluster`s. This is strictly to prevent a thundering herd once per config reload.

levkk: We are introducing a contention on writes, by design.

pgdog/src/backend/schema/cache/mod.rs
sgrif: Using the purpose built thing is more obscure than implementing it yourself in a worse way? I do not understand why you're being so wildly dismissive here.

levkk: Yeah, this one is kinda tough. We may need to refactor the config system from an ArcSwap to a watch, or something. to ensure this works across the whole app.

Right now, the race condition is real, but it may not have a real impact. i.e., if they race, they happen around the same time, so the schema is probably identical anyway. This was built specifically to busted when the user wants it to, e.g., they create a new table, then hit `RELOAD` in admin. Either way, the table already exists, and the schema loaded in the cache is fresh.

pgdog/src/backend/schema/cache/mod.rs
sgrif: `Mutex` contends on both reads and writes. You cannot have multiple readers from a `Mutex`

pgdog/src/backend/schema/cache/mod.rs
levkk: I don't mind leaving comments like this around. By reading it, it wasn't obvious to me, since we are reaching inside two dash maps, which are Sync, but still. It's something a more experienced Rust engineer will find redundant, but I'm not there yet and I like to leave breadcrumbs for past me and others, explaining my thinking for why something is ok (or may not be).

pgdog/src/backend/schema/cache/mod.rs
levkk: Based on my understanding of what you're saying, this is not currently an issue and what you're proposing is a change in style, not substance.

sgrif: You're sticking this in a `OnceCell` at the end anyway. You can remove the extra step by just using that primitive here

pgdog/src/backend/schema/cache/mod.rs
sgrif: I think a comment saying *why* this is synchronized is useful. Simply saying that the language guarantees exist doesn't seem useful

pgdog/src/backend/schema/cache/mod.rs
levkk: I figured it out. We should attach all config-scoped "globals" to the `Cluster`. Duh. It's already versioned in a way that each config version gets its own `Cluster`, so as long as the cache is scoped to the lifetime of that `Cluster`, we don't have a race between config versions.

We should ban globals lol. Everything from now on should be `Cluster`-scoped.

pgdog/src/backend/pool/cluster.rs
```diff
@@ -126,54 +127,60 @@ impl ClusterShardConfig {
...
-    pub load_schema: LoadSchema,
-    pub resharding_parallel_copies: usize,
-    pub resharding_copy_retry_max_attempts: usize,
-    pub resharding_copy_retry_min_delay: u64,
-    pub resharding_replication_retry_max_attempts: usize,
-    pub resharding_replication_retry_min_delay: u64,
-    pub regex_parser_limit: usize,
-    pub pub_sub_enabled: bool,
-    pub identity: &'a Option<String>,
+    name: &'a str,
```
levkk, line 130: @meskill Ignore this diff, just removed `pub`.

pgdog/src/backend/schema/cache/mod.rs
levkk: Added in [2fe3347](https://github.com/pgdogdev/pgdog/pull/1258/commits/2fe3347b3d21cf84ce834c7c7338429612bbf794)

pgdog/src/backend/schema/cache/mod.rs
meskill: newtype is still useful even with auto conversion - it has clear meaning on type levels and allows for specific functionality. And the migration could be also done partially by converting to/from String on the boundaries.

not saying do it here, but we should prefer this for new code maybe

pgdog/src/backend/schema/cache/mod.rs
meskill: hey, I had to agree with Sage - the OnceCell is just built for exactly this but does it with simpler code and uses just AtomicBool instead of mutex for synchronization.

doc just for reference - https://docs.rs/tokio/latest/tokio/sync/struct.OnceCell.html

meskill: there is a subtle difference though - `guard.is_loaded` used to check the initialization instead of Option. I guess the empty schema will be refetched every time because of this 🤷🏻‍♂️

levkk: Yeah, schema is not a well-designed struct, so I was trying to follow the original pattern just for the cache.

> AtomicBool instead of mutex

Mutex uses an Atomic underneath for the fast path. Fast path = atomic, slow path = futex. If there is no contention, there is no difference as far as I understand it. This feels like a coding style choice / nit, so I'm not sure why Sage decided this to be blocking and not just a nit.

pgdog/src/backend/pool/cluster.rs
```diff
@@ -126,54 +127,60 @@ impl ClusterShardConfig {
...
+    regex_parser_limit: usize,
+    pub_sub_enabled: bool,
+    identity: &'a Option<String>,
+    schema_cache: SchemaCache,
 }
 
 impl<'a> ClusterConfig<'a> {
+    /// Dependencies for creating a Cluster.
+    ///
+    /// TODO(lev): This is getting unruly. We may need a struct for a struct :)
```
meskill, line 174: maybe you'd like bon? https://bon-rs.com/guide/overview#method-builder

levkk: Yeah. `ClusterConfig` was meant to be a "manual bon" for `Cluster::new`. So now we need a `ClusterConfigConfig` :)

PR #1259, opened 2026-07-24
fix: don't buffer client request before checking maintenance mode

### LINKED ISSUE #1242: [Query engine] maintenance mode should not allow queries from disconnected clients

levkk:
- fixes #1242 
- add some comments
- refactor some methods to be `pub(crate)`

### DIFF
--- a/pgdog/src/frontend/client/mod.rs
+++ b/pgdog/src/frontend/client/mod.rs
@@ -531,43 +531,46 @@ impl Client {
 
     /// Handle client messages.
     async fn client_messages(&mut self, query_engine: &mut QueryEngine) -> Result<(), Error> {
-        // Check maintenance mode.
-        if !self.in_transaction()
-            && !self.admin
-            && let Some(waiter) = maintenance_mode::waiter(&self.database)
-        {
-            let state = query_engine.get_state();
-            query_engine.set_state(State::Waiting);
-            waiter.await;
-            query_engine.set_state(state);
-        }
-
         // If client sent multiple requests, split them up and execute individually.
         let spliced = self.client_request.spliced()?;
         if spliced.is_empty() {
+            // Execute the original request as-is.
             let mut context = QueryEngineContext::new(self);
             query_engine.handle(&mut context).await?;
             self.transaction = context.transaction();
         } else {
+            // Execute the spliced requests one at a time.
             let total = spliced.len();
             let mut reqs = spliced.into_iter().enumerate();
+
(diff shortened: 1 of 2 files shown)


pgdog/src/frontend/client/mod.rs
```diff
@@ -582,13 +585,29 @@ impl Client {
...
-    /// sent a complete request.
+    /// sends a complete request.
+    ///
     async fn buffer(&mut self, state: State) -> Result<BufferEvent, Error> {
         self.client_request.clear();
 
+        // Check maintenance mode.
+        //
+        // This will pause the client until the maintenance mode is off
+        // without checking any timeouts or reading data off of the socket
```
meskill, line 599: the buffer future is created once during `select!` in loop and in order to it to be recreated we should proceed to the next iteration of the loop.

that means we can get the following flow:
1. client connects and stays idle
2. no maintenance mode so we don't wait and go straight to buffer read
3. we enable maintenance
4. the idle client runs some command and we execute them without checks for maintenance 

yes, other branches in select! could recreate the future, but that's not reliable. 

So, moving the check upfront could it make it worse, because the check actually moved to the connection time whilst previously it was at execution time, when we actually got some commands.

I think we really need integration tests here to verify the behavior.

pgdog/src/frontend/client/mod.rs
```diff
@@ -582,13 +585,29 @@ impl Client {
...
+        // so if the client disconnects while we wait, we won't send
+        // its pending query to the database.
```
meskill, line 601: I guess the socket data will still be in the kernel buffer and we'll still read and try to execute as soon as we read proper spliced part. The query will be executed then and we'll probably get an error when we write.

that's why another pr for the issue was trying to always read the stream to catch the EOF to understand if the client is disconnected during maintenance.

meskill (changes requested):

pgdog/src/frontend/client/mod.rs
levkk: Oh wow. That's such a footgun...really good catch.

pgdog/src/frontend/client/mod.rs
levkk: I see that now.

PR #1265, opened 2026-07-24
fix: don't reset/set parameters unnecessarily

### LINKED ISSUE #1262: [Query engine] Parameter syncrhonization resets the world for one parameter change
**PgDog version**
e.g., `v0.1.50`

**Description**
if only one parameter differs between client and server, we send `RESET` queries for all parameters we know about. Instead, we should skip the ones we know to be identical

levkk:
We were calling `RESET` on parameters we were about to set with `SET`. This caused unnecessary traffic on the DB and also was super confusing in APM, because it was done for no obvious reason. 

fix #1262 

cc @frodsan would appreciate a review if you got a second.

### DIFF
--- a/pgdog/src/net/parameter.rs
+++ b/pgdog/src/net/parameter.rs
@@ -383,7 +383,7 @@ impl Parameters {
             // Ignore untracked parameters.
             .filter(|(k, _)| !UNTRACKED_PARAMS.contains(k))
             // Ignore parameters that have identical values, they don't need to be updated.
-            .filter(|(k, v)| other.get(k).map(|other| other == *v).unwrap_or(true))
+            .filter(|(k, v)| other.get(k).map(|other| other != *v).unwrap_or(true))
             .map(|(k, v)| (k.clone(), v.clone()))
             .collect::<BTreeMap<_, _>>();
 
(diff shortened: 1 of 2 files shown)


frodsan (approved):
Looks good! Thanks ❤️

pgdog/src/net/parameter.rs
```diff
@@ -406,9 +437,18 @@ impl Parameters {
...
-    pub fn reset_queries(&self) -> Vec<Query> {
+    /// Create a list of `RESET` queries that will reset parameters
+    /// back to their default value.
+    ///
+    /// This will ignore all parameters that are about to be SET
+    /// by incoming parameters. It will only reset parameters
+    /// that are currently set on the server and which do not
+    /// have a value on the incoming client.
+    ///
+    pub(crate) fn reset_queries(&self, other: &Self) -> Vec<Query> {
```
meskill, line 448: let's make sure we have a unit test for the updated behavior

pgdog/src/net/parameter.rs
```diff
@@ -406,9 +437,18 @@ impl Parameters {
...
         self.params
             .keys()
+            .filter(|name| !other.contains_key(*name))
             .map(|name| Query::new(format!(r#"RESET "{}""#, name)))
```
meskill, line 452: well, there could be some inconsistency because how the `UNTRACKED_PARAMS` is handled, and previously some parameters will be resetted anyway.

I don't think this is the issue with this code in particular, but the way `UNTRACKED_PARAMS` is handled - we use it in many functions, but not all of them. Maybe we could simplify it in the future to avoid the issues and simplify managing and handling like hash

pgdog/src/net/parameter.rs
levkk: Pretty sure I added one...maybe the clanker lied.

pgdog/src/net/parameter.rs
levkk: Yes. This struct is incredibly overloaded and difficult to understand. It's serving many different / adjacent concerns, and the caller has to be super careful to call methods in the right order, which is not great (and where this bug came from as well).

We need to probably refactor this into `ClientParameters` and `ServerParameters` and make those use `Parameters` underneath for shared functionality and keep the client/server methods separate.

pgdog/src/net/parameter.rs
levkk: Ah you're totally right. Let me add a test as a follow up.

levkk: Added in #1267

PR #1266, opened 2026-07-26
feat(config): add query_cache_memory_limit and query_cache_idle_timeout

### LINKED ISSUE #1261: Query cache (`query_cache_limit`) is bounded by entry count, not memory, and has no TTL — heavy queries balloon RSS until restart
### Summary

PgDog's parser AST cache (`query_cache_limit`, default `1000`) is a single per-process LRU (`static CACHE` in `frontend::router::parser::cache`), shared across all clients and databases, keyed by query text. Two properties combine to let it use far more memory than the entry count suggests:

1. **The limit caps the number of entries, not their total memory.** A cached parse tree for a wide/complex query is orders of magnitude larger than one for `SELECT 1` (MBs vs bytes). With a count limit, the cache's memory footprint is effectively unbounded — `1000 × (arbitrary entry size)`.
2. **Entries never expire by time (no TTL / time-to-idle).** An entry only leaves the cache when 1000 *newer distinct* queries evict it via LRU, on `RESET QUERY_CACHE`, or on restart. A burst of heavy distinct queries can therefore hold a large cache for a long time on a low-diversity workload.

### Reproduction

Self-contained docker-compose repro: https://gist.github.com/IgorOhrimenko/2b886c7d506ff975b8a9ac22acd63c5b

Three phases against a stock pgdog (`query_cache_limit=1000`, `_RJEM_MALLOC_CONF=background_thread:true`):

| phase | RSS | cache_size |
|---|---|---|
| baseline | ~30 MB | 0 |
| 2000 distinct **heavy** queries | ~3 GB | 1000 |
| 120 s light traffic, 100 fixed (hits, no turnover) | stays ~2.5 GB | 1000 |
| 2000 **new** distinct light queries (evict heavy) | → ~150 MB | 1000 |

`cache_size` stays pinned at 1000 throughout — the LRU count is bounded; the memory is not, because entry size varies by orders of magnitude. Memory is released only once the heavy entries are evicted (or via `RESET QUERY_CACHE`), never on a timer.

### Root cause

`frontend::router::parser::cache::Cache` — an `LruCache` sized by `query_cache_limit` (count only). No per-entry byte accounting and no idle expiry.

### Proposed fix

1. **`query_cache_memory_limit`** (bytes): evict LRU until the summed entry size is under a byte budget.
2. **`query_cache_ttl`** (seconds, time-to-idle): drop entries not accessed within the window, via the existing maintenance sweep.

Implemented in #1266.

IgorOhrimenko:
Implements the memory-based cache limit and idle expiry proposed in #1261.

Adds two `[general]` options (also `PGDOG_QUERY_CACHE_MEMORY_LIMIT` / `PGDOG_QUERY_CACHE_IDLE_TIMEOUT` env), both default `0` (off):

- **`query_cache_memory_limit`** (bytes): evict LRU until the summed entry size is under the budget. `query_cache_limit` bounds the *number* of entries but not their memory — a parsed tree for a wide/complex query is orders of magnitude larger than for `SELECT 1`, so a count-based cap alone can't bound RAM. Entry size is measured via jemalloc's per-thread allocation counters (clamped to 0, since cross-thread frees under a work-stealing runtime can make the delta negative); falls back to the query text length on non-jemalloc builds.
- **`query_cache_idle_timeout`** (milliseconds): entries not accessed within the window are dropped by the existing 1s maintenance sweep, releasing an idle working set without waiting for LRU eviction.

Both are opt-in (`0` = off); default behaviour is unchanged. Unit tests cover count/byte eviction, byte accounting and the idle-expiry sweep; config parsing is covered in `pgdog-config`. JSON schema regenerated.

Closes #1261.

### DIFF
--- a/.schema/pgdog.schema.json
+++ b/.schema/pgdog.schema.json
@@ -88,6 +88,8 @@
         "prepared_statements_ttl_jitter": 30000,
         "pub_sub_channel_size": 0,
         "query_cache_limit": 1000,
+        "query_cache_memory_limit": 0,
+        "query_cache_ttl": 0,
         "query_log": null,
         "query_log_stdout": false,
         "query_parser": "auto",
@@ -1066,6 +1068,20 @@
           "type": "integer",
           "format": "uint",
           "default": 1000,
+          "minimum": 0
+        },
+        "query_cache_memory_limit": {
+          "description": "Approximate memory budget (bytes) for the query (AST) cache; entries are evicted (LRU) once the sum of their sizes exceeds it. Bounds RAM regardless of query complexity, unlike the count-based `query_cache_limit`. `0` disables the memory cap.\n\n**Note:** Entry sizes are measured with jemalloc; on builds without it the budget is approximated from query text length.\n\n_Default:_ `0`\n\n<https://docs.pgdog.dev/configuration/pgdog.toml/general/#query_cache_memory_limit>",
+          "type": "integer",
+          "format": "uint",
+          "default": 0,
+          "minimum": 0
+        },
+        "query_cache_ttl": {
+          "description": "Time-to-idle (seconds) for query (AST) cache entries; an entry not accessed within this window is dropped by the maintenance sweep. `0` disables idle expiry.\n\n_Default:_ `0`\n\n<https://docs.pgdog.dev/configuration/pgdog.toml/general/#query_cache_ttl>",
+          "type": "integer",
+          "format": "uint",
+          "default": 0,
           "minimum": 0
         },
         "query_log": {
--- a/.schema/pgdog.schema.json
+++ b/.schema/pgdog.schema.json
@@ -87,9 +87,9 @@
         "prepared_statements_ttl": null,
         "prepared_statements_ttl_jitter": 30000,
         "pub_sub_channel_size": 0,
+        "query_cache_idle_timeout": 0,
         "query_cache_limit": 1000,
         "query_cache_memory_limit": 0,
-        "query_cache_ttl": 0,
         "query_log": null,
         "query_log_stdout": false,
         "query_parser": "auto",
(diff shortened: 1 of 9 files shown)


IgorOhrimenko:
@levkk — a direction check on this one.

The failure mode this PR targets: the query cache is bounded by entry count but not by bytes, and entries never expire. On a workload with large or diverse queries, RSS grows monotonically until the container hits its memory limit — at which point the kernel OOM killer SIGKILLs the process. No drain, no graceful shutdown, exit 137: every client connection is dropped mid-transaction, and it tends to happen at peak traffic, because that's when the cache churns hardest. For a connection pooler — the component that exists to make everything behind it look stable — that's the worst possible way to die.

Both knobs are opt-in (`0` = off), so default behaviour is byte-for-byte unchanged. What they buy is turning "hard kill at an unpredictable time" into a bounded steady state the operator chooses: `query_cache_memory_limit` caps the total, `query_cache_ttl` releases an idle working set without waiting for eviction.

IgorOhrimenko:
Renamed `query_cache_ttl` → `query_cache_idle_timeout` (b78543c9) after a closer look at the naming.

What the option implements is time-to-*idle*, not time-to-*live*: the clock counts from the last access, so a hot entry never expires — only entries nobody has touched for the whole window are swept. A true TTL counts from insertion and would evict hot entries too, which is the wrong tool here (a parsed AST can't go stale, so there's nothing to refresh). Calling a time-to-idle knob `ttl` would over-promise expiry that never happens for hot entries.

The precise term for this is TTI (`timeToIdle` in Ehcache, `AccessedExpiryPolicy` in JCache, `expireAfterAccess` in Caffeine), but it's obscure outside the Java caching world — and the config already has an established vocabulary for exactly this semantic: `idle_timeout`, `client_idle_timeout`. So the option now follows it: `query_cache_idle_timeout`, and the unit changed from seconds to milliseconds to match the other `*_timeout` options. Env var is now `PGDOG_QUERY_CACHE_IDLE_TIMEOUT`; schema regenerated; PR description updated.

IgorOhrimenko:
A note on the Codecov report: the two remaining uncovered lines in `cache_impl.rs` are defensive branches that aren't reachable through the public API.

- Line 101 — the `None => break` arm in `enforce()`. It guards against an infinite eviction loop if the byte accounting ever drifted out of sync with an empty cache. With `insert`/`sweep` keeping `bytes` consistent, `pop_lru()` can't return `None` while a limit is still violated.
- Line 154 — the fallback in `measure_build()` for when jemalloc's per-thread allocation counters are unavailable at runtime. On builds where jemalloc is active (including the coverage CI build), the `Some` path is always taken.

Everything else in the patch, including `configure()` with a non-zero idle timeout and `resize()`, is now covered.

PR #1270, opened 2026-07-27
Persist coordinator gid in the WAL for crash recovery

### LINKED ISSUE #1175: [2pc] SIGTERM may not clean up transactions saved in 2pc WAL
**PgDog version**
 `v0.1.48` 

**Description**
Upon receiving `SIGTERM`, some 2pc transactions were left in an unfinished state when pgdog rebooted. The WAL was on, so they should of been cleaned up, even though SIGTERM causes an immediate shutdown without running the WAL manager.

**Configuration**

2pc WAL was enabled.

crodas:
Attempt to fix #1175

After a SIGTERM and reboot, prepared 2PC transactions were left orphaned on the shards even with the WAL enabled. Recovery replayed the in-flight transactions but drove COMMIT PREPARED / ROLLBACK PREPARED against a gid that no longer matched what Postgres held, so the statements failed and the monitor retried them forever.

The gid a transaction is prepared with embeds a per-process instance_id, which is a fresh random value on every start unless NODE_ID is set. The WAL only persisted the inner transaction id, so a restarted PgDog reconstructed the gid with a different instance_id prefix and missed the orphan. The gid used at recovery must be byte-identical to the gid used at PREPARE, so it has to come from the WAL, not from live process state.

Store the full coordinator gid in the Begin and Checkpoint records (serde default keeps old segments readable) and thread it through recovery so cleanup resolves each prepared xact with the exact gid it was created with. Live transactions still render the gid from process state, which is correct in the process that created them. An empty stored gid falls back to the previous behavior.

The crash-safety WAL helper now records the full gid too, so the integration spec exercises the real recovery path.

### DIFF
--- a/pgdog/src/backend/pool/connection/binding.rs
+++ b/pgdog/src/backend/pool/connection/binding.rs
@@ -365,7 +365,7 @@ impl Binding {
 
     pub(crate) async fn two_pc_on_guards(
         servers: &mut [Guard],
-        transaction: TwoPcTransaction,
+        transaction: &TwoPcTransaction,
         phase: TwoPcPhase,
         ignore_missing: bool,
     ) -> Result<(), Error> {
@@ -399,7 +399,7 @@ impl Binding {
     /// Execute two-phase commit transaction control statements.
     pub(crate) async fn two_pc(
         &mut self,
-        transaction: TwoPcTransaction,
+        transaction: &TwoPcTransaction,
         phase: TwoPcPhase,
         ignore_missing: bool,
     ) -> Result<(), Error> {
--- a/pgdog/src/frontend/client/query_engine/two_pc/statement.rs
+++ b/pgdog/src/frontend/client/query_engine/two_pc/statement.rs
@@ -22,7 +22,7 @@ impl TwoPcTransactionOnShard {
 
     /// Get the coordinator transaction.
     pub(crate) fn transaction(&self) -> TwoPcTransaction {
-        self.transaction
+        self.transaction.clone()
     }
 }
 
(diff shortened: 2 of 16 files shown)


pgdog/src/backend/pool/connection/binding.rs
```diff
@@ -364,14 +364,14 @@ impl Binding {
 
     pub(crate) async fn two_pc_on_guards(
         servers: &mut [Guard],
-        transaction: TwoPcTransaction,
+        coordinator_gid: &str,
```
levkk: I'm not a fan of passing strings around like this (no type safety, easy for a bug to fly through). Could we maybe use the deserialization code in here: https://github.com/pgdogdev/pgdog/blob/ea4a02d09f62e6a42e5c01008f42583e9e6d790f/pgdog/src/frontend/client/query_engine/two_pc/statement.rs#L35

pgdog/src/frontend/client/query_engine/two_pc/statement.rs
```diff
@@ -46,18 +46,39 @@ impl FromStr for TwoPcTransactionOnShard {
...
+/// Same as [`phase_control`] but takes the coordinator gid as an already
+/// rendered string.
+///
+/// WAL recovery must use this: a prepared xact's gid embeds the
+/// `instance_id` of the process that created it, which a restarted PgDog
+/// does not reproduce. The gid therefore has to come from the WAL, not
+/// from re-rendering a [`TwoPcTransaction`] against live process state.
+/// The per-shard participant name is `<coordinator_gid>_<shard>`, matching
+/// [`TwoPcTransactionOnShard`]'s `Display`.
+pub(crate) fn phase_control_gid(coordinator_gid: &str, shard: usize, phase: TwoPcPhase) -> String {
```
levkk: Same comment as above. String identifiers generated by calling `to_string()` or a real Rust type are the root of all evil :)

Issue #1269, opened 2026-07-27
test_read_write_split_exclude_primary_no_primary test is flaky


levkk:
**PgDog version**
`v0.1.50`

**Description**
This test keeps failing randomly in CI.

```
running 1 test
test backend::pool::lb::test::test_read_write_split_exclude_primary_no_primary ... FAILED

failures:

failures:
    backend::pool::lb::test::test_read_write_split_exclude_primary_no_primary

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 2041 filtered out; finished in 0.03s


--- STDERR:              pgdog backend::pool::lb::test::test_read_write_split_exclude_primary_no_primary ---

thread 'backend::pool::lb::test::test_read_write_split_exclude_primary_no_primary' (7805) panicked at pgdog/src/backend/pool/lb/test.rs:555:5:
assertion `left == right` failed
  left: 1
 right: 2
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

https://github.com/pgdogdev/pgdog/actions/runs/30279404035/job/90021641578

PR #1272, opened 2026-07-28
fixed flakey test

KrishnaSindhur:
This should fix the flakey test #1269


levkk:
This removes test coverage for `random` load balancing strategy. Maybe we increase the number of attempts from 10 to 100? This pretty much guarantees the random number generator produces at least two results, I believe?

KrishnaSindhur:
> This removes test coverage for `random` load balancing strategy. Maybe we increase the number of attempts from 10 to 100? This pretty much guarantees the random number generator produces at least two results, I believe?

ok if we have to use random, then we have to change the existing way, which is with 2 pools and 10 random tries!!
This will work bumping 10 → 100

PR #1273, opened 2026-07-28
feat: support readable time values in config

l4l:
## Human-readable durations in config

Time settings now accept strings like `"1h5m15s"` alongside plain numbers, so `ban_timeout = 300000` can be written as `ban_timeout = "5m"`.
Motivation: make these value readable by mere mortals.
The spaces can be put anywhere, so " 1 h 2 m " also should work. The plain numbers still work and handled as before, this is an extension for params view.

New `pgdog-config/src/duration.rs` parses `ms`/`s`/`m`/`h`/`d` via serde(with=..) for all time fields. `PGDOG_*` env vars supports the same syntax.

Limitations:
- only single symbol for easier parsing ("1s" works but not "1sec");
- no support for admin commands like `SET ban_timeout TO '5m'`;
- only integers (i.e no "1.5s");

### DIFF
--- a/.schema/pgdog.schema.json
+++ b/.schema/pgdog.schema.json
@@ -409,12 +409,15 @@
         },
         "idle_timeout": {
           "description": "Overrides the `idle_timeout` setting. Idle server connections exceeding this timeout will be closed automatically.\n\n<https://docs.pgdog.dev/configuration/pgdog.toml/databases/#idle_timeout>",
-          "type": [
-            "integer",
-            "null"
+          "anyOf": [
+            {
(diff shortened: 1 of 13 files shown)


l4l:
> 2026-07-28T03:57:42.309299Z ERROR pgdog::net::stream: [127.0.0.1:47602] <-- FATAL: 28000 password for user "pgdog" and database "pgdog" is wrong, or the database does not exist

not sure if i broke it myself, perhaps some issue with ci credentials or missing waiting for startup time?

PR #1275, opened 2026-07-28
Feat/multi statement query support

hamin:
## Context and motivation

At my company we're using PgDog as part of a PostgreSQL version upgrade migration strategy. We have a polyglot environment with services across different languages, ORMs, query builders, and driver versions. Some have hard dependencies we can't control (in our case, Kong) that generate raw multi-statement simple queries we couldn't inspect or modify. The goal was safety in the read path without having to audit every query generated across every language and library. We're not using sharding, so the intention was to keep changes minimal there.

## What changed

- Removed `MultiStatementMixedSet` — mixed batches no longer error
- Router routes multi-statement non-SET batches to the write primary instead of only using the first statement for routing
- Added `ClientRequest::spliced_simple()` behind `#[cfg(feature = "new_parser")]` that splits a `Q` message into per-statement requests using `stmt_location`/`stmt_len` from `pg_raw_parse`
- Wired `spliced_simple()` into the client loop after `spliced()`, so simple-query batches go through the same splice processing path as extended-protocol pipelines when the new parser is enabled
- Added `simple_query_splice` context flag and `suppress_rfq` logic in four places (`query.rs`, `fake.rs`, `start_transaction.rs`, `end_transaction.rs`) to hold back intermediate `ReadyForQuery` messages so the client sees exactly one `Z` at the end, matching simple query protocol semantics
- All-`SET` batches continue to be intercepted by PgDog's parameter tracking; single-statement queries and parse failures fall through unchanged

**Default builds** (without `new_parser`): multi-statement batches route to the write primary as a whole. `spliced_simple()` returns empty vec and callers fall through.

**`new_parser` builds**: full per-statement splitting with `SET` interception and correct `ReadyForQuery` handling.

## On the complexity

The `suppress_rfq` work across four handlers is more invasive than I originally wanted. It was necessary because `SET`, `BEGIN`, and `COMMIT` all have code paths that generate synthetic `ReadyForQuery` responses without going through the backend, so each one needed the same fix. The motivation for splitting rather than forwarding verbatim is that libraries we can't control frequently send batches like `SET statement_timeout TO '30s'; SELECT ...` as part of connection setup. If PgDog forwards the whole `Q` verbatim it never sees the `SET`, and when that backend connection gets returned to the pool the next client inherits stale parameter state silently.

If you feel this is too much complexity for what's being solved, or there are implications for the broader codebase I'm not seeing, happy to scale this back to just the routing fix.

## On @sgrif's review comments

- **Old parser / `new_parser` feature flag**: `spliced_simple()`, `split_statements()`, and the related unit tests are now all gated behind `#[cfg(feature = "new_parser")]` using `pg_raw_parse`. Default builds fall through unchanged.
- **`&str -> Vec<&str>` signature**: done, `split_statements` returns slices directly now
- **Error propagation in `split_statements`**: parse failures fall through to the single-statement path, which forwards to the backend and lets Postgres return the error to the client. Feels safer than surfacing a proxy-level parse error for something the backend might handle fine
- **Should router error if it sees more than one statement**: with the latest changes this code path no longer sees multi-statement batches directly when `new_parser` is enabled. `spliced_simple()` handles the split upstream before routing runs so each statement routes individually. The `stmts.len() > 1` guard in the router is now just a fallback for parse failures, session mode, and default builds
- **Implicit transactional behavior**: this was the right concern to raise. Added tests documenting the behavior. PgDog splits the batch so each statement gets its own auto-commit rather than an implicit transaction wrapping the whole thing. `test_multi_statement_implicit_transaction_behavior` documents this as a known divergence from standard PostgreSQL behavior

## Testing

I don't have permissions to run CI on this repo so I iterated locally. I ran the full Rust integration suite (`cargo nextest run --profile integration`) and the Go, Python, and SQL suites directly against a locally built binary. It took a bit of work to get local integration testing set up properly but wanted to make sure I wasn't breaking anything as I iterated. The failures I saw were all pre-existing environment issues (sharded pool config, unique_id node config, bundler version mismatch) and unrelated to these changes. The 9 new multi-statement tests all pass and I didn't see regressions in the suites I was able to run.

Note: the splitting behavior and new tests only activate under `new_parser`. The default build integration tests exercise the routing-only path.

### DIFF
--- a/pgdog/src/frontend/client_request.rs
+++ b/pgdog/src/frontend/client_request.rs
@@ -7,11 +7,6 @@ use std::ops::{Deref, DerefMut};
 use lazy_static::lazy_static;
 use regex::Regex;
 
-#[cfg(not(feature = "new_parser"))]
-use pg_query;
-#[cfg(feature = "new_parser")]
-use pg_raw_parse;
-
 use crate::{
     frontend::router::Ast,
     net::{
@@ -25,40 +20,26 @@ use super::{PreparedStatements, router::Route};
 
 pub use super::BufferedQuery;
 
-#[cfg(feature = "new_parser")]
-fn split_statements(query: &str) -> Vec<(usize, usize)> {
-    let Ok(parsed) = pg_raw_parse::parse(query) else {
-        return vec![];
-    };
-    let inner = parsed.into_inner();
-    if inner.len() <= 1 {
-        return vec![];
-    }
-    inner
-        .into_iter()
-        .map(|stmt| {
-            let loc = stmt.stmt_location.max(0) as usize;
-            let len = stmt.stmt_len as usize;
-            (loc, len)
-        })
-        .collect()
-}
-
-#[cfg(not(feature = "new_parser"))]
-fn split_statements(query: &str) -> Vec<(usize, usize)> {
+/// Split a multi-statement simple query string into individual statement slices.
+/// Returns an empty vec if parsing fails or the query contains only one statement.
+fn split_statements(query: &str) -> Vec<&str> {
     let Ok(parsed) = pg_query::parse(query) else {
         return vec![];
     };
-    let stmts = &parsed.protobuf.stmts;
+    let stmts = parsed.protobuf.stmts;
     if stmts.len() <= 1 {
         return vec![];
     }
     stmts
         .iter()
-        .map(|stmt| {
+        .filter_map(|stmt| {
             let loc = stmt.stmt_location.max(0) as usize;
-            let len = stmt.stmt_len as usize;
-            (loc, len)
+            let end = if stmt.stmt_len == 0 {
+                query.len()
+            } else {
+                loc + stmt.stmt_len as usize
+            };
+            query.get(loc..end).map(str::trim).filter(|s| !s.is_empty())
         })
         .collect()
 }
@@ -419,29 +400,17 @@ impl ClientRequest {
     /// the request is not a simple Query, contains only one statement, or parsing fails.
     /// In all those cases the caller falls through to the normal query engine path.
     pub fn spliced_simple(&self) -> Vec<Self> {
-        let Some(ProtocolMessage::Query(q)) = self.messages.iter().find(|m| m.code() == 'Q') else {
+        let Some(ProtocolMessage::Query(q)) = self.messages.iter().find(|m| m.code() == 'Q')
+        else {
             return vec![];
         };
-        let query_text = q.query();
-
-        split_statements(query_text)
+        split_statements(q.query())
             .into_iter()
-            .filter_map(|(offset, len)| {
-                let end = if len == 0 {
-                    query_text.len()
-                } else {
-                    offset + len
-                };
-                let text = query_text.get(offset..end)?.trim();
-                if text.is_empty() {
-                    return None;
-                }
-                Some(Self {
-                    messages: vec![ProtocolMessage::Query(Query::new(text))],
-                    route: None,
-                    ast: None,
-                    last_parse: None,
-                })
+            .map(|text| Self {
+                messages: vec![ProtocolMessage::Query(Query::new(text))],
+                route: None,
+                ast: None,
+                last_parse: None,
             })
             .collect()
     }
--- a/pgdog/src/frontend/client_request.rs
+++ b/pgdog/src/frontend/client_request.rs
@@ -465,6 +465,13 @@ mod test {
 
     use super::*;
 
+    #[cfg(not(feature = "new_parser"))]
+    #[test]
+    fn test_spliced_simple_returns_empty_without_new_parser() {
+        let req = ClientRequest::from(vec![Query::new("SELECT 1; SELECT 2").into()]);
+        assert!(req.spliced_simple().is_empty());
+    }
+
     #[cfg(feature = "new_parser")]
     #[test]
     fn test_spliced_simple_two_selects() {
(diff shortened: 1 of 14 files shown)


hamin:
@levkk :) 

First of thank you for pgcat and pgdog! 
cc would love for you take look when have the opportunity! Also pinged you on Discord FYI

pgdog/src/frontend/client_request.rs
```diff
@@ -20,6 +25,44 @@ use super::{PreparedStatements, router::Route};
 
 pub use super::BufferedQuery;
 
+#[cfg(feature = "new_parser")]
+fn split_statements(query: &str) -> Vec<(usize, usize)> {
```
sgrif: Should the signature of this be `&str -> Vec<&str>`?

pgdog/src/frontend/client_request.rs
```diff
@@ -20,6 +25,44 @@ use super::{PreparedStatements, router::Route};
...
+    let Ok(parsed) = pg_raw_parse::parse(query) else {
+        return vec![];
```
sgrif, line 33: Why are we ignoring errors here? Do you think we should propagate them?

pgdog/src/frontend/router/parser/query/mod.rs
```diff
@@ -290,19 +290,18 @@ impl QueryParser {
...
+                return Ok(command);
+            }
+            // Mixed or non-SET batch: forward the full query to the write primary.
+            return Ok(Command::Query(Route::write(
+                context.shards_calculator.shard(),
+            )));
         }
 
-        //
-        // Get the root AST node.
```
sgrif, line 301: Now that we are relying on there only being one query per command, should we return an error if we got more than one?

pgdog/src/frontend/router/parser/query/mod.rs
```diff
@@ -555,19 +554,18 @@ impl QueryParser {
 
                 let stmts = &statement.parse_result().protobuf.stmts;
 
-                // Handle multi-statement SET commands (e.g. "SET x TO 1; SET y TO 2").
```
sgrif, line 558: The old parser is considered legacy code, new features don't need to support it.

pgdog/src/frontend/client_request.rs
```diff
@@ -20,6 +25,44 @@ use super::{PreparedStatements, router::Route};
...
+        .into_iter()
+        .map(|stmt| {
+            let loc = stmt.stmt_location.max(0) as usize;
+            let len = stmt.stmt_len as usize;
+            (loc, len)
+        })
+        .collect()
+}
+
+#[cfg(not(feature = "new_parser"))]
```
sgrif: New features don't need to support the old parser.

sgrif (changes requested):
I don't think this handles the implicit transactional behavior of a multi-statement query. Can you add some integration tests for that?

pgdog/src/frontend/client_request.rs
hamin: gottya

pgdog/src/frontend/router/parser/query/mod.rs
hamin: ah awesome didn't know that. This makes perfect sense. Will remove this!

hamin:
> I don't think this handles the implicit transactional behavior of a multi-statement query. Can you add some integration tests for that?

@sgrif thanks for your review and the is the biggest point you mentioned. I was hoping to see tests run on CI to see if I was failing anything but managed to finally run them on my machine and yeah saw that this definitely broke things...taking a different approach but will ping you all soon

pgdog/src/frontend/client_request.rs
hamin: I updated the PR description for more context and my decision on this. Let me know how u feel

pgdog/src/frontend/router/parser/query/mod.rs
hamin: I addressed this in the PR description...let me know what u think

hamin:
OK @sgrif , thank u for catching the implicit transaction behavior! Force me to make sure integration tests pass locally before making further changes. I think I took @levkk 's initial comments in a Discord thread a bit too literally trying to share some code eventhough his comment was more about approach. I gave @levkk a bit more context in the Discord channel but I updated the PR description with what motivated this PR to begin with.


@sgrif @levkk would love a review/comments/feedback. FYI if you all feel this is way off base or conflicts with the future decisions that I'm completely unaware please let me know. Totally dogfooding this because we have a big PG upgrade migration that we really want to test out and use pgdog with to introduce it to our toolset not just for this but other purposes too (maybe sharding too but unsure of that currently)

pgdog/src/frontend/client_request.rs
```diff
@@ -20,6 +20,30 @@ use super::{PreparedStatements, router::Route};
 
 pub use super::BufferedQuery;
 
+/// Split a multi-statement simple query string into individual statement slices.
+/// Returns an empty vec if parsing fails or the query contains only one statement.
+fn split_statements(query: &str) -> Vec<&str> {
+    let Ok(parsed) = pg_query::parse(query) else {
```
sgrif: Sorry, I should have been more clear. This shouldn't use `pg_query` with the old parser, we can just feature flag the change behind the new parser for now.

hamin: Gottya, anything else?

hamin: Ok, updated `spliced_simple()`, `split_statements()`, and the related unit tests to all be gated behind `#[cfg(feature = "new_parser")]` using `pg_raw_parse`. On a default build `spliced_simple()` returns  empty vec and callers fall through to the existing path.

PR #1276, opened 2026-07-28
Remove dead `Oids` type

sgrif:
This struct doesn't actually contain the information needed to handle mapping of user-defined types. And such a type would need to live at a different level than this one. An actual mapping will include types which are not `Copy`, and we'd likely want to pass it around by reference. As such it would need to live outside of a type protected by a Mutex.


levkk:
I think a safe assumption you can make is anything loaded by the cluster/pool is good for the lifetime of that struct. This is why we call `reload_from_existing` when we detect DDL - this reloads the schema from the database, which could also include OIDs.

Agreed that these things living on the pool is kinda weird. Maybe on the `Shard` is better, since we'll fetch them from the primary, just like the schema.

sgrif:
So it will need to get passed through the pool somehow, as the place we ultimately need the reference to this is on `PreparedStatements` (which also gets cleared on DDL), but yeah `Shard` is where I'd like the ownership to be if I can make it work. The issue isn't with it living on `Pool` being weird, but that it was living on a type protected by a `Mutex`, and we want to be able to hold references to this without requiring a lock

PR #1279, opened 2026-07-28
feat: sharding key lookups with fixed shard placement

rlittlefield:
# Sharding key lookups: route by asking the database, with fixed shard placement

Sharded tables gain a `lookup_query`: instead of deriving the shard from
the key itself, pgdog asks the database. Combined with
`lookup_result = "shard"`, the application owns placement outright and
the shard count can change without moving any rows. That property is the
foundation for growing a cluster in place, which two follow-up PRs build
on (fleet coordination, then ADD SHARD).

```toml
[[sharded_tables]]
database = "prod"
column = "tenant_id"
data_type = "bigint"
lookup_query = "SELECT shard_id FROM tenants WHERE id = $1"
lookup_result = "shard"
```

## How routing works

- The lookup runs lazily, per key, with the sharding key bound as `$1`.
  The query executes on one shard picked round robin. The lookup table
  must be omnisharded, so any shard is authoritative.
- Results are cached per cluster, bounded by approximate memory use with
  LRU eviction. `sharding_lookup_cache_size` and
  `sharding_lookup_timeout` in `[general]` tune it. Cache hits run with
  fully borrowed keys and allocate nothing.
- `lookup_result = "value"` hashes the translated value.
  `lookup_result = "shard"` treats it as the 0-based shard number and
  never hashes. A returned shard outside the cluster's range fails the
  statement loudly rather than routing it anywhere wrong.
- A value with no row fails the statement. Absence is not cached, so a
  row inserted later is picked up on the next statement without any
  invalidation step.
- The first routing pass collects cache misses as pending lookups. The
  query engine resolves them and routes once more, handing the resolved
  translations to the second pass so it cannot miss. Routing stays
  deterministic.
- Lookups apply everywhere keys appear: statements, prepared statement
  parameters (text and binary), `COPY` rows, `pgdog_sharding_key`
  comments, `SET pgdog.sharding_key`, and replicated changes during
  resharding data sync (which runs its lookups on the source cluster,
  the one with the complete mapping).
- Omnisharded tables always win over a sharding key comment: a comment
  can pick a shard for sharded tables, but writes to omni tables always
  broadcast.

## Observability

Lookup metrics (hits, misses, evictions, lookup latency) are exposed on
cluster stats and the OpenMetrics endpoint. The former mirror stats
structure was absorbed into `ClusterMetrics`, which carries both.

## Also in this PR

- Debug builds default `[memory] stack_size` to 8 MiB, and test-harness
  threads get `RUST_MIN_STACK = 4 MiB` via `.cargo/config.toml`.
  Unoptimized query-path futures are several times larger than release
  and overflow the release-tuned 2 MiB default on deep parse trees. The
  two knobs cover different thread populations: `[memory] stack_size`
  sizes the runtime's workers, while libtest and nextest spawn their own
  threads that never read pgdog.toml. Release builds are unchanged.
- Config validation refuses lookup queries that are not a single SELECT
  with a `$1` placeholder, and refuses `lookup_result = "shard"` without
  a `lookup_query`. There is no way to configure a lookup that silently
  falls back to hashing.


pgdog-config/src/core.rs
```diff
@@ -120,6 +120,23 @@ impl ConfigAndUsers {
         self.config.check();
         self.users.check(&self.config);
         self.validate_server_auth()?;
+        self.validate_sharded_tables()?;
+        Ok(())
+    }
+
+    fn validate_sharded_tables(&self) -> Result<(), Error> {
+        for table in &self.config.sharded_tables {
+            if let Some(query) = &table.lookup_query
```
levkk: nit: we should use `pg_parse_raw::parse` here:

1. will validate that the query is syntactically valid
2. doesn't rely on string interpolation

pgdog/src/backend/replication/logical/subscriber/copy.rs
```diff
@@ -352,9 +366,24 @@ impl CopySubscriber {
...
+        let mut result = self.copy.shard(&self.buffer)?;
         self.buffer.clear();
 
+        // Rows whose sharding key isn't in the lookup cache are
+        // deferred; resolve the lookups and shard them now. The
+        // destination's copy of the lookup table may still be syncing,
+        // so lookup queries run on the source, which has the complete
+        // mapping, and fill the destination's cache, which the row
+        // sharder reads.
+        let pending = self.copy.pending_lookups();
```
levkk: nit: copy/pasta.

pgdog/src/backend/server.rs
```diff
@@ -855,6 +856,53 @@ impl Server {
             .collect::<Result<Vec<T>, _>>()?)
     }
 
+    /// Execute a query with text-format parameters using the extended
+    /// protocol and return all rows. Parameters are bound by the server,
+    /// so values don't need to be escaped.
+    pub async fn fetch_all_params<T: From<DataRow>>(
```
levkk, line 949: refactor: we should change `fetch_all` and consequently `execute_checked` and `execute` to accept a generic type that resolves to either a `Vec<Query>` or `Vec<"extended protocol request">`. The protocol handling below is brittle and needs to re-use existing logic.

pgdog/src/frontend/client/query_engine/route_query.rs
```diff
@@ -91,8 +92,41 @@ impl QueryEngine {
...
+        // Resolve sharding key lookups that missed the cache and route
+        // the query again, this time with a warm cache. A lookup that
+        // can't be resolved fails the statement: routing by the
+        // untranslated value could put it on the wrong shard.
+        if result.is_ok() {
+            let pending = self.router.take_pending_lookups();
+            if !pending.is_empty() {
+                match lookup::resolve(cluster, pending).await {
+                    Ok(()) => {
+                        let router_context = RouterContext::new(
```
levkk, line 111: nit: I'd probably construct a new router here instead of relying on internals and calling `take_pending_lookups()` twice.

pgdog/src/backend/replication/sharded_tables.rs
```diff
@@ -21,6 +21,9 @@ struct Inner {
     common_mapping: Option<CommonMapping>,
     omnisharded_sticky: bool,
     system_catalogs: SystemCatalogsBehavior,
+    /// Cached sharding key lookup query results. Scoped here so the
+    /// cache is recreated, i.e. emptied, on config reload.
+    lookup_cache: LookupCache,
```
levkk, line 28: I lost track of this one. We want to double check that it lives on the `Cluster`, i.e., we get a new version of this when we call `reload_from_existing` or `reload`.

levkk:
Looks pretty good! I need to read it some more, but architecturally it's sound. Just a few nits around code maintenability / quality, but nothing major/blocking I think.

pgdog/src/frontend/client/query_engine/route_query.rs
```diff
@@ -91,8 +92,41 @@ impl QueryEngine {
...
-        match self.router.query(router_context) {
-            Ok(command) => {
+        let mut result = self.router.query(router_context).map(|_| ());
+
+        // Resolve sharding key lookups that missed the cache and route
+        // the query again, this time with a warm cache. A lookup that
+        // can't be resolved fails the statement: routing by the
+        // untranslated value could put it on the wrong shard.
+        if result.is_ok() {
+            let pending = self.router.take_pending_lookups();
```
levkk: These state modifications are hard to track. I think we introduced this pattern and should not do it anymore. Maybe we can return these as part of the `Route`? That one is created for each request, so no concerns about leaking state between calls.

pgdog/src/frontend/router/parser/query/mod.rs
```diff
@@ -120,6 +127,10 @@ impl QueryParser {
 
     /// Parse a query and return a command.
     pub fn parse(&mut self, context: RouterContext) -> Result<Command, Error> {
+        // Pending lookups are per-statement; don't leak them into
+        // the next statement if this parser is reused.
+        self.pending_lookups.clear();
```
levkk: Same comment re: `Route` as above. We shouldn't leak this between calls like this in the first place.

pgdog/src/frontend/router/parser/query/select.rs
```diff
@@ -101,10 +102,13 @@ impl QueryParser {
                         context.router_context.parameter_hints.search_path,
                     ),
                     statement_parser.extract_tables(),
+                    pending_lookups,
                 )
             }
         };
 
+        self.pending_lookups.extend(pending_lookups);
```
levkk: Maybe move this to `QueryParserContext`?  A new one is created with each request, so data between calls won't leak.

pgdog/src/frontend/router/parser/copy.rs
```diff
@@ -367,23 +467,61 @@ impl CopyParser {
...
                         if tuple.end() {
                             let terminator = (-1_i16).to_be_bytes();
-                            rows.push(CopyRow::new(
-                                &terminator,
-                                self.schema_shard.clone().unwrap_or(Shard::All),
-                            ));
+                            // The terminator must ship after every data row;
+                            // hold it back while rows are deferred.
+                            if !self.deferred.is_empty() {
+                                self.deferred.push(DeferredRow {
```
levkk: Maybe we make this method async instead? I get the defer pattern, but it sure makes things hard to follow. I'm _pretty_ sure (i.e., not sure at all) that this method only gets called in async contexts anyway.

pgdog/src/frontend/router/parser/statement.rs
```diff
@@ -1356,64 +1360,107 @@ impl<'a, 'b: 'a, 'c> StatementParser<'a, 'b, 'c> {
...
 
+    /// Translate a sharding key value through the table's lookup, if one
+    /// is configured. Returns the translated value on a cache hit. On a
+    /// cache miss, records a pending lookup and returns `None`; the
+    /// query engine resolves it and routes the statement again, or
+    /// fails it if the lookup table has no row for the value.
+    fn translate_sharding_key(&mut self, table: &ShardedTable, value: &str) -> Option<Arc<str>> {
+        let query = table.lookup_query.as_deref()?;
+
+        match self.schema.tables.lookup_cache().get(query, value) {
```
levkk: This should really be a lookup by `ShardedTable` (name, schema, database is the fully-qualified lookup). It's possible (but unlikely) to have two sharded databases with the same lookup query, and different lookup tables, so this could route things incorrectly.

pgdog/src/frontend/router/sharding/lookup.rs
```diff
@@ -0,0 +1,191 @@
...
+    let schema = cluster.sharding_schema();
+    resolve_with(cluster.shards(), schema.tables.lookup_cache(), pending).await
+}
+
+/// Like [`resolve`], with the servers running the queries decoupled
+/// from the cache the results fill. Resharding data sync runs lookups
+/// on the source cluster, which has the complete mapping while the
+/// destination is still syncing, and fills the destination's cache,
+/// which the row sharder reads.
+pub(crate) async fn resolve_with(
```
levkk, line 453: Need to think about this one. Nothing wrong with it, but a lot of that code is duplicated, and I'm thinking we should either move it or add a function to `Cluster` to execute a query against a shard using round robin.

pgdog/src/frontend/router/sharding/lookup.rs
```diff
@@ -0,0 +1,191 @@
...
+    pub query: String,
+    /// Sharding key value, normalized to text.
+    pub value: String,
+}
+
+/// Sharding key lookup cache.
+#[derive(Debug, Default, Clone)]
+pub struct LookupCache {
+    // Keyed by lookup query.
+    cache: Arc<DashMap<String, Translations>>,
```
levkk: Another thing that came up lately is we really should have memory-bound/unit-bound caching for this kind of stuff. Maybe use https://github.com/moka-rs/moka? We otherwise have the `lru` crate for a simpler LRU cache, which we could also use here, albeit it won't be sharded, so maybe that's why moka would be a better choice.

levkk:
image

I think we just need to make all hardcoded timeouts / attempts configurable and this might be it. I need to double check the parser changes a couple more times, that one is always hard to review in GitHub because of how large that file is, but if the tests are passing, I suspect it's a-ok.

levkk:
Oh and before I forget. Metrics! We need to track lookup queries: cache hits and misses and actual cache size + evictions. That one is gonna be an important metric of performance. For cache misses, we should track the latency of lookups.

For those, I would follow the cluster-global metrics pattern https://github.com/pgdogdev/pgdog/blob/b2a61ddcb5067f930af819fd537e2a261edfe02d/pgdog/src/backend/pool/cluster.rs#L506

... which should be renamed to `ClusterMetrics` of course. Currently used for mirroring stats only, but this would expand it.

rlittlefield:
> Oh and before I forget. Metrics! We need to track lookup queries: cache hits and misses and actual cache size + evictions. That one is gonna be an important metric of performance. For cache misses, we should track the latency of lookups.
> 
> For those, I would follow the cluster-global metrics pattern
> 
> https://github.com/pgdogdev/pgdog/blob/b2a61ddcb5067f930af819fd537e2a261edfe02d/pgdog/src/backend/pool/cluster.rs#L506
> 
> ... which should be renamed to `ClusterMetrics` of course. Currently used for mirroring stats only, but this would expand it.

Just added metrics and configurable lookup timeout and lookup attempts!

pgdog/src/config/mod.rs
```diff
@@ -74,6 +75,78 @@ pub fn set(mut config: ConfigAndUsers) -> Result<ConfigAndUsers, Error> {
...
+            let mut found = HashSet::new();
+            for stmt in ast.stmts() {
+                pg_raw_parse::walk::walk(stmt, |node| {
+                    if let pg_raw_parse::Node::ParamRef(param) = node {
+                        found.insert(param.number);
+                    }
+                });
+            }
+            params = found;
+        }
```
sgrif: The new parser version of this can be quite a bit simpler by breaking early if we ever see a parameter with a number > 1, and seeing if we got a value.

```suggestion
        {
            let ast = pg_raw_parse::parse(query)
                .map_err(|err| error(format!("\"lookup_query\" is invalid: {}", err)))?;
            let stmt = ast.stmts().exactly_one().map_err(|_| error("\"lookup_query\" must be a single statement".into()))?;
            if pg_raw_parse::walk::walk_manual(stmt, |node| match node {
                Node::ParamRef(param) if param.number != 1 => ControlFlow::Break(()),
                _ => Recurse::yes(),
            }).is_some() {
                return Err(error(
                    "\"lookup_query\" must reference exactly one parameter, \"$1\"".into(),
                ));
            }
        }
```

sgrif: (That's going to need some extra imports and formatting)

sgrif: I would also split the whole function in two rather than a single statement, but we're also considering the old parser to be legacy at this point so it's completely fine to just not support this feature without using the new parser

pgdog/src/frontend/router/parser/comment/directive.rs
```diff
@@ -11,7 +11,7 @@ use super::super::Shard;
 pub(super) static SHARD: Lazy<Regex> =
     Lazy::new(|| Regex::new(r#"pgdog_shard: *([0-9]+)"#).unwrap());
 pub(super) static SHARDING_KEY: Lazy<Regex> = Lazy::new(|| {
-    Regex::new(r#"pgdog_sharding_key: *(?:"([^"]*)"|'([^']*)'|([0-9a-zA-Z_-]+))"#).unwrap()
+    Regex::new(r#"pgdog_sharding_key: *(?:"([^"]*)"|'([^']*)'|([0-9a-zA-Z-]+))"#).unwrap()
```
levkk, line 14: Just want to double check this isn't a breaking change (since I see tests were changed too).

pgdog/src/backend/pool/cluster.rs
```diff
@@ -619,6 +634,29 @@ impl Cluster {
         self.resharding_copy_retry_max_attempts
     }
 
+    /// How long a sharding key lookup query can run before the
+    /// statement waiting on it fails. Zero (a bare test cluster
+    /// built without config) falls back to the config default.
+    pub fn sharding_lookup_timeout(&self) -> Duration {
```
levkk, line 711: Defaults should propagate from `pgdog-config/general.rs`.

pgdog/src/backend/replication/logical/subscriber/context.rs
```diff
@@ -71,7 +71,7 @@ impl StreamContext {
             }
 
             attempts += 1;
-            if attempts > 3 {
+            if attempts > cluster.sharding_lookup_routing_attempts() {
```
levkk: I'm a bit hazy on this one: why do we have attempts at all? A lookup failure indicates that 1) sharding map is out of date or 2) the database isn't reachable.

I guess it could reduce the failure rate in case of 2). Just thinking out loud, this piece kinda stood out to me since we don't do retries anywhere else in the query engine, so why is this place special?

pgdog/src/stats/lookup.rs
```diff
@@ -0,0 +1,157 @@
...
+            "Lookup cache entries evicted to stay within the memory bound.",
+            "counter",
+        );
+        let mut queries = Series::new(
+            "sharding_lookup_queries",
+            "Lookup queries run to resolve cache misses.",
+            "counter",
+        );
+        let mut query_time = Series::new(
+            "sharding_lookup_query_time_microseconds",
```
levkk: We use ms with floating point 3 point precision everywhere else.

pgdog/src/frontend/router/parser/comment/directive.rs
rlittlefield: We use uxid style ids, so we had `org_1234` as the sharding key, but I didn't realize that there was a quoting option. So in an earlier commit I added the _ to the list, but I think that rather than add support for _, I'm just going to put the quotes around it on my app side. I think the net difference in this PR should be no actual changes to this Regex::new call.

pgdog/src/config/mod.rs
```diff
@@ -74,6 +75,78 @@ pub fn set(mut config: ConfigAndUsers) -> Result<ConfigAndUsers, Error> {
...
+            let ast = pg_query::parse(query)
+                .map_err(|err| error(format!("\"lookup_query\" is invalid: {}", err)))?;
+
+            if ast.protobuf.stmts.len() != 1 {
+                return Err(error("\"lookup_query\" must be a single statement".into()));
+            }
+
+            params = ast
+                .protobuf
+                .nodes()
```
sgrif: I'm not sure this will actually work, `pg_query::protobuf::ParseResult::nodes` does not actually iterate over all nodes as you might expect from the name.

rlittlefield: Fixed in the current head. The new-parser path switched to pg_raw_parse::walk_manual, which does visit every subtree (breaks on any ParamRef != 1, requires a $1). The legacy pg_query path still uses nodes() but now documents it as best-effort.

Because this is just a config-time validation, I think the correctness is still generally going to be okay because it still won't misroute, and will instead produce an error.

pgdog/src/stats/lookup.rs
rlittlefield: Fixed! Now calling the metric `sharding_lookup_query_time`, and the help text indicates milliseconds.

pgdog/src/frontend/router/sharding/lookup.rs
```diff
@@ -0,0 +1,721 @@
...
+            Some(translated) => {
+                // Shard-mode lookups never hash: the translation is
+                // the shard number.
+                if lookup.result == LookupResult::Shard {
+                    return Ok(ShardOrLookup::Shard(parse_shard_index(
+                        translated.as_ref(),
+                        schema.shards,
+                    )?));
+                }
+                let ctx = ContextBuilder::infer_from_from_and_config(translated.as_ref(), schema)?
```
levkk, line 258: Can't believe that function name slipped through...that was in August last year, it wasn't even that crazy of a month... :)

levkk (approved):
Let's roll

PR #1280, opened 2026-07-28
Handle custom types with differing OIDs across shards

sgrif:
Handle custom types with differing OIDs across shards

At the wire protocol level, type information is entirely handled by OID, not type name. An OID is a unique identifier that PG assigns for virtually every schema related entity. They are extremely dependent on the order that plugins are loaded, DDL is run, etc. Because of this it's quite likely that a user or extension defined type will not have the same OID across multiple shards.

To solve this, we need to do three things:

- Ensure clients receive a single canonical answer when they query for
  type information
- Rewrite any messages being sent to the server to ensure they have the
  correct OIDs for a given type for that sever
- Rewrite any messages being sent to the client to ensure they have the
  canonical OIDs for a given type

This commit's solution to ensuring the client gets canoincal type information is straightforward, but not ideal. We look for any select statements that could be loading type information (referencing the relevant tables in `pg_catalog`, or doing certain casts), and route them to shard 0.

This means that we have a single point of failure for schema information. Ideally we would be responding to these queries ourselves rather than sending them to the server, but the breadth of different ways different client libraries will load this type information would mean we essentially have to implement our own shitty SQL server. It's possible, and probably necessary in the long run, but outside of the scope of this specific fix.

It's not clear that there'd be any reason one shard would be better than another for the canonical information, but if there is one, we could probably surface this as a configuration option. But since there's no clear reason for it, we can just do the reasonable default.

For rewriting messages, we need to keep track of the mappings for each shard on our end. This was actually somewhat tricky to make work, as schema loading happens on `Shard`, but the message rewriting needs to happen all the way down in `Server`, and these are not types that share an API bounary.

To rectify this, we use a new primitive, which is the async equivalent of `std::sync::OnceLock`. It ensures that only one writer will ever attempt to write to it, while also allowing waiters to wait for a value to be set without attempting to initialize themselves.

With this primitive, all we need to do is stick it in an `Arc` to have it shared in as many disparate places as we need. We use the `SchemaCache` that was previously introduced to ensure we only have one copy of this mapping per database, regardless of how many user/clusters we have, but its use is relatively minimal.

Since this sits in a very hot path, I've done as much as I could to avoid excess allocations or loops if they're not strictly necessary. I'm assuming that the majority of users won't have custom types that vary between shards, so this change should ideally be free for those users, short of the query to determine that there are no mappings at startup.

Tests were a little bit trickier. Resolving the OIDs happens at cluster launch, and the code that needs to do the rewriting will never run without the cluster launching and loading first. But many of our tests were constructing clusters without launching them.

I worked around this by making `Oids::default` return an instance that is already resolved to an empty set of mappings. Since the normal construction requires canonical type information to be passed in, we don't need to worry about production code accidentally calling this incorrectly.

### DIFF
--- a/integration/rust/Cargo.toml
+++ b/integration/rust/Cargo.toml
@@ -6,6 +6,9 @@ edition = "2024"
 [lib]
 test = true
 
+[features]
+new_parser = []
+
 [dependencies]
 tokio-postgres = {version = "0.7.13", features = ["with-uuid-1"]}
 sqlx = { version = "0.8.6", features = ["postgres", "runtime-tokio", "tls-native-tls", "bigdecimal", "chrono", "json", "rust_decimal"]}
--- a/pgdog/src/backend/pool/cluster/schema_loader.rs
+++ b/pgdog/src/backend/pool/cluster/schema_loader.rs
@@ -31,24 +31,18 @@ impl SchemaLoader for FromServer {
             let canonical_oids = Arc::clone(&cluster.canonical_oids);
             let shard = shard.clone();
             tasks::spawn("load canonical oids", async move {
-                // FIXME: This shutdown signal/retry if error logic seems like
-                // it might be worth abstracting
-                let shutdown = tasks::shutdown_signal();
-
                 loop {
-                    let loader = async {
-                        canonical_oids
-                            .load(&mut *shard.primary_or_replica(&Default::default()).await?)
-                            .await
-                    };
-                    let result = select! {
-                        _ = shutdown.cancelled() => break,
-                        result = loader => { result },
-                    };
+                    let result = tasks::shutdown_signal()
+                        .run_until_cancelled(async {
+                            canonical_oids
+                                .load(&mut *shard.primary_or_replica(&Default::default()).await?)
+                                .await
+                        })
+                        .await;
 
                     match result {
-                        Ok(_) => break,
-                        Err(err) => {
+                        Some(Ok(_)) | None => break,
+                        Some(Err(err)) => {
                             if shard.online() {
                                 error!("error loading canonical type information: {err}");
                                 safe_sleep(Duration::from_millis(100)).await;
--- a/pgdog/src/backend/pool/cluster/schema_loader.rs
+++ b/pgdog/src/backend/pool/cluster/schema_loader.rs
@@ -28,7 +28,7 @@ impl SchemaLoader for FromServer {
 
         // For now we treat shard 0 as the canonical OID source
         if let Some(shard) = cluster.shards().first()
-            && let Some(canonical_oids) = dbg!(&cluster.canonical_oids)
+            && let Some(canonical_oids) = &cluster.canonical_oids
         {
             let canonical_oids = Arc::clone(canonical_oids);
             let shard = shard.clone();
--- a/pgdog/src/backend/pool/cluster/schema_loader.rs
+++ b/pgdog/src/backend/pool/cluster/schema_loader.rs
@@ -51,7 +51,7 @@ impl SchemaLoader for FromServer {
                         Err(err) => {
                             if shard.online() {
                                 error!("error loading canonical type information: {err}");
-                                sleep(Duration::from_millis(100)).await;
+                                safe_sleep(Duration::from_millis(100)).await;
                             } else {
                                 // Cluster is shutting down
                                 break;
--- a/pgdog/src/backend/pool/cluster/schema_loader.rs
+++ b/pgdog/src/backend/pool/cluster/schema_loader.rs
@@ -3,6 +3,7 @@ use crate::backend::pool::ee::schema_changed_hook;
 use crate::tasks;
 use crate::util::safe_sleep;
 use dyn_clone::DynClone;
+use std::sync::Arc;
 use std::time::Duration;
 use tokio::select;
 use tracing::error;
@@ -23,6 +24,42 @@ impl SchemaLoader for FromServer {
                 shard.schema_not_needed();
             }
             return;
+        }
+
+        // For now we treat shard 0 as the canonical OID source
+        if let Some(shard) = cluster.shards().first() {
+            let canonical_oids = Arc::clone(&cluster.canonical_oids);
+            let shard = shard.clone();
+            tasks::spawn("load canonical oids", async move {
+                // FIXME: This shutdown signal/retry if error logic seems like
+                // it might be worth abstracting
+                let shutdown = tasks::shutdown_signal();
+
+                loop {
+                    let loader = async {
+                        canonical_oids
+                            .load(&mut *shard.primary_or_replica(&Default::default()).await?)
+                            .await
+                    };
+                    let result = select! {
+                        _ = shutdown.cancelled() => break,
+                        result = loader => { result },
+                    };
+
+                    match result {
+                        Ok(_) => break,
+                        Err(err) => {
+                            if shard.online() {
+                                error!("error loading canonical type information: {err}");
+                                sleep(Duration::from_millis(100)).await;
+                            } else {
+                                // Cluster is shutting down
+                                break;
+                            }
+                        }
+                    }
+                }
+            });
         }
 
         for shard in cluster.shards() {
--- a/pgdog/src/backend/pool/cluster/schema_loader.rs
+++ b/pgdog/src/backend/pool/cluster/schema_loader.rs
@@ -27,8 +27,10 @@ impl SchemaLoader for FromServer {
         }
 
         // For now we treat shard 0 as the canonical OID source
-        if let Some(shard) = cluster.shards().first() {
-            let canonical_oids = Arc::clone(&cluster.canonical_oids);
+        if let Some(shard) = cluster.shards().first()
+            && let Some(canonical_oids) = dbg!(&cluster.canonical_oids)
+        {
+            let canonical_oids = Arc::clone(canonical_oids);
             let shard = shard.clone();
             tasks::spawn("load canonical oids", async move {
                 loop {
@@ -54,6 +56,10 @@ impl SchemaLoader for FromServer {
                     }
                 }
             });
+        } else {
+            for shard in cluster.shards() {
+                shard.skip_loading_oids()
+            }
         }
 
         for shard in cluster.shards() {
--- a/pgdog/src/backend/pool/shard/oids.rs
+++ b/pgdog/src/backend/pool/shard/oids.rs
@@ -1,0 +1,132 @@
+use super::{Request, Shard};
+use crate::{
+    backend::{Error, Server},
+    net::DataRow,
+    sync::SetOnceCell,
+};
+use std::collections::HashMap;
+use std::sync::Arc;
+use tracing::info;
+
+#[derive(Debug)]
+/// The mapping from a shards type OID to a canonical one
+pub(crate) struct Oids {
+    canonical_oids: Arc<CanonicalOids>,
+    mappings: SetOnceCell<OidMappings>,
+}
+
+impl Oids {
+    pub(crate) fn new(canonical_oids: &Arc<CanonicalOids>) -> Arc<Self> {
+        Arc::new(Self {
+            canonical_oids: Arc::clone(canonical_oids),
+            mappings: Default::default(),
+        })
+    }
+
+    pub(crate) async fn load(&self, shard: &Shard) -> Result<&OidMappings, Error> {
+        self.mappings
+            .get_or_try_init(|| async {
+                let mut server = shard.primary_or_replica(&Request::default()).await?;
+                let oids = load_oids(&mut server).await?;
+                let canonical = self.canonical_oids.oids.wait().await;
+                let mut canonical_to_shard = HashMap::new();
+                let mut shard_to_canonical = HashMap::new();
+                for (type_name, oid) in oids {
+                    let canonical = canonical
+                        .get(&type_name)
+                        .copied()
+                        .ok_or(Error::MissingCanonicalOid(type_name))?;
+                    if canonical == oid {
+                        continue;
+                    }
+                    canonical_to_shard.insert(canonical, oid);
+                    shard_to_canonical.insert(oid, canonical);
+                }
+
+                debug_assert_eq!(canonical_to_shard.len(), shard_to_canonical.len());
+                info!(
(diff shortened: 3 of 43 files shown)


sgrif:
This is now a working implementation. It needs some cleanup and backfill of unit tests, but should be able to make it into tomorrow's release

pgdog/src/sync.rs
```diff
@@ -0,0 +1,85 @@
+use std::fmt;
+use tokio::sync::{Semaphore, SemaphorePermit, SetError, SetOnce, TryAcquireError};
+
+/// A combination of `SetOnce` and `OnceCell`.
+/// It allows only a single writer to run, with no contention on future reads,
+/// while also allowing callers to wait on the value.
+pub(crate) struct SetOnceCell<T> {
```
sgrif, line 7: Prior art: https://github.com/tokio-rs/tokio/issues/4788 https://doc.rust-lang.org/stable/std/sync/struct.OnceLock.html

(Likely going to try to upstream this to tokio since it was previously rejected as a new feature on `OnceCell`, but it was mentioned it may be accepted as a new type)

pgdog/src/backend/pool/shard/oids.rs
```diff
@@ -0,0 +1,111 @@
...
+impl Oids {
+    pub(crate) fn new(canonical_oids: &Arc<CanonicalOids>) -> Arc<Self> {
+        Arc::new(Self {
+            canonical_oids: Arc::clone(canonical_oids),
+            mappings: Default::default(),
+        })
+    }
+
+    pub(crate) async fn load(&self, shard: &Shard) -> Result<&OidMappings, Error> {
+        self.mappings
```
levkk, line 27: I'm not sure if I'm reading this right, but I think you may want the shard OIDs to be cached on the `Schema`, and once you have all shards loaded, mutate it to compute the shard 0 (canonical) vs. others mapping.

`Schema` is cached by `SchemaCache` which will be called once per (database, shard) pair. For deployments with 50+ users, this oid fetching will produce 100s of additional queries, which will block startup. Again, I'm not entirely sure this happens here because of the `get_or_try_init`, but it seems that `Oid` is attached to a `Shard`, which is created for each user in `users.toml`. That's where our thundering herd came from last time.

pgdog/src/backend/schema/cache/mod.rs
```diff
@@ -5,15 +5,20 @@ use dashmap::DashMap;
...
 
 /// Schema cache.
 #[derive(Debug, Default, Clone)]
 pub(crate) struct SchemaCache {
     // Database => shard => Schema
-    cache: Arc<DashMap<String, DashMap<usize, Entry>>>,
+    cache: Arc<DashMap<(String, usize), Entry>>,
+    /// The canonical mapping of type names to OID
+    /// A cluster's canonical mapping of type names to OID
+    canonical_oids: Arc<DashMap<String, Arc<CanonicalOids>>>,
```
levkk, line 19: Ah there they are. Must of missed them earlier.

pgdog/src/backend/pool/shard/oids.rs
sgrif: `Oids` is guaranteed to only be set exactly once, regardless of how many callers it gets. We're making sure that for any database/pair combo, everyone has a handle to the same `Oids` instance, so no additional synchronization or mutation is needed.

pgdog/src/net/messages/parse.rs
```diff
@@ -117,34 +112,44 @@ impl Parse {
...
+        for _ in 0..bytes.get_u16() {
+            let canonical_oid = bytes.get_u32();
+            if let Some(&shard_oid) = mapping.get(&canonical_oid) {
+                // Note: We can't use BufMut here, as that writes to the end of the
+                // buffer, which will allocate. We just want to write to the
+                // existing bytes.
+                let pos = bytes.position() as usize;
+                let oid_bytes = &mut bytes.get_mut()[(pos - 4)..pos];
+                oid_bytes.copy_from_slice(&shard_oid.to_be_bytes());
+                rewritten = true;
```
meskill, line 135: should we check that the value is actually different and not mark it as rewritten if they are the same?

pgdog/src/net/messages/parse.rs
```diff
@@ -117,34 +112,44 @@ impl Parse {
...
+            }
+        }
+        self.data_types = bytes.into_inner().freeze();
+        rewritten
```
meskill, line 140: we probably need to reset `Parse::original` if we did the rewrite. On to_bytes impl the `original` if present will be sent and our change will be ignored.

small fix, but the tests didn't catch this..

meskill (changes requested):

pgdog/src/net/messages/parse.rs
sgrif: The mapping only contains OIDs that are different. Since this is a hot path, I wanted to make sure we could bail early in the (presumably most common) case where no types differ

pgdog/src/backend/prepared_statements.rs
```diff
@@ -288,10 +295,7 @@ impl PreparedStatements {
 
             'T' => {
                 if let Some(describe) = self.describes.pop_front() {
-                    self.add_row_description(
-                        &describe,
-                        &RowDescription::from_bytes(message.to_bytes())?,
-                    );
+                    self.add_row_description(&describe, message)?;
```
meskill: the `rewrite_row_description_data_types` is side-effect of `self.add_row_description` but `self.add_row_description` is only called if the describes are present. And they are not present for Simple queries and maybe some other cases, I'm not sure. 

The rewrite should be not related to row_description

pgdog/src/backend/pool/shard/oids.rs
```diff
@@ -0,0 +1,132 @@
...
+            .await
+            .map(|_| ())
+    }
+}
+
+async fn load_oids(server: &mut Server) -> Result<impl Iterator<Item = (String, u32)>, Error> {
+    // OIDs < 10,000 are reserved for PG's internal use and are assumed to be stable
+    Ok(server
+        .fetch_all::<DataRow>(
+            "SELECT DISTINCT ON (typname) typname, pg_type.oid FROM pg_type WHERE pg_type.oid >= 10000",
```
meskill: but shouldn't we include the type namespace here to avoid typename clashes?

pgdog/src/backend/replication/logical/subscriber/tests.rs
```diff
@@ -1584,6 +1590,7 @@ async fn full_identity_nothing_rejected() {
 #[tokio::test]
 async fn full_identity_omni_no_unique_index_rejected() {
     let cluster = Cluster::new_test_single_shard(&config());
+    cluster.force_dummy_schema();
```
meskill: would it be simpler if we put `force_dummy_schema` inside the `new_test_*` and if don't need to use then run another function?

pgdog/src/backend/pool/cluster/schema_loader.rs
```diff
@@ -25,6 +26,42 @@ impl SchemaLoader for FromServer {
...
+                loop {
+                    let loader = async {
+                        canonical_oids
+                            .load(&mut *shard.primary_or_replica(&Default::default()).await?)
+                            .await
+                    };
+                    let result = select! {
+                        _ = shutdown.cancelled() => break,
+                        result = loader => { result },
+                    };
```
meskill: https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html#method.run_until_cancelled

that could be a little simpler

pgdog/src/backend/prepared_statements.rs
```diff
@@ -459,6 +472,36 @@ impl PreparedStatements {
...
+        let mut parameter_description = ParameterDescription::from_bytes(message.payload())?;
+        parameter_description.rewrite_data_types(&self.oid_mappings().shard_to_canonical);
+        message.replace_payload(parameter_description.to_bytes());
+        Ok(())
+    }
+
+    fn oid_mappings(&self) -> &OidMappings {
+        self.oids
+            .get()
+            .expect("cluster startup always loads OID information or sets an empty mapping")
```
meskill: can we guarantee this won't panic? or rather, can we allow ourselves to have panic here if the actual guarantee should be hold by another part of the code that could diverge?

pgdog/src/backend/pool/shard/oids.rs
```diff
@@ -0,0 +1,132 @@
...
+
+impl Oids {
+    pub(crate) fn new(canonical_oids: &Arc<CanonicalOids>) -> Arc<Self> {
+        Arc::new(Self {
+            canonical_oids: Arc::clone(canonical_oids),
+            mappings: Default::default(),
+        })
+    }
+
+    pub(crate) async fn load(&self, shard: &Shard) -> Result<&OidMappings, Error> {
```
meskill, line 26: I didn't find this part: do you ignore this load for the canonical shard (i.e. shard=0) somewhere?

sgrif: No. I don't want to encode the fact that we consider shard 0 to be the canonical source in more places than is strictly necessary, as we may end up having it be configurable or even decided at runtime. Since this is only a single query that gets run at cluster startup, I don't think the cost is a problem

pgdog/src/backend/prepared_statements.rs
sgrif: The problem is that these are loaded asynchronously, we won't have them at the point `PreparedStatements` is constructed. Since it's a field on `Server`, and we need a `Server` to load the OIDs in the first place, we need an `Option` somewhere. I opted to panic instead of `unwrap_or_default` or similar, since if this ever *does* panic it indicates a bug in our code (likely from a different part of the code base diverging, as you said), and so returning an empty mapping would just silently hide that bug

The only other option I can think of is to have this function be async, but that'd just replace the panic with a future that never resolves which seems even worse.

I'm open to suggestions though

pgdog/src/backend/prepared_statements.rs
sgrif: The reason we rewrite on add is because we need to ensure the cached version is in the canonical form. Attempting to do it any later would mean we wouldn't know what mappings to use.

As for rewriting in the simple protocol, I think you're right that this code path doesn't get hit. I need to find the right place to add some more unit test coverage for this.

pgdog/src/backend/pool/shard/oids.rs
sgrif: You're right. For some reason I dismissed this because there's no way to set the schema in the grammar using `CREATE TYPE`.

sgrif:
Between making sure we have adequate test coverage for `RowDescription` on the simple protocol, and that python test failure, this needs a bit more work.

pgdog/src/backend/replication/logical/subscriber/tests.rs
sgrif: I was hesitant to do that, since it'd mean that this:

```
let cluster = Cluster {
    schema_loader: Box::new(ManuallyLoad),
    ..Cluster::new_test_single_shard(&config())
};
cluster.launch()
```

Would no longer the caller simulate schema loading. Right now the only caller setting `schema_loader` manually is calling `new_test`, but I felt it'd be weird to have `new_test` not force the dummy schema but `new_test_*` force it. The logical replication tests are the only code that doesn't launch the cluster after creating it, so these tests were the only place that needed it. Any thoughts?

integration/rust/Cargo.toml
```diff
@@ -6,6 +6,9 @@ edition = "2024"
 [lib]
 test = true
 
+[features]
+new_parser = []
```
levkk, line 10: CI sets this flag I think, so you may want to not set it here since the "legacy" tests will probably fail. You may need to pass the flag down here: https://github.com/pgdogdev/pgdog/blob/ff676a0e295d15e77159a0620325b4dacad44d28/integration/rust/dev.sh#L6

I can add this as a follow up / PR for your PR, since it involves a lot of glue.

levkk (approved):
LGTM!

Might be worth considering using `fnv` (already in `Cargo.toml`) instead of hashbrown for mappings. It's faster and also we don't need dos protection since oids are database-generated.

Bonus points: include the code our friends gave us as repro into our Ruby test suite to make sure it works as expected and doesn't break in the future.

integration/rust/Cargo.toml
sgrif: This isn't setting a flag, it's declaring it as something that can be set.

levkk: Oh right!

pgdog/src/backend/pool/cluster/schema_loader.rs
```diff
@@ -27,28 +27,24 @@ impl SchemaLoader for FromServer {
         }
 
         // For now we treat shard 0 as the canonical OID source
-        if let Some(shard) = cluster.shards().first() {
-            let canonical_oids = Arc::clone(&cluster.canonical_oids);
+        if let Some(shard) = cluster.shards().first()
+            && let Some(canonical_oids) = dbg!(&cluster.canonical_oids)
```
meskill: `dbg!` leftovers

pgdog/src/backend/pool/shard/oids.rs
meskill: yeah, totally. 

but I have concerns about possible dead-locks below for the canonical shard: `let mut server = shard.primary_or_replica(&Request::default()).await?;` will hold the guard i.e. one connection from pool. 

yeah, it's probably an issue only for 1-sized pool 😄 but might it worth a drop(server) before wait?

meskill (approved):

pgdog/src/backend/pool/shard/oids.rs
sgrif: I don't think that's a concern since we never attempt to load the oids until after we've started loading the canonical ones, but sure, I'm happy to make sure we drop the server asap

sgrif: Fixed to let them run concurrently as well

sgrif:
>  include the code our friends gave us as repro into our Ruby test suite to make sure it works as expected and doesn't break in the future.

The integration tests written are a distillation of their repro script, so it's mostly redundant. I can follow-up with a regression test for Sequel specifically though

PR #1282, opened 2026-07-29
fix(memory): prepared statements cache retains gigabytes after statement spike

IgorOhrimenko:
## Problem

Long-lived clients that keep preparing unique SQL statements (report jobs, ORMs interpolating values into SQL) grow the global prepared statements cache to millions of entries. When they disconnect, the maintenance sweep correctly trims entries down to `prepared_statements_limit` — but the process keeps holding gigabytes of RSS indefinitely:

1. `HashMap`/`HashSet` never return capacity to the allocator. After a spike of 10M statements, the three tables in `GlobalCache` (`statements`, `names`, `unused`) keep ~13.5M slots (~5.9 GiB) at 100 live entries, forever.
2. `prepared_statements_memory_used` is blind to this: `MemoryUsage for HashMap` sums live entries via `iter()`, so it reports ~44 KiB while RSS holds ~10 GiB.

Reproduction (docker compose + load generator): https://gist.github.com/IgorOhrimenko/8ba71664c7cf27f005e9ef80a48cf97c

Measured on a synthetic load test (10M unique prepared statements at ~7k/s through one PgDog instance, then all clients disconnect; all values settled):

| build | peak RSS | `memory_used` after disconnect | `capacity` after disconnect | RSS after disconnect |
|---|---|---|---|---|
| v0.1.50 | 15.5 GiB | 44 KiB | — | 10.1 GiB, held |
| + accounting fix | 15.5 GiB | 5.91 GiB (real) | 13 478 814 | 9.9 GiB, held |
| + shrink fix | 15.5 GiB | 148 KiB (real — memory returned) | 224 | 2.5 GiB |

## Changes

**`fix(stats)`: count hash table capacity in memory accounting**
- `MemoryUsage for HashMap`/`HashSet` now counts allocated slots (plus control bytes) and is generic over the hasher, so `FnvHashSet` is covered.
- New `prepared_statements_capacity` gauge: `capacity` far above `prepared_statements` signals memory held from a past spike, which makes this failure mode visible on a dashboard.

**`fix(memory)`: shrink the tables in the maintenance sweep**
- `close_unused()` calls `shrink_to(len * 2)` on all three tables once they are mostly empty (`capacity > 8 * len`) and large enough to matter (`> 4096` slots). Hysteresis avoids rehashing on every sweep; with jemalloc's `background_thread` (default since #1260) the freed pages are returned to the OS within seconds.
- `reset()` also releases capacity.

Peak memory under load is unchanged — entries held by connected clients are legitimate cache content; this PR is about returning the memory after they are gone.

## Testing

- Unit tests for the capacity accounting, the shrink, and the hysteresis guards (no shrink for small or mostly-full tables).
- `cargo nextest run --test-threads=1` over the workspace packages against a local Postgres — green.
- Verified end-to-end with the load test above.

### DIFF
--- a/pgdog/src/frontend/prepared_statements/global_cache.rs
+++ b/pgdog/src/frontend/prepared_statements/global_cache.rs
@@ -239,6 +239,8 @@ impl GlobalCache {
         // unused will hold the remaining elements that was not extracted above
         self.unused = unused;
 
+        self.maybe_shrink();
+
         removed
     }
 
@@ -250,6 +252,23 @@ impl GlobalCache {
     /// Get all prepared statements in the global cache, keyed by global unique key.
     pub(crate) fn statements(&self) -> &HashMap<CacheKey, CachedStmt> {
         &self.statements
+    }
+
+    /// Return table memory to the allocator after a spike of unique
+    /// statements. Hysteresis (mostly-empty table, above a minimum size)
+    /// avoids rehashing on every sweep; shrinking to twice the current
+    /// len leaves headroom for new statements.
+    fn maybe_shrink(&mut self) {
+        const SHRINK_FACTOR: usize = 8;
+        const MIN_CAPACITY: usize = 4096;
+
+        if self.statements.capacity() > MIN_CAPACITY
+            && self.statements.capacity() / SHRINK_FACTOR > self.statements.len()
+        {
+            self.statements.shrink_to(self.statements.len() * 2);
+            self.names.shrink_to(self.names.len() * 2);
+            self.unused.shrink_to(self.unused.len() * 2);
+        }
     }
 
     /// Remove statement from global cache.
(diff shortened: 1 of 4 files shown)


pgdog/src/stats/memory.rs
```diff
@@ -3,6 +3,10 @@ use lru::LruCache;
 use std::collections::{BTreeMap, HashMap, HashSet, VecDeque};
 use std::hash::Hash;
 
+/// Approximate bytes attributable to a value, for metrics.
```
IgorOhrimenko, line 6: A note on the accounting precision, since it is not obvious from the diff:

`MemoryUsage` impls are mixed — scalars and small structs (`usize`, `CacheKey`, `CachedStmt`) report their inline size, while `String`/`BytesMut` report heap capacity. With the new capacity term in the `HashMap`/`HashSet` impls this means the inline portion of *live* entries is counted twice: once inside the table's capacity, once in the per-element sum.

The overcount is bounded by `len * inline_size`, which matters only when the table is full (a few hundred MB at millions of entries, against multi-GiB actual usage) and vanishes in the post-spike state this metric is meant to catch (few live entries, huge capacity). Making it exact would require splitting the trait into inline/heap halves across every impl, which does not seem worth it for an observability gauge — the doc comment on the trait now states the upper-bound semantics.

pgdog/src/frontend/prepared_statements/global_cache.rs
```diff
@@ -124,7 +124,7 @@ impl MemoryUsage for GlobalCache {
             + self.names.memory_usage()
             + self.counter.memory_usage()
             + self.versions.memory_usage()
-            + self.unused.len() * std::mem::size_of::<usize>()
+            + self.unused.memory_usage()
```
levkk: I remember removing this because it was very expensive to calculate (O(n)).

pgdog/src/frontend/prepared_statements/global_cache.rs
```diff
@@ -326,9 +335,28 @@ impl GlobalCache {
...
+        self.maybe_shrink();
+
         remove.len()
     }
 
+    /// Return table memory to the allocator after a spike of unique
+    /// statements. Hysteresis (mostly-empty table, above a minimum size)
+    /// avoids rehashing on every sweep; shrinking to twice the current
+    /// len leaves headroom for new statements.
+    fn maybe_shrink(&mut self) {
```
levkk, line 276: Why not call `shrink_to_fit` instead? https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.shrink_to_fit

pgdog/src/stats/memory.rs
```diff
@@ -54,12 +58,17 @@ impl<V: MemoryUsage> MemoryUsage for Vec<V> {
...
+impl<K: MemoryUsage, V: MemoryUsage, S> MemoryUsage for HashMap<K, V, S> {
     #[inline(always)]
     fn memory_usage(&self) -> usize {
-        self.iter()
-            .map(|(k, v)| k.memory_usage() + v.memory_usage())
-            .sum::<usize>()
+        // The table allocates capacity() slots (plus one control byte each),
+        // not len(): spare capacity left behind by removed entries still
+        // occupies memory and has to be counted.
+        self.capacity() * (std::mem::size_of::<(K, V)>() + 1)
```
levkk, line 67: This is O(n). We should be careful not to call this frequently. I had to debug CPU usage issues with this before.

pgdog/src/frontend/prepared_statements/global_cache.rs
IgorOhrimenko: You're right, restored O(1) in d0e0a503 — `GlobalCache::memory_usage()` no longer iterates anything: it reports table capacities plus a `content_bytes` counter that is maintained incrementally on insert/remove/rewrite/describe. The `unused` set isn't touched at all.

pgdog/src/frontend/prepared_statements/global_cache.rs
IgorOhrimenko: Done, switched to `shrink_to_fit` in d0e0a503. The hysteresis guards stay (only shrink tables over 4096 slots that are less than 1/8 full), so the once-a-second sweep never rehashes in steady state — the shrink fires once after a spike drains.

pgdog/src/stats/memory.rs
IgorOhrimenko: Fair concern — the per-entry `iter()` sum actually predates this PR, but with millions of cached statements it did run on every metrics scrape. As of d0e0a503 the global cache no longer uses this impl at all: reporting is O(1) (capacities + incrementally tracked content bytes), and a test verifies the counter against a full recomputation across every mutation path. The generic impl here remains O(n) but its remaining call sites are bounded-size maps (per-server statements, capped by `prepared_statements_limit`). Measured on a synthetic run: scrape latency stays flat (~60-150 ms, dominated by pool metrics) from 100 to 227k cached statements.

levkk (approved):
Nice!

PR #1283, opened 2026-07-29
 feat(tls): allow password users to connect over TLS without a  client certificate

bhargavtheertham-cb:
## Problem

Setting `tls_client_ca_certificate` makes client certificates mandatory for every TLS connection, because rustls' default client verifier denies anonymous clients. On a listener serving both mTLS users (`identity`) and password users (`password`/`password_hash`), that leaves the password users unable to negotiate TLS at all — they get a `certificate_required` alert, so their only working option is a plaintext connection:

```
$ psql "host=... user=some_password_user sslmode=require"
psql: error: connection to server at "..." failed:
      SSL error: tlsv13 alert certificate required

$ psql "host=... user=some_password_user sslmode=disable"
psql (17.7)          # works, unencrypted
```

PgDog can't resolve this during the handshake, which completes before the startup packet carrying `user=` arrives — at that point it has no idea who is connecting. So the choice is currently all-or-nothing for every connection on the listener.

Worth noting the current requirement doesn't actually protect those users: they're already reachable unauthenticated over plaintext. It only forces the legitimate ones onto the unencrypted path.

## Change

Build the verifier with `allow_unauthenticated()` so a certificate is verified when presented but its absence no longer fails the handshake, and move the requirement to authentication, where the user is known.

A new per-user `tls_client_certificate_required` defaults to `true` and is enforced only when a client CA is configured **and** the client connected over TLS:

- without a client CA, no certificate is ever requested (`with_no_client_auth`), so requiring one could never be satisfied — enforcing it unconditionally would reject every user in every deployment that doesn't configure client CAs;
- plaintext connections stay governed by `tls_client_required`.

```toml
[[users]]
name = "app"
database = "db"
password_hash = "SCRAM-SHA-256$4096:..."
tls_client_certificate_required = false   # TLS, no client cert
```

## Backwards compatibility

Behavior is unchanged at the default:

| Client | Before | After |
|---|---|---|
| TLS, no cert, client CA set | rejected (TLS alert) | rejected (`NoClientCertificate`) |
| TLS, no cert, **opted out** | rejected | **authenticates over TLS** |
| TLS, valid cert | verified | verified, unchanged |
| TLS, untrusted cert | rejected at handshake | rejected at handshake |
| Plaintext | governed by `tls_client_required` | unchanged |
| No client CA configured | cert never requested | unchanged |

The one visible difference at default is that a certless client is now rejected after the handshake rather than during it, so the error is a Postgres auth failure instead of a TLS alert. That also means slightly more work per rejected connection — happy to gate the verifier change behind a config flag instead if you'd prefer to keep the early rejection.

## mTLS is unaffected

Identity enforcement never lived in the handshake — `client/mod.rs` already requires `stream.tls_identity() == Some(identity)` for any user with a configured identity, and a certless client gets `None`, so it still fails with `NoIdentity`. An entry with neither identity nor password still fails closed via the existing `NoPasswordConfig` guard. Certificates that *are* presented are verified against the CA bundle exactly as before; only the empty-certificate case changes.

 
## Testing

- `cargo fmt --all -- --check` and `cargo clippy --all-targets -- -D warnings` clean
- `cargo test -p pgdog-config` — 83 passed
- New: `client_cert_verifier_allows_anonymous_clients` asserts `offer_client_auth()` stays true while `client_auth_mandatory()` becomes false; two config tests cover the unset and opted-out cases

### DIFF
--- a/pgdog/src/frontend/client/mod.rs
+++ b/pgdog/src/frontend/client/mod.rs
@@ -95,18 +95,27 @@ pub struct Client {
     query_size_limit: Option<usize>,
 }
 
-/// Reject a client that owed us a certificate and didn't send one.
-///
-/// Only meaningful when a client CA is configured, since otherwise no certificate
-/// is ever requested, and only over TLS, since plaintext is governed by
-/// `tls_client_required`.
-fn missing_required_client_certificate(
+/// Inputs to the per-user client certificate check.
+struct ClientCertificateCheck {
+    /// A client CA is configured, so certificates are requested at all.
     client_ca_configured: bool,
+    /// The client connected over TLS.
     is_tls: bool,
+    /// This user's `tls_client_certificate_required`.
     required: bool,
+    /// The client presented a certificate during the handshake.
     presented: bool,
-) -> bool {
-    client_ca_configured && is_tls && required && !presented
+}
+
+impl ClientCertificateCheck {
+    /// The client owed us a certificate and didn't send one.
+    ///
+    /// Only meaningful when a client CA is configured, since otherwise no
+    /// certificate is ever requested, and only over TLS, since plaintext is
+    /// governed by `tls_client_required`.
+    fn rejected(&self) -> bool {
+        self.client_ca_configured && self.is_tls && self.required && !self.presented
+    }
 }
 
 impl Client {
@@ -296,12 +305,14 @@ impl Client {
                         } else {
                             AuthResult::NoIdentity
                         }
-                    } else if missing_required_client_certificate(
+                    } else if (ClientCertificateCheck {
                         client_ca_configured,
-                        stream.is_tls(),
-                        cluster.tls_client_certificate_required(),
-                        stream.tls_client_certificate(),
-                    ) {
+                        is_tls: stream.is_tls(),
+                        required: cluster.tls_client_certificate_required(),
+                        presented: stream.tls_client_certificate(),
+                    })
+                    .rejected()
+                    {
                         // Asked for a certificate and declined. Users that opt out
                         // fall through to password authentication instead.
                         AuthResult::NoClientCertificate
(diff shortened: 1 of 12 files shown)


pgdog/src/frontend/client/mod.rs
```diff
@@ -95,6 +95,20 @@ pub struct Client {
     query_size_limit: Option<usize>,
 }
 
+/// Reject a client that owed us a certificate and didn't send one.
+///
+/// Only meaningful when a client CA is configured, since otherwise no certificate
+/// is ever requested, and only over TLS, since plaintext is governed by
+/// `tls_client_required`.
+fn missing_required_client_certificate(
```
levkk: nit: you're better off passing in the original structs by reference, that way you can be sure the right bools are in the right place.

levkk: Alternatively, you can make it a struct, so it's less likely to make a typo:

```rust
struct MissingParams {
  client_ca_configured: bool,
  is_tls: bool,
  // ...
}
```

levkk (approved):
LGTM! Happy to merge as-is, but let me know if you'd want to refactor that 4 boolean method.

bhargavtheertham-cb:
Took the struct suggestion — thanks, it was a real hole rather than just a style point.

Because the body is a flat `&&` and only `presented` is negated, most transpositions were harmless, but swapping `required` and `presented` inverted the check: a user who owed a certificate and sent none evaluated to `false` and would have been let through. Silently, and only in that one combination.

```rust
} else if (ClientCertificateCheck {
    client_ca_configured,
    is_tls: stream.is_tls(),
    required: cluster.tls_client_certificate_required(),
    presented: stream.tls_client_certificate(),
})
.rejected()
{
```

I went with the params struct over passing `&General`/`&Stream`/`&Cluster` for one reason: constructing a real `Stream` needs a live TLS handshake, so that version couldn't be unit tested and the truth table would have had to go. The struct keeps it trivially constructible.

The test also reads better now — it varies one field at a time from a rejected baseline, so each case names the condition it's exercising instead of relying on a `// client CA, is_tls, required, presented` comment above four bare booleans.

levkk (approved):

Issue #1264, opened 2026-07-24
[Metrics] Connection pinning metric for advisory locks


levkk:
Add connection pinning metric for advisory locks to OpenMetrics/OTEL. This should just count how many clients are `locked` (see `SHOW CLIENTS`).

PR #1286, opened 2026-07-30
Add ClientsLocked OTEL metric to track count of clients locked to a specific server

KennanHunter:
Implements #1264

I used this issue to get familiar with the codebase. Happy to rework any of it to fit how you want the stats module structured.

# Summary

This PR adds an OTEL metric called ClientsLocked that records the current number of locked clients.

On naming: I kept “locked” to match the internal and SHOW CLIENTS verbiage, but I’d actually argue for using the word “pinned”. Pinning describes the pooler-side behavior: a server connection bound to a specific client, excluded from multiplexing. Meanwhile, “locked” names one client-side cause and could be misread as row-level locking. Happy to do that rename separately if you agree. I also wonder whether there will be other scenarios where pinning connections are useful beyond locking.

# Validation

My friend Claude and I wrote a script to query and validate that the metric updates. You can view it in [this commit.](https://github.com/KennanHunter/pgdog/commit/cfd5ce9559b38776d135e136fdd841a2b2158239)

Here’s the script output:

```sh
✔ Container pgdog-pgdog-1   Started                                                                                    0.1s
 ✔ Container pgdog-shard_0-1 Started                                                                                    0.2s
 ✔ Container pgdog-shard_1-1 Started                                                                                    0.1s
 ✔ Container pgdog-shard_2-1 Started                                                                                    0.1s
== waiting for pgdog on :6432 ==
== waiting for /metrics on :9090 ==
== baseline ==
  clients_locked = 0
== opening persistent psql session via FIFO ==
== acquiring advisory lock ==
== polling for clients_locked = 1 ==
  clients_locked = 1
== releasing advisory lock on same connection ==
== polling for clients_locked = 0 ==
  clients_locked = 0
== closing psql session ==
PASS
```

### DIFF
--- a/pgdog/src/backend/pool/connection/binding.rs
+++ b/pgdog/src/backend/pool/connection/binding.rs
@@ -13,7 +13,6 @@ use crate::{
 };
 
 use futures::future::join_all;
-use tracing::warn;
 
 use super::*;
 
--- a/pgdog/src/backend/pool/connection/binding.rs
+++ b/pgdog/src/backend/pool/connection/binding.rs
@@ -13,6 +13,7 @@ use crate::{
 };
 
 use futures::future::join_all;
+use tracing::warn;
 
 use super::*;
 
@@ -486,6 +487,44 @@ impl Binding {
                 servers.iter_mut().for_each(|s| s.mark_dirty(true))
             }
             _ => (),
+        }
+    }
+
+    /// Propagate the client's lock state to every held Guard so each pool's
+    /// `sv_locked` reflects the pin.
+    pub(super) fn set_locked(&mut self, locked: bool) {
+        match self {
+            Binding::Direct(server, ..) => server.set_locked(locked),
+            Binding::MultiShard(servers, _) => {
+                for server in servers {
+                    server.set_locked(locked);
+                }
+            }
+            _ => (),
+        }
+    }
+
+    /// Aggregate lock state across the held Guard(s). All shards in a
+    /// multi-shard binding are set/cleared together via [`Self::set_locked`],
+    /// so they should always agree; if they don't, warn and err on the side
+    /// of "locked" so we don't recycle a pinned connection.
+    pub(super) fn is_locked(&self) -> bool {
+        match self {
+            Binding::Direct(server, ..) => server.is_locked(),
+            Binding::MultiShard(servers, _) => {
+                let mut any = false;
+                let mut all = true;
+                for server in servers {
+                    let l = server.is_locked();
+                    any |= l;
+                    all &= l;
+                }
+                if any && !all {
+                    warn!("inconsistent lock state across multi-shard binding; assuming locked");
+                }
+                any
+            }
+            _ => false,
         }
     }
 
--- a/pgdog/src/backend/pool/connection/binding.rs
+++ b/pgdog/src/backend/pool/connection/binding.rs
@@ -512,17 +512,12 @@ impl Binding {
         match self {
             Binding::Direct(server, ..) => server.is_locked(),
             Binding::MultiShard(servers, _) => {
-                let mut any = false;
-                let mut all = true;
-                for server in servers {
-                    let l = server.is_locked();
-                    any |= l;
-                    all &= l;
-                }
-                if any && !all {
-                    warn!("inconsistent lock state across multi-shard binding; assuming locked");
-                }
-                any
+                debug_assert!(
+                    servers.iter().all(|s| s.is_locked()) == servers.iter().any(|s| s.is_locked()),
+                    "Shards disagree on lock status {servers:?}"
+                );
+
+                servers.iter().any(|s| s.is_locked())
             }
             _ => false,
         }
--- a/pgdog/src/backend/pool/guard.rs
+++ b/pgdog/src/backend/pool/guard.rs
@@ -17,6 +17,10 @@ pub struct Guard {
     server: Option<Box<Server>>,
     pub(super) pool: Pool,
     pub(super) reset: bool,
+
+    /// Frontend has pinned this guard to its client
+    // NOTE: We cache this value seperately to prevent lock contention to read locked state
+    locked: bool,
 }
 
 impl std::fmt::Debug for Guard {
@@ -43,7 +47,23 @@ impl Guard {
             server: Some(server),
             pool,
             reset: false,
+            locked: false,
         }
+    }
+
+    /// Mark or unmark this checkout as pinned to its client. Propagates to
+    /// the pool so `sv_locked` reflects reality per pool.
+    pub(crate) fn set_locked(&mut self, locked: bool) {
+        self.locked = locked;
+        if let Some(server) = self.server.as_deref() {
+            self.pool.set_locked(server.id(), locked);
+        }
+    }
+
+    /// Whether this guard is currently pinned to its client.
+    #[inline]
+    pub(crate) fn is_locked(&self) -> bool {
+        self.locked
     }
 
     /// Rollback any unfinished transactions and check the connection
--- a/pgdog/src/backend/pool/guard.rs
+++ b/pgdog/src/backend/pool/guard.rs
@@ -19,7 +19,7 @@ pub struct Guard {
     pub(super) reset: bool,
 
     /// Frontend has pinned this guard to its client
-    // NOTE: We cache this value seperately to prevent lock contention to read locked state
+    // NOTE: We cache this value seperately to minimize pool lock contention
     locked: bool,
 }
 
@@ -54,6 +54,11 @@ impl Guard {
     /// Mark or unmark this checkout as pinned to its client. Propagates to
     /// the pool so `sv_locked` reflects reality per pool.
     pub(crate) fn set_locked(&mut self, locked: bool) {
+        if self.locked == locked {
+            // No-op, don't bother aquiring a lock
+            return;
+        }
+
         self.locked = locked;
         if let Some(server) = self.server.as_deref() {
             self.pool.set_locked(server.id(), locked);
--- a/pgdog-stats/src/pool.rs
+++ b/pgdog-stats/src/pool.rs
@@ -254,6 +254,8 @@ pub struct State {
     pub out_of_sync: usize,
     /// Re-synced servers.
     pub re_synced: usize,
+    /// Locked
+    pub locked: usize,
     /// Statistics
     pub stats: Stats,
     /// Max wait.
--- a/pgdog/src/backend/pool/connection/mod.rs
+++ b/pgdog/src/backend/pool/connection/mod.rs
@@ -52,7 +52,6 @@ pub struct Connection {
     binding: Binding,
     cluster: Option<Cluster>,
     mirrors: Vec<MirrorHandler>,
-    locked: bool,
     pub_sub: PubSubClient,
 }
 
@@ -69,7 +68,6 @@ impl Connection {
             user: user.to_owned(),
             database: database.to_owned(),
             mirrors: vec![],
-            locked: false,
             pub_sub: PubSubClient::new(),
         };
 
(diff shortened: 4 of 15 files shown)


levkk:
Hey! I think this is good. I'm wondering though, maybe we should expose that metric on each pool as `sv_locked` as well. While knowing how many clients are pinned is good, what is really helpful is knowing how many server connections are blocked from being re-used by transaction pooling. This can be then compared to the total number of connections in the pool to calculate pinned percentage, for example.

You can't really do the same calculation with `clients_locked` only because you don't know which pool they belong too. Wdyt?

KennanHunter:
Just took some time to read through more of the connection pooling code. Doing per pool makes a lot of sense.

Two ways to wire locked status to our existing pool metrics. I lean toward the second.

The quickest is to add `locked: usize` to `pgdog_stats::State`, then when the frontend locks, we increment/decrement inside `Connection.lock()` (`backend/pool/connection/mod.rs:433`) for each Guard in the Binding, then export it in `stats/pools.rs` alongside `sv_idle`/`sv_active`.

That changes the least amount of code but would work, the main concern is making sure that it's kept accurate every time there's a connection dropoff.

A refactor I like is to move the flag to the checkout itself, in `Checkout` (`backend/pool/taken.rs:10`):

```rust
struct Checkout {
    backend: BackendPid,
    key: BackendKeyData,
    locked: bool,
}
```

`Connection.lock()` sets/clears the flag via the pool for each held Guard. On disconnect the existing lifecycle logic (`Taken::check_in()`) removes the entry entirely, and the state lives in one place. When we export we walk through the current Checkout map similar to counting the active amount of connections.

If you agree I think I'll try to implement the variable on Checkout tomorrow or Friday.

KennanHunter:
Implemented! We now export sv_locked per pool.

```sh
== waiting for pgdog on :6432 ==
== waiting for /metrics on :9090 ==
== baseline ==
  clients_locked = 0
  sum(sv_locked) = 0
== opening persistent psql session via FIFO ==
== acquiring advisory lock ==
== polling for clients_locked = 1 ==
  clients_locked = 1
== polling for sum(sv_locked) = 1 ==
  sum(sv_locked) = 1
  per-shard breakdown:
    sv_locked{user="postgres",database="postgres",host="shard_0",port="5432",shard="0",role="primary"} 0
    sv_locked{user="postgres",database="postgres",host="shard_1",port="5432",shard="1",role="primary"} 1
    sv_locked{user="postgres",database="postgres",host="shard_2",port="5432",shard="2",role="primary"} 0
== releasing advisory lock on same connection ==
== polling for clients_locked = 0 ==
  clients_locked = 0
== polling for sum(sv_locked) = 0 ==
  sum(sv_locked) = 0
== closing psql session ==
PASS
```

<details>
    <summary>Verification script for completeness</summary>
    
```sh
#!/usr/bin/env bash
# Verify that pgdog_clients_locked and sv_locked react to a held
# session-scoped advisory lock.
#
# Uses one long-lived psql session driven via a FIFO so we can:
#   1. Acquire the advisory lock
#   2. Poll the metrics while the session is still connected
#   3. Explicitly release the lock (not just kill the session)
#   4. Poll the metrics again on the same connection
#
# Assumes docker/pgdog.toml enables `openmetrics_port = 9090` and
# docker-compose.yml maps `9090:9090` on the pgdog service.

set -euo pipefail

: "${METRICS_PORT:=9090}"
: "${PGDOG_PORT:=6432}"
: "${LOCK_ID:=42}"

METRICS_URL="http://127.0.0.1:${METRICS_PORT}/metrics"
export PGPASSWORD=postgres
PSQL=(psql -h 127.0.0.1 -p "${PGDOG_PORT}" -U postgres -d postgres -qXA)

# fetch a top-level, label-less metric value (e.g. clients_locked)
fetch() {
  curl -sf "$METRICS_URL" | awk -v n="$1" '$1 == n { print $2 }'
}

# sum a labeled metric across all its series (e.g. sv_locked{...})
fetch_sum() {
  curl -sf "$METRICS_URL" \
    | awk -v n="$1" 'index($0, n"{") == 1 { sum += $NF } END { print (sum == "" ? 0 : sum) }'
}

wait_for_metric() {
  local name=$1 expected=$2 v=""
  for _ in $(seq 1 5); do
    v=$(fetch "$name")
    if [[ "$v" == "$expected" ]]; then
      echo "$v"
      return 0
    fi
    sleep 1
  done
  echo "$v"
  return 1
}

wait_for_sum() {
  local name=$1 expected=$2 v=""
  for _ in $(seq 1 5); do
    v=$(fetch_sum "$name")
    if [[ "$v" == "$expected" ]]; then
      echo "$v"
      return 0
    fi
    sleep 1
  done
  echo "$v"
  return 1
}

FIFO=""
PSQL_PID=""
cleanup() {
  exec 3>&- 2>/dev/null || true
  [[ -n "$PSQL_PID" ]] && kill "$PSQL_PID" 2>/dev/null || true
  [[ -n "$FIFO" && -e "$FIFO" ]] && rm -f "$FIFO"
}
trap cleanup EXIT

echo "== waiting for pgdog on :${PGDOG_PORT} =="
until "${PSQL[@]}" -c 'SELECT 1' >/dev/null 2>&1; do sleep 1; done

echo "== waiting for /metrics on :${METRICS_PORT} =="
until curl -sf "$METRICS_URL" >/dev/null; do sleep 1; done

echo "== baseline =="
baseline_cl=$(fetch clients_locked)
baseline_sv=$(fetch_sum sv_locked)
echo "  clients_locked = ${baseline_cl}"
echo "  sum(sv_locked) = ${baseline_sv}"
[[ "$baseline_cl" == "0" ]] || { echo "FAIL: expected baseline clients_locked 0"; exit 1; }
[[ "$baseline_sv" == "0" ]] || { echo "FAIL: expected baseline sum(sv_locked) 0"; exit 1; }

echo "== opening persistent psql session via FIFO =="
FIFO=$(mktemp -u -t pgdog_test.XXXXXX).fifo
mkfifo "$FIFO"
"${PSQL[@]}" < "$FIFO" > /dev/null &
PSQL_PID=$!
exec 3>"$FIFO"

send() { printf '%s\n' "$1" >&3; }

echo "== acquiring advisory lock =="
send "SELECT pg_advisory_lock(${LOCK_ID});"

echo "== polling for clients_locked = 1 =="
if during=$(wait_for_metric clients_locked 1); then
  echo "  clients_locked = ${during}"
else
  echo "FAIL: expected clients_locked=1 while lock held, last saw '${during}'"
  exit 1
fi

echo "== polling for sum(sv_locked) = 1 =="
if during_sv=$(wait_for_sum sv_locked 1); then
  echo "  sum(sv_locked) = ${during_sv}"
  echo "  per-shard breakdown:"
  curl -s "$METRICS_URL" | grep '^sv_locked{' | sed 's/^/    /'
else
  echo "FAIL: expected sum(sv_locked)=1 while lock held, last saw '${during_sv}'"
  echo "  per-shard breakdown:"
  curl -s "$METRICS_URL" | grep '^sv_locked{' | sed 's/^/    /'
  exit 1
fi

echo "== releasing advisory lock on same connection =="
send "SELECT pg_advisory_unlock(${LOCK_ID});"

echo "== polling for clients_locked = 0 =="
if after=$(wait_for_metric clients_locked 0); then
  echo "  clients_locked = ${after}"
else
  echo "FAIL: expected clients_locked=0 after unlock, last saw '${after}'"
  exit 1
fi

echo "== polling for sum(sv_locked) = 0 =="
if after_sv=$(wait_for_sum sv_locked 0); then
  echo "  sum(sv_locked) = ${after_sv}"
else
  echo "FAIL: expected sum(sv_locked)=0 after unlock, last saw '${after_sv}'"
  exit 1
fi

echo "== closing psql session =="
send "\\q"
exec 3>&-
wait "$PSQL_PID" 2>/dev/null || true
PSQL_PID=""

echo "PASS"
```
</details>

pgdog/src/backend/pool/connection/binding.rs
```diff
@@ -489,6 +490,44 @@ impl Binding {
...
+
+    /// Aggregate lock state across the held Guard(s). All shards in a
+    /// multi-shard binding are set/cleared together via [`Self::set_locked`],
+    /// so they should always agree; if they don't, warn and err on the side
+    /// of "locked" so we don't recycle a pinned connection.
+    pub(super) fn is_locked(&self) -> bool {
+        match self {
+            Binding::Direct(server, ..) => server.is_locked(),
+            Binding::MultiShard(servers, _) => {
+                let mut any = false;
```
levkk: Assuming you're using `set_locked` (above), I don't think you need to check for this. `set_locked` is sync, so it cannot be interrupted and leave the shards in an inconsistent state.

Additionally, if this happens, this is a pretty bad bug, so a warning won't do, you almost have to error out here and abort the transaction I think.

pgdog/src/backend/pool/guard.rs
```diff
@@ -43,9 +47,25 @@ impl Guard {
             server: Some(server),
             pool,
             reset: false,
+            locked: false,
+        }
+    }
+
+    /// Mark or unmark this checkout as pinned to its client. Propagates to
+    /// the pool so `sv_locked` reflects reality per pool.
+    pub(crate) fn set_locked(&mut self, locked: bool) {
```
levkk, line 56: I would just double check that we don't call this unconditionally, i.e., on each transaction, with locked=false. We basically only want to take the pool mutex iff we are about to change the state (mutexes are pretty expensive).

levkk:
This looks really good! Just a couple comments (maybe just to double check) and this should be good to go :rocket:

pgdog/src/backend/pool/connection/binding.rs
KennanHunter: You're definitely right that it's unreachable as of now. I'm thinking of doing a debug_assert! just in case things drift accidentally in the future. Probably overkill but I see you use debug_assert! here and there and keeps our invariant defined. Zero cost in production.

```rust
Binding::MultiShard(servers, _) => {
    debug_assert!(
        servers.iter().all(|s| s.is_locked()) == servers.iter().any(|s| s.is_locked()),
        "Shards disagree on lock status {servers:?}"
    );

    servers.iter().any(|s| s.is_locked())
}
```

pgdog/src/backend/pool/guard.rs
KennanHunter: Good catch. As is this gets called on every transaction. Since we already duplicate the lock state in the Guard to prevent having to lock on reads, we can use it to short circuit if we are going to change the state to the same value.

```rust
/// Mark or unmark this checkout as pinned to its client. Propagates to
/// the pool so `sv_locked` reflects reality per pool.
pub(crate) fn set_locked(&mut self, locked: bool) {
    if self.locked == locked {
        // No-op, don't bother aquiring a lock
        return;
    }

    self.locked = locked;
    if let Some(server) = self.server.as_deref() {
        self.pool.set_locked(server.id(), locked);
    }
}
```

KennanHunter:
Thanks for the tips! Lmk if there's anything else you'd like to see changed. Feel free to squash merge if that's cleaner.

pgdog/src/backend/pool/guard.rs
levkk: Perfect

levkk:
One last `cargo clippy --fix --all-targets && cargo fmt` and we're good to go.

KennanHunter:
My bad got sloppy, ran and pushed

levkk (approved):
Amazing, thank you!

KennanHunter:
Noticed an issue that's stats-only but I don't want to lose the thought.
 
`set_locked` is a bit broken after a pool move. The guard hangs onto the pool it was created from, so once `move_conns_to` has emptied that pool's `Taken`, any later set_locked call from that old guard does a `backend_to_frontend` lookup, finds nothing, and returns without doing anything. 

Doesn't affect routing or check-in at all, nothing looks at `Checkout.locked` except the `sv_locked` stat. So worst case the pinned count on the new pool is a little off after a reload until the old guards drop.

If we want to fix it, `Inner::set_locked` could just forward through self.moved like `maybe_check_in` already does.

Does seem like as Pool state gets more complicated, manually resolving these moved-pool semantics will only get more annoying, but I'm not sure what a better alternative would be.

PR #1288, opened 2026-07-30
Fix flaky test test_reload_connection_count_stable

sgrif:
Whether or not `RELOAD` needs to establish a new connection or not is going to be highly dependent on timing during startup, and may or may not actually be stable.

However, running `RELOAD` once after startup should give us a reasonable baseline of how many connections are actually needed to do schema loading after the fact, and subsequent `RELOAD`s should not change that number as long as there are no other clients

### DIFF
--- a/integration/rust/tests/integration/reload.rs
+++ b/integration/rust/tests/integration/reload.rs
@@ -37,10 +37,12 @@ async fn test_reload() {
 #[tokio::test]
 #[serial]
 async fn test_reload_connection_count_stable() {
-    sleep(Duration::from_secs(1)).await;
     let admin = admin_tokio().await;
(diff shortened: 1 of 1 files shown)


levkk (approved):
I think this is right? I'm not sure what `settled` is for. Are we waiting for pgdog to close some extra connections or something? I know the original intent behind the test was to make sure `RELOAD` keeps the existing connections in the pool, i.e., doesn't act like `RECONNECT`, but I don't know why we need to do that here 20 times and why we stop when `settled` is true.

sgrif:
I think the intent was to make sure it didn't check this mid-schema loading where some but not all of the connections needed have been checked out.

But I've never witnessed any of the 20 attempts have a different value.

PR #1290, opened 2026-07-30
fix: migrate to safe_sleep/safe_interval/safe_timeout

meskill:
Kind of fix for the https://github.com/pgdogdev/pgdog/issues/1017 and a hope that proper fix will land on tokio with https://github.com/tokio-rs/tokio/pull/8334

## Issue

The details are explained in the https://github.com/tokio-rs/tokio/pull/8334 but in short: the tokio internally uses a special wheel to represent different segments of time. The last segment (level) is wraps the timers higher than 2^30 ms (12 days). When we get close to this mark the new sleep/interval timers go to the top level that later processed when reaching the 2^30 deadline.

The bug resides in the top level implementation - into the top level all the timers that are higher than 2^30 goes to this level. And surprisingly we have such timers that are delayed in the long future https://github.com/pgdogdev/pgdog/blob/b86c26540423db1103a0a45fa84b2b1de4d6b7f3/pgdog-config/src/lib.rs#L49

this timer goes to the top level initially and new timers go to the top level after 12 days and boom - the MAX_DURATION is identified by tokio wheel as the next one, instead of our 333ms next interval. That's exactly because only the top level have special meaning and accepts all the delays more than the level itself (> 2^36).

That's a rare case I believe - without MAX_DURATION timer everything works because the next occupied slot will be any actual near timer. Also, this

https://github.com/pgdogdev/pgdog/blob/9b27309d37c690f9e280768575fc16c8ce05f258/pgdog/src/backend/pool/connection/binding.rs#L93

was not the issue, since the sleep checks for overflows and basically ignores this timer.

The issue were reproducing for me locally because the default `lsn_check_timeout` is using MAX_DURATION and without it was working fine.

## Reproduction

Two tokio fork branches that could be used to verify the bug:
- https://github.com/meskill/tokio/pull/1
- https://github.com/meskill/tokio/pull/2

1. Clone, go to branch, 
2. on pgdog update cargo.toml to read from local package for tokio
3. enable logs for maintenance tasks
4. read the description of the pr for setup
5. make sure any MAX_DURATION triggered
6. the maintenance should stuck after some seconds

If alt timers enabled it's better to use 1 worker to reproduce since the alt timers have distribute wheel that could not lead to hang reliably.

## Fix

If fixed by https://github.com/tokio-rs/tokio/pull/8334 if it'll be merged.

And the current one is to replace all the calls to sleep/interval/timeout to "safe" versions that ignore the triggers longer than 2^36 ms (2 years btw), so until we reach the 2 years of uptime, boys!

### DIFF
--- a/pgdog/src/util.rs
+++ b/pgdog/src/util.rs
@@ -2,11 +2,13 @@
 
 use chrono::{DateTime, Local, Utc};
 use once_cell::sync::Lazy;
-use pgdog_config::MAX_DURATION;
 use rand::{Rng, distr::Alphanumeric};
 #[cfg(feature = "new_parser")]
 use std::ops::ControlFlow;
-use std::{env, future::Future, num::ParseIntError, time::Duration};
+use std::panic::Location;
+use std::{env, future::Future, future::pending, num::ParseIntError, time::Duration};
+use tokio::time::Interval;
+use tracing::warn;
 
 use crate::net::Parameters; // 0.8
 
@@ -298,14 +300,20 @@ pub fn sanitize_log_sample(s: &str, limit: usize) -> String {
     truncate_utf8(s, limit).replace(|c: char| c.is_control(), " ")
 }
 
-/// Avoid calling [`tokio::time::timeout`] on [`Duration`] that lasts effectively forever.
+/// Longest deadline Tokio's timer wheel can address (`64^6` ms, a bit over
+/// two years). A longer timer eventually stops every other timer in the
+/// runtime from firing.
 ///
-/// The Tokio timers can run hot and, if we can, we should avoid that mutex.
-///
-/// We did switch to the experimental timer that uses thread-locals, so
-/// this shouldn't be a problem anymore, but I still would like to avoid
-/// calling into time runtime if we don't have to.
-///
+/// See <https://github.com/pgdogdev/pgdog/issues/1017> and
+/// <https://github.com/tokio-rs/tokio/pull/8334>.
+const MAX_TIMER_DURATION: Duration = Duration::from_millis(1 << 36);
+
+fn armable(duration: Option<Duration>) -> Option<Duration> {
+    duration.filter(|duration| *duration < MAX_TIMER_DURATION)
+}
+
+/// [`tokio::time::timeout`] that waits forever instead of arming a timer
+/// outside [`MAX_TIMER_DURATION`].
 pub(crate) async fn safe_timeout<F>(
     duration: Duration,
     future: F,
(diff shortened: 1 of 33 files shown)


pgdog/src/util.rs
```diff
@@ -298,25 +300,79 @@ pub fn sanitize_log_sample(s: &str, limit: usize) -> String {
...
+        None => Ok(future.await),
+    }
+}
+
+/// [`tokio::time::sleep`] that waits forever instead of arming a timer
+/// outside [`MAX_TIMER_DURATION`].
+pub(crate) async fn safe_sleep(duration: Duration) {
+    match armable(Some(duration)) {
+        Some(duration) => tokio::time::sleep(duration).await,
+        None => pending().await,
```
sgrif, line 335: Should this sleep for the maximum time we possibly can instead of returning a future that will never resolve?

sgrif: Actually now that I think about it, is there a reason we couldn't break it up into multiple `Duration`s and await them in sequence?

meskill: if you have reasons to delay something like more than 2 years in the future than yeah, maybe. But most of the cases the max value was used as sentinel to basically disable some functionality. And with safe_* I want to avoid big numbers if they occasionally were put in the config

sgrif: 👍 Makes sense

levkk:
Oops, let me fix the conflict here.

pgdog/src/util.rs
```diff
@@ -298,25 +300,79 @@ pub fn sanitize_log_sample(s: &str, limit: usize) -> String {
...
-///
-/// We did switch to the experimental timer that uses thread-locals, so
-/// this shouldn't be a problem anymore, but I still would like to avoid
-/// calling into time runtime if we don't have to.
+/// Longest deadline Tokio's timer wheel can address (`64^6` ms, a bit over
+/// two years). A longer timer eventually stops every other timer in the
+/// runtime from firing.
 ///
+/// See <https://github.com/pgdogdev/pgdog/issues/1017> and
+/// <https://github.com/tokio-rs/tokio/pull/8334>.
```
levkk, line 308: https://p.kagi.com/proxy/aeae72_9f4bfdfdc48c4548badb1d7f64705020~mv2.gif?c=LsHiRSPxhD29sXqLhdI0j7IaAnHg3QTXq4ITFuYSEQbPvoFWx6zj9ILKxFBlVTxD8myWQKRMYmc8QzSFbAFxTY5-Bq1MlvrUFzaCr7OZjrLNPHClJRXxK71aw6jZy2Br

levkk (approved):
image

Image

Issue #1289, opened 2026-07-30
[Metrics] Piping OTEL metrics directly to Prometheus causes two errors


KennanHunter:
**PgDog version**
`c4f63fd6` latest main

**Description**
I was testing some OTEL code locally and pointed pgdog straight at Prometheus's native OTLP receiver (`--web.enable-otlp-receiver`). Hit two errors.

Pgdog's OTEL exporter looks like it's built for Datadog, and Prometheus users are supposed to hit the OpenMetrics `/metrics` endpoint instead. Both Datadog and the OpenTelemetry Collector tolerate these errors, so this only surfaces if you point OTLP at Prometheus directly. I'm filing anyway because the failure was confusing to me, the fixes are pretty simple, and selfishly I'm a big Prometheus fan.

**1. DELTA temporality on counters is rejected by Prometheus**

pgdog hardcodes `aggregation_temporality: 1` (DELTA) for counters at [`pgdog/src/stats/otel.rs:285`](./pgdog/src/stats/otel.rs). Prometheus's OTLP receiver is CUMULATIVE-only by default and rejects the whole batch:

```
invalid temporality and type combination for metric "pgdog.total_server_errors"
```

Because Prometheus 400s the entire request, every gauge sharing that batch with a DELTA counter gets dropped with it.

There's a workaround on Prometheus's side: `--enable-feature=otlp-deltatocumulative`, but it's nice if users don't have to. We already calculate the delta using an accumulated value, so if we just return the accumulated value it should be a pretty fast fix. 

The best fix is to make us match the exporter spec and switch between delta/cumulative based on environment variable as described in [the spec](https://opentelemetry.io/docs/specs/otel/metrics/sdk_exporters/otlp/)

> ...MUST set temporality preference to Cumulative for all instrument kinds by default.

> The exporter MUST configure the default aggregation temporality on the basis of instrument kind using the OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE variable as described below.

**2. `target_info` out-of-order writes from parallel batching**

```
Out of order sample from remote write ... series="{__name__=\"target_info\", host_name=\"...\", instance=\"...\", job=\"pgdog\"}"
```

The push loop in [`pgdog/src/stats/otel_exporter.rs`](./pgdog/src/stats/otel_exporter.rs) splits metrics into batches of 10 and fires them concurrently with `futures::future::join_all`. Each batch calls `otel::build_request()`, which grabs its own `now_nanos()`.

[...]

PR #1292, opened 2026-07-30
fix: use one timestamp for OTEL metric batches

KennanHunter:
Quick fix to prevent batch chunks having out of order timestamps, which messes up certain strict OTEL ingest pipelines like described in #1289.

This makes every chunk have the same timestamp, so if duplicate data is sent in different chunks, particularly metadata such as [target_info](https://prometheus.io/docs/guides/opentelemetry/#including-resource-attributes-at-query-time), it dedupes cleanly.


levkk (approved):
That's an awesome find. Thanks! I'll just for wait for CI and we can merge.

PR #1293, opened 2026-07-31
[Metrics] Datapoints follow OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE

### LINKED ISSUE #1289: [Metrics] Piping OTEL metrics directly to Prometheus causes two errors

KennanHunter:
Closes #1289

Wires OTEL metric exports up to `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` so operators can pick delta, cumulative, or low-memory temporality per the spec.

## Opinions

- **Case-insensitive parsing of `OtelTemporalityPreference`** from both the env var and the `[otel]` config section, aligned with OTEL's spec ([env var read](https://github.com/KennanHunter/pgdog/blob/otel_sum_improvements/pgdog-config/src/otel.rs#L113-L119), [custom `Deserialize`](https://github.com/KennanHunter/pgdog/blob/otel_sum_improvements/pgdog-config/src/otel_temporality.rs#L24-L38)).
- **Migrated `OpenMetric` to a real enum for metric type** instead of magic strings, so adding future variants like histogram stays ergonomic ([`OpenMetricType`](https://github.com/KennanHunter/pgdog/blob/otel_sum_improvements/pgdog/src/stats/open_metric.rs#L27-L41)).
- **Refactored the OTLP data-point calculation** to make the relationship between metric type, temporality, and the computed value explicit ([`value_for_data_point`](https://github.com/KennanHunter/pgdog/blob/otel_sum_improvements/pgdog/src/stats/otel.rs#L244-L272), [`wrap_data_points`](https://github.com/KennanHunter/pgdog/blob/otel_sum_improvements/pgdog/src/stats/otel.rs#L273-L292), [`build_attributes`](https://github.com/KennanHunter/pgdog/blob/otel_sum_improvements/pgdog/src/stats/otel.rs#L293-L308)).

I think before merge some level of integration testing could be nice. How would you go about that @levish0 ? We could also add Prometheus+graphana in the examples folder and include a graphana dashboard script.

The default export has been changed from Delta to Cumulative to match spec. I assume DataDog should have no issue with that, but it might be good for someone familiar with it to test it.

### DIFF
--- a/pgdog/src/stats/otel.rs
+++ b/pgdog/src/stats/otel.rs
@@ -120,7 +120,7 @@ pub struct Gauge {
 // little serde trick to let us serialize directly as the integer representation
 #[derive(serde_repr::Serialize_repr, Clone, Copy)]
 #[repr(u8)]
-pub enum SumAggregationTemporality {
+pub enum AggregationTemporality {
     Delta = 1,
     Cumulative = 2,
 }
@@ -128,7 +128,7 @@ pub enum SumAggregationTemporality {
 #[derive(Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct Sum {
-    pub aggregation_temporality: SumAggregationTemporality,
+    pub aggregation_temporality: AggregationTemporality,
     pub is_monotonic: bool,
     pub data_points: Vec<NumberDataPoint>,
 }
@@ -246,7 +246,7 @@ fn value_for_data_point(
     metric_name: &str,
     measurement: &Measurement,
     metric_type: OpenMetricType,
-    temporality: SumAggregationTemporality,
+    temporality: AggregationTemporality,
     now: &str,
 ) -> Option<(f64, Option<String>)> {
     let cumulative = measurement_to_f64(&measurement.measurement);
@@ -260,8 +260,8 @@ fn value_for_data_point(
             };
             let start = state.start_time(&key, now);
             let value = match temporality {
-                SumAggregationTemporality::Cumulative => cumulative,
-                SumAggregationTemporality::Delta => state.delta(&key, cumulative)?,
+                AggregationTemporality::Cumulative => cumulative,
+                AggregationTemporality::Delta => state.delta(&key, cumulative)?,
             };
             Some((value, Some(start)))
         }
@@ -272,7 +272,7 @@ fn value_for_data_point(
 /// the source metric type: `Gauge` for gauges, `Sum` (monotonic) for counters.
 fn wrap_data_points(
     metric_type: OpenMetricType,
-    temporality: SumAggregationTemporality,
+    temporality: AggregationTemporality,
     data_points: Vec<NumberDataPoint>,
 ) -> (Option<Gauge>, Option<Sum>) {
     match metric_type {
@@ -313,9 +313,9 @@ pub fn build_request(metrics: &[&Metric], now: &str) -> ExportMetricsServiceRequ
     let otel = &config.config.otel;
 
     let temporality = match otel.effective_temporality_preference() {
-        OtelTemporalityPreference::Cumulative => SumAggregationTemporality::Cumulative,
+        OtelTemporalityPreference::Cumulative => AggregationTemporality::Cumulative,
         OtelTemporalityPreference::Delta | OtelTemporalityPreference::LowMemory => {
-            SumAggregationTemporality::Delta
+            AggregationTemporality::Delta
         }
     };
 
@@ -329,7 +329,7 @@ pub fn build_request(metrics: &[&Metric], now: &str) -> ExportMetricsServiceRequ
 /// without touching global config or the process-wide static.
 fn build_request_with_state(
     state: &CounterState,
-    temporality: SumAggregationTemporality,
+    temporality: AggregationTemporality,
     namespace: Option<&str>,
     now: &str,
     metrics: &[&Metric],
@@ -680,7 +680,7 @@ mod test {
         let m1 = counter("total_queries", vec![], 10);
         let r1 = build_request_with_state(
             &state,
-            SumAggregationTemporality::Delta,
+            AggregationTemporality::Delta,
             None,
             &now_nanos(),
             &[&m1],
@@ -690,7 +690,7 @@ mod test {
         let m2 = counter("total_queries", vec![], 25);
         let r2 = build_request_with_state(
             &state,
-            SumAggregationTemporality::Delta,
+            AggregationTemporality::Delta,
             None,
             &now_nanos(),
             &[&m2],
@@ -705,7 +705,7 @@ mod test {
         let m1 = counter("total_queries", vec![], 10);
         let _ = build_request_with_state(
             &state,
-            SumAggregationTemporality::Delta,
+            AggregationTemporality::Delta,
             None,
             &now_nanos(),
             &[&m1],
@@ -714,7 +714,7 @@ mod test {
         let m2 = counter("total_queries", vec![], 3);
         let r2 = build_request_with_state(
             &state,
-            SumAggregationTemporality::Delta,
+            AggregationTemporality::Delta,
             None,
             &now_nanos(),
             &[&m2],
@@ -760,7 +760,7 @@ mod test {
         let m1 = build(10, 100);
         let _ = build_request_with_state(
             &state,
-            SumAggregationTemporality::Delta,
+            AggregationTemporality::Delta,
             None,
             &now_nanos(),
             &[&m1],
@@ -770,7 +770,7 @@ mod test {
         let m2 = build(15, 100);
         let r2 = build_request_with_state(
             &state,
-            SumAggregationTemporality::Delta,
+            AggregationTemporality::Delta,
             None,
             &now_nanos(),
             &[&m2],
@@ -804,7 +804,7 @@ mod test {
         let m1 = counter("total_queries", vec![], 1);
         let r1 = build_request_with_state(
             &state,
-            SumAggregationTemporality::Cumulative,
+            AggregationTemporality::Cumulative,
             None,
             &now_nanos(),
             &[&m1],
@@ -826,7 +826,7 @@ mod test {
         let m2 = counter("total_queries", vec![], 2);
         let r2 = build_request_with_state(
             &state,
-            SumAggregationTemporality::Cumulative,
+            AggregationTemporality::Cumulative,
             None,
             &now_nanos(),
             &[&m2],
--- a/pgdog/src/stats/otel.rs
+++ b/pgdog/src/stats/otel.rs
@@ -89,11 +89,18 @@ pub struct Gauge {
     pub data_points: Vec<NumberDataPoint>,
 }
 
+// little serde trick to let us serialize directly as the integer representation
+#[derive(serde_repr::Serialize_repr)]
+#[repr(u8)]
+pub enum SumAggregationTemporality {
+    Delta = 1,
+    Cumulative = 2,
+}
+
 #[derive(Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct Sum {
-    /// 1 = DELTA, 2 = CUMULATIVE
-    pub aggregation_temporality: u32,
+    pub aggregation_temporality: SumAggregationTemporality,
     pub is_monotonic: bool,
     pub data_points: Vec<NumberDataPoint>,
 }
@@ -281,7 +288,7 @@ pub fn build_request(metrics: &[&Metric], now: &str) -> ExportMetricsServiceRequ
                 (
                     None,
                     Some(Sum {
-                        aggregation_temporality: 1, // DELTA
+                        aggregation_temporality: SumAggregationTemporality::Delta,
                         is_monotonic: true,
                         data_points,
                     }),
--- a/pgdog/src/stats/otel.rs
+++ b/pgdog/src/stats/otel.rs
@@ -9,6 +9,7 @@ use std::time::{SystemTime, UNIX_EPOCH};
 
 use once_cell::sync::Lazy;
 use parking_lot::Mutex;
+use pgdog_config::otel_temporality::OtelTemporalityPreference;
 use serde::Serialize;
 
 use crate::util::hostname;
@@ -90,7 +91,7 @@ pub struct Gauge {
 }
 
 // little serde trick to let us serialize directly as the integer representation
-#[derive(serde_repr::Serialize_repr)]
+#[derive(serde_repr::Serialize_repr, Clone, Copy)]
 #[repr(u8)]
 pub enum SumAggregationTemporality {
     Delta = 1,
@@ -227,6 +228,13 @@ pub fn build_request(metrics: &[&Metric], now: &str) -> ExportMetricsServiceRequ
 
     let common_attrs = &*RESOURCE_ATTRIBUTES;
 
+    let aggregation_temporality = match config.config.otel.temporality_preference {
+        OtelTemporalityPreference::Cumulative => SumAggregationTemporality::Cumulative,
+        OtelTemporalityPreference::Delta | OtelTemporalityPreference::LowMemory => {
+            SumAggregationTemporality::Delta
+        }
+    };
+
     let otel_metrics: Vec<OtelMetric> = metrics
         .iter()
         .map(|metric| {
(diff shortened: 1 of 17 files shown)


KennanHunter:
Also added support to send the starting timestamp for a metric, in accordance with OTEL spec.

KennanHunter:
Looking at https://docs.datadoghq.com/opentelemetry/guide/otlp_delta_temporality/?tab=python#overview

> Datadog works best with delta aggregation temporality for monotonic sums, histograms, and exponential histograms.

I don't know quite what "works best" means, but it might be worth going out of spec and using delta as the default to just not break stuff. Weird.

KennanHunter:
Matching docs update

https://github.com/pgdogdev/docs/compare/main...KennanHunter:docs-1:otel-temporality-docs

KennanHunter:
Prometheus working well locally

levkk:
Nice! I will review tomorrow first thing!

> I assume DataDog should have no issue with that, but it might be good for someone familiar with it to test it.

IIRC, last time I tested this with DD, it only worked with delta. So, maybe we can default it it to delta if `datadog_api_key` is set in the config: https://github.com/pgdogdev/pgdog/blob/e37160da8177599ee4a9f56695fc3c07777d8f94/pgdog-config/src/otel.rs#L55

KennanHunter:
Great! I'll probably have an hour tonight to add that Datadog behavior, update and PR the documentation, and make any other changes you'd like to see.

I also wanted to quickly sneak in something to handle the case where these environment variables are defined but `[otel]` isn't present in the configuration, which currently disables any exporting even if the program has an OTEL_EXPORTER_ENDPOINT, which I know tripped me up. One way we could do that is to add a warning prompting them to add `[otel]` to their configuration, or we could change the behavior to use OTEL if the exporter is enabled via the env var, even when `[otel]` is not present. What are your thoughts?

Final thing: Would you be interested in this [graphana_prometheus example](https://github.com/KennanHunter/pgdog/commit/597b53b33d3d34dbac0e082d1887d3bbb16e639f)? I could clean it (currently largely the work of Claude) and throw up a PR. I mostly implemented it for the sweet dopamine release of seeing that locked percent gauge redline.

levkk:
> I also wanted to quickly sneak in something to handle the case where these environment variables are defined but [otel] isn't present in the configuration, which currently disables any exporting even if the program has an OTEL_EXPORTER_ENDPOINT, which I know tripped me up. One way we could do that is to add a warning prompting them to add [otel] to their configuration, or we could change the behavior to use OTEL if the exporter is enabled via the env var, even when [otel] is not present. What are your thoughts?

I think you should be able to make it a not-Option in `pgdog-config/core.rs` and give it a `Default` implementation which fetches stuff from the env. If nothing in the env, flag it as disabled, or something like that? 

> Final thing: Would you be interested in this [graphana_prometheus example](https://github.com/KennanHunter/pgdog/commit/597b53b33d3d34dbac0e082d1887d3bbb16e639f)? I could clean it (currently largely the work of Claude) and throw up a PR. I mostly implemented it for the sweet dopamine release of seeing that locked percent gauge redline.

Yes definitely. More examples = better!

> [...]  to add that Datadog behavior,

I need to test this manually against our DD account. I can do it quick, it's not an issue. The OTEL exporter was developed against datadog (it was me, my friend Claude and a DD dashboard), so I want to double check it still works there.

levkk:
Sorry for delay here. The day got away from me a bit. I still need to double check this works against Datadog. :+1:

KennanHunter:
Useful info from <https://docs.datadoghq.com/opentelemetry/guide/otlp_delta_temporality/?tab=python#implications-of-using-cumulative-aggregation-temporality>:

> If you opt to send OTLP monotonic sums, histograms, or exponential histograms with cumulative aggregation temporality, Datadog takes the difference between consecutive points on a timeseries. This means that:
>
> -    Your deployment is stateful, so you need to send all points on a timeseries to the same Datadog Agent or Datadog exporter. This affects how you scale your OpenTelemetry Collector deployments.
> -    Datadog might not send the first point it receives from a given timeseries if it cannot ensure this point is the true start of the timeseries. This may lead to missing points upon restarts.
> -    The minimum and maximum cannot be recovered for cumulative OTLP Histograms; they may be missing or approximated depending on the histograms export mode.

KennanHunter:
Trying a new pattern. Let me know how you like it. If you do, we could in the future look at pulling this out and using it in other configuration spots if that proves useful.

For every `[otel]` field, we check both sources. The env var wins over the TOML value. If both are set and differ, it logs a `warn!` naming the field and the two values, then uses the env value.

`[otel].temporality_preference` resolution order (first match wins):

1. Env var `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE`.
2. Explicit value in the `[otel]` TOML section.
3. `Delta` when `datadog_api_key` is set (via TOML or `DD_API_KEY`).
4. `Cumulative` otherwise.

If we set cumulative explicitly but have datadog's api key present it gives a pretty detailed warning as to why that's a bad idea, silence-able with a env var incase anyone finds a reason to want that behavior. 

Take your time getting that datadog integration tested. When this does get merged, I'll throw up the PR for the grafana example just to keep history simple.

levkk:
Checking back in here. The new config pattern would need to be backwards-compatible, i.e., the config set in the file must always take priority over what's in the env. The env is used as a default fallback, e.g.,:  https://github.com/pgdogdev/pgdog/blob/755f2ec62f71f40ddc87adb152758a775451faee/pgdog-config/src/general.rs#L1033

I'm not sure how to achieve this though. Given that we want the `[otel]` block to be non-optional with serde defaults... :thinking:

KennanHunter:
We can switch the order of preference where if both env and config are present we prefer config:

```diff
- fn warn_if_disagree_prefer_env<T: PartialEq + std::fmt::Debug>(
+ fn warn_if_disagree_prefer_config<T: PartialEq + std::fmt::Debug>(
    field: &str,
    toml_val: Option<T>,
    env_val: Option<T>,
) -> Option<T> {
    if let (Some(t), Some(e)) = (toml_val.as_ref(), env_val.as_ref())
        && t != e
    {
        warn!(
            "otel.{field} in TOML ({t:?}) disagrees with environment variable ({e:?}); using TOML value"
        );
    }

-    env_val.or(toml_val)
+    toml_val.or(env_val)
}
```

> Given that we want the [otel] block to be non-optional with serde defaults... 🤔

That's what cab17505 does: `otel: Otel` is `#[serde(default)]` on Config, and `Otel::default()` reads env vars via a resolve step. The reason for splitting a raw TOML struct from the resolved `Otel` is the cross-field defaults: temporality picks Delta when `datadog_api_key` is set via TOML or `DD_API_KEY`, which needs somewhere both config fields are visible at once. `#[serde(default)]` is shallow and can't do that.

I added the disagreement warning because it's trivial when you pull apart env and TOML parsing, but it's a little weird that it only exists for `[otel]`. If the pattern is useful, I'd say it's worth pulling it out into a separate PR that generalizes it across `general.rs` and the other serde-default sections. Otherwise happy to drop it here for consistency.

Longer term, worth taking inspiration from [`config`](https://docs.rs/config/) / [`figment`](https://docs.rs/figment/), they handle layered env + file resolution behind a macro so each config section stays flat instead of needing its own raw-struct pair. Out of scope here, but a cleaner shape if we ever want to generalize this across the codebase.

If you'd rather shrink scope to unblock this, the minimum is: keep Delta as default everywhere, use the plain serde-default pattern for env fallback (matching `general.rs`), and drop the `OtelRawToml` scaffolding.

pgdog/src/stats/otel_exporter.rs
```diff
@@ -26,6 +31,30 @@ pub async fn run() {
         None => return,
     };
 
+    // Only returns Ok(()) if the value has yet to have been checked
+    // Nonblocking since HAS_IGNORE_DATADOG_WARNING_BEEN_CHECKED is only on this thread
+    if let Ok(()) = HAS_IGNORE_DATADOG_WARNING_BEEN_CHECKED.set(()) {
```
levkk: This is something we should check during configuration checking at startup, see `Config::check` in `pgdog-config/src/core.rs`.

KennanHunter: Moved, also refactored to use a accessor method called `effective_temporality_preference` to avoid the fact that Config::check and Config::load don't get called in tests. IMO having config as a global static seems to cause some problems I'd be interested in what a refactor of that would look like.

KennanHunter:
Working well with datadog as of [81dc14f](https://github.com/pgdogdev/pgdog/pull/1293/commits/81dc14f7e29de434c6276ad207d5d777964a6473)

CI Failure seems to be unrelated, going to mark this as ready for review @levkk

levkk:
Taking a look this week!

pgdog/src/stats/otel.rs
```diff
@@ -89,11 +117,18 @@ pub struct Gauge {
     pub data_points: Vec<NumberDataPoint>,
 }
 
+// little serde trick to let us serialize directly as the integer representation
+#[derive(serde_repr::Serialize_repr, Clone, Copy)]
+#[repr(u8)]
+pub enum SumAggregationTemporality {
+    Delta = 1,
+    Cumulative = 2,
+}
```
alexkarp-umd: Hi, @KennanHunter, I've started work on implementing histograms for query latency and I'd like to suggest a change here: could `SumAggregationTemporality` drop the Sum prefix and just become `AggregationTemporality`?

OTLP defines a single `AggregationTemporality` enum, and Histograms need the same exact type as Sums. With the current name I'd have to write either `aggregation_temporality: SumAggregationTemporality` on a histogram, or a duplicate `HistogramAggregationTemporality` and I'd like to avoid either!

I'm happy to submit a patch if you don't want to do the refactoring yourself. Just let me know!

KennanHunter: Good change idea, I'll throw a refactor up within the next hour!

KennanHunter:
@alexkarp-umd renamed [SumAggregationTemporality to AggregationTemporality](https://github.com/pgdogdev/pgdog/pull/1293/commits/92ddcf3c513b1f973ec9ba5bcc568f01b8e330e4)

Glad to see histograms added, the point serialization logic I put in should be pretty easy to add more OpenMetricType variants to. Let me know if there's anything you run into, and feel free to reach me at `Kennnan` on the discord :)

PR #1296, opened 2026-07-31
fix: Binary COPY tuple encoding copies column data one byte at a time

### LINKED ISSUE #1294: Binary COPY tuple encoding copies column data one byte at a time
**PgDog version**

`v0.1.51` / `e37160da`

`Tuple::to_bytes` (`pgdog/src/frontend/router/parser/binary/tuple.rs:173`) re-encodes every row when sharding binary `COPY` data. It currently does:

```rust
let mut result = BytesMut::new();
...
for col in &self.row {
    result.put_i32(col.encoded_len());
    if let Data::Column(col) = col {
        result.extend(col);
    }
}
```

1. **`result.extend(col)` writes one byte at a time.** `BytesMut`'s `Extend` impl is a `for b in iter { self.put_u8(b) }` loop - the bytes crate marks it `// TODO: optimize` (bytes 1.11.1, `src/bytes_mut.rs:1399`). A 4 KB `bytea` column becomes 4096 iterations instead of one `memcpy`.

2. **The buffer starts at zero capacity and regrows.** The encoded size is known up front — the struct already computes it in `bytes_read()` - so `BytesMut::with_capacity` can allocate exactly once.

current flamegraph release showed the number as follows. ( release build numbers not seen yet )

Image

Note - AI has been used to frame the wording of this issue.

ygxio:
fixes: #1294

### DIFF
--- a/pgdog/src/frontend/router/parser/binary/tuple.rs
+++ b/pgdog/src/frontend/router/parser/binary/tuple.rs
@@ -171,21 +171,39 @@ impl Tuple {
 
 impl ToBytes for Tuple {
     fn to_bytes(&self) -> Bytes {
-        let mut result = BytesMut::new();
         if self.end {
+            let mut result = BytesMut::with_capacity(std::mem::size_of::<i16>());
             result.put_i16(-1);
-        } else {
-            result.put_i16(self.row.len() as i16);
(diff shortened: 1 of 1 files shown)


levkk (approved):
Awesome!

PR #1299, opened 2026-08-01
Track a client's parameter changes on the server connection

IgorOhrimenko:
## Problem

A `SET` or `RESET` the client sends once a server is attached changes that server's session, but nothing wrote it down. What we did instead was clear the whole parameter cache whenever a `CommandComplete` said `RESET` — which trades one problem for another, because that cache is exactly what tells us to undo the change for the next client.

`pg_dump -t <table>` walks straight into it:

```sql
SET search_path TO '';
BEGIN;
SET TRANSACTION ISOLATION LEVEL REPEATABLE READ, READ ONLY;
RESET search_path;
ROLLBACK;
```

The `ROLLBACK` undoes the `RESET`, so the empty `search_path` is back on the server, but the cache no longer mentions it. The next client gets that connection and every unqualified name fails with `42P01`.

A `SET` has the same hole in the other direction, and it needs no `pg_dump` at all:

```sql
BEGIN;
SELECT 1;                          -- a server is ours from here on
SET statement_timeout TO '5s';
COMMIT;
```

On `main` the next client sees `statement_timeout = 5s`. Nothing resets it, because the connection's parameter cache never learned about it — `link_client` only ever recorded what the client had at checkout time.

## Fix

Record the change where it happens, instead of reacting to a tag afterwards.

`RESET` gets the transaction handling `SET` always had — `reset` vs `reset_transaction`, next to `insert` and `insert_transaction` — so a rollback restores what it cleared and a commit makes it permanent. That also fixes a smaller thing on the way: an autocommit `RESET` used to leave its old value in `reset_params` forever, so an unrelated transaction rolling back later would resurrect it.

The server connection then keeps the same record its client does, and the existing parameter diff does the rest. The next client is handed a precise `RESET "search_path"` for what it doesn't want, rather than a connection nobody dares reuse.

The `CommandComplete` fallback stays for RESETs we don't see coming: with the query parser off, that is still all we have.

## Testing

- `Parameters`: rollback restores an in-transaction reset, a commit makes it permanent, and a reset outside a transaction is not resurrected by a later rollback.
- `Server`: a recorded reset survives `ROLLBACK` and still produces `RESET "search_path"` for the next client; a committed one produces nothing; a recorded `SET` is undone for the next client; resetting one parameter no longer forgets the others; an untracked `RESET` still clears the cache.
- `integration/python/test_session_params_leak.py`: the `pg_dump` sequence, a `SET` committed after connecting, and a committed `RESET`. Verified to fail on `main` (the first two) and pass here.
- On a live pool with one server connection, the wire shows exactly one `RESET "search_path"` when the connection is handed over — no `RESET ALL`, no churn.

### DIFF
--- a/integration/pgdog.toml
+++ b/integration/pgdog.toml
@@ -56,6 +56,18 @@ role = "replica"
 read_only = true
 
 # ------------------------------------------------------------------------------
+# ----- Database :: pgdog_leak -------------------------------------------------
+# Exactly one server connection, kept around: tests for session state leaking
+# between clients need the next client to get the same connection back.
+
+[[databases]]
+name = "pgdog_leak"
+host = "127.0.0.1"
+database_name = "pgdog"
+pool_size = 1
+min_pool_size = 1
+
+# ------------------------------------------------------------------------------
 # ----- Database :: pgdog_sharded ----------------------------------------------
 
 [[databases]]
--- a/integration/pgdog.toml
+++ b/integration/pgdog.toml
@@ -57,8 +57,7 @@ read_only = true
 
 # ------------------------------------------------------------------------------
 # ----- Database :: pgdog_leak -------------------------------------------------
-# Exactly one server connection, kept around: tests for session state leaking
-# between clients need the next client to get the same connection back.
+# One server connection, never reaped: the next client has to get the same one.
 
 [[databases]]
 name = "pgdog_leak"
--- a/integration/python/test_session_params_leak.py
+++ b/integration/python/test_session_params_leak.py
@@ -1,0 +1,71 @@
+"""Session parameters must not survive the client that set them.
+
+Runs against a database with a single server connection, so the next client
+always gets the connection the previous one used. current_setting() is used
+instead of SHOW because SHOW can be answered by PgDog itself.
+"""
+
+import psycopg
+
+
+def connect():
+    conn = psycopg.connect(
+        user="pgdog",
+        password="pgdog",
+        dbname="pgdog_leak",
+        host="127.0.0.1",
+        port=6432,
+    )
+    # Without autocommit every statement runs in a transaction that is rolled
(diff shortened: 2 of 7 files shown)


IgorOhrimenko:
Related: #1302 fixes a third way session state survives checkin — a client's SQL-level `PREPARE` (which `pg_dump` relies on) stayed on the connection and collided with the next client's statement of the same name.

levkk:
I don't think this is the right solution. We should implement handling `RESET` in `Parameters` just like we do with `SET` inside transactions. Marking a connection "dirty" is a trick of last resort: it causes connection churn and in production would be quite disruptive. We do it when we literally can't do anything else to cleanup the connection state, but in this case, I think we can.

IgorOhrimenko:
Reworked along those lines — no dirty flag, the connection keeps track of its parameters.

`RESET` now has the transaction handling `SET` always had (`reset` vs `reset_transaction`), so a rollback restores what it cleared and a commit makes it permanent. The server connection records the same change its client does, and the existing `link_client` diff does the rest: the next client is handed a precise `RESET "search_path"`, which is what the wire shows on a one-connection pool.

Digging into it turned up a second hole the same mechanism closes, with no `pg_dump` involved:

```sql
BEGIN;
SELECT 1;                          -- a server is ours from here on
SET statement_timeout TO '5s';
COMMIT;
```

On `main` the next client sees `statement_timeout = 5s`. The cache is only written at checkout, so a `SET` that lands after that is never undone. There's an integration test for it.

The `CommandComplete` fallback is still there for RESETs we don't see coming — with the query parser off it's all we have — but it no longer fires when we already know what changed, which is what made it destructive before.

One thing I did not do: an autocommit `RESET` used to stash its old value forever, so an unrelated transaction rolling back later resurrected it. `reset` no longer stashes and `reset_transaction` does, which changes two existing tests to use the transaction variant.

PR #1300, opened 2026-08-01
feat: ADD SHARD orchestrated shard provisioning (stacked on #1279)

rlittlefield:
**Do not review this diff.** It carries the whole stack against main
because GitHub cannot base an upstream PR on a fork branch. Review
happens layer by layer:

1. pgdogdev#1279: sharding key lookups with fixed shard placement
2. rlittlefield/pgdog#1: fleet visibility and the write barrier
3. rlittlefield/pgdog#2: ADD SHARD (this PR's own layer)

As each layer merges to main, this PR's diff shrinks to match; it
converts from draft to ready when only the ADD SHARD layer remains.

---

# ADD SHARD: orchestrated shard provisioning with WAL catch-up

## Stack

This PR sits on two others:

1. Sharding key lookups with fixed shard placement (`lookup_result =
   "shard"`, pgdogdev#1279): stored placement is what makes growing a
   cluster in place safe. This PR automates the growing.
2. Fleet visibility and the write barrier: the instance registry, the
   omni-write barrier, and the shared cutover registry this PR's
   coordinated cutover is built from.

Layer-scoped review happens on the fork, where the stack's branches can
be base branches: rlittlefield/pgdog#1 (fleet visibility and the write
barrier) and rlittlefield/pgdog#2 (this layer). This PR carries the full
stack against main and shrinks to just its own layer as the bases merge.

## What this PR adds

An `ADD SHARD` admin command (and a `pgdog add-shard` CLI) that grows a
cluster by one shard end to end while production traffic keeps flowing.

You declare each future shard once, in its final shape. The
`provisioning` flag keeps it out of the serving topology until its cutover
activates it:

```toml
[[databases]]
name = "prod"        # The database gaining a shard: this entry joins
                     # the existing [[databases]] entries for "prod".
host = "10.0.0.3"    # The new server.
shard = 2            # The shard number it will serve. Shards are added
                     # in order: with 0 and 1 serving, 2 goes first.
provisioning = true  # Keeps the entry out of the serving topology until
                     # the cutover. Declare as many future shards as you
                     # like (2, 3, 4...); each command names one.
```

```
-- Provision shard 2 of the "prod" database, from its provisioning
-- entry: sync all DDL, copy the omnisharded data, stream WAL until
-- caught up. Then park at "awaiting cutover".
psql admin -c "ADD SHARD prod 2"

-- Cut over: pause omni writes, drain to zero, activate the shard,
-- resume. (Also addressable by the task id from SHOW TASKS:
-- CUTOVER <id>.)
psql admin -c "CUTOVER SHARD prod 2"

-- Or both in one step, without the confirmation checkpoint.
psql admin -c "ADD SHARD prod 2 AUTO"
```

You create an empty database on the new server, deploy the entry, and
run the command. The database's existing `schema_admin` user covers the
new shard. pgdog does the rest:

1. **Guards**. The task refuses to run unless every sharded table is
   placement-stable, meaning `lookup_result = "shard"` or an explicit
   mapping. Hash-routed tables need a real reshard, because adding a shard
   would move their rows. The task also refuses when the named shard has
   no provisioning entry, when it isn't the next shard number (declared
   shards are added in order, one ADD SHARD each), when the new shard
   isn't empty, when another topology change for the database is already
   running, and when another pgdog instance holds the provisioning lock.
   That lock is a session-scoped advisory lock taken on the new shard
   itself, so a crashed holder releases it automatically.
2. **Schema sync**. The existing pg_dump-based schema sync provisions all
   DDL onto the new shard.
3. **Omni data**. Omnisharded tables are snapshot-copied from shard 0
   using the resharding data-sync machinery, then WAL-streamed until
   caught up. A publication scoped to the omnisharded tables is created
   automatically and dropped when the task ends. Pass one explicitly to
   manage it yourself. A database with no omnisharded tables skips this
   phase and the cutover's write pause entirely: the task syncs schema,
   parks, and activates.
4. **Cutover**. Only omnisharded writes pause. Sharded traffic and all
   reads flow throughout. The task drains replication to zero, flips the
   `provisioning` flag off in the running config, reloads, refreshes
   `pgdog.config` on every shard, persists the files when
   `cutover_save_config` is on, and resumes omni writes. The pause is the
   drain plus coordination, typically around a second.

Progress is visible in `SHOW TASKS` (validating, syncing schema, syncing
data, replicating, awaiting cutover, draining, swapping topology).
`STOP_TASK` aborts cleanly any time before the swap: slots and the
auto-created publication are dropped, and the topology never changed. The
swap is the point of no return. Everything after it is best-effort, and
each failure warns with its recovery step.

## Sequence

```mermaid
sequenceDiagram
    autonumber
    actor Op as You
    participant App as Application
    participant PD as pgdog running ADD SHARD
    participant Fleet as Other pgdog instances
    participant S0 as Shard 0
    participant NEW as New shard
    participant ALL as All shards

    Note over Op,NEW: MANUAL - create empty DB, deploy the provisioning entry, RELOAD

    Op->>PD: ADD SHARD prod 2
    PD->>PD: guards - placement stable, shard 2 is next, new shard empty
    PD->>NEW: take the cross-instance advisory lock
    PD->>S0: create publication for omni tables
    PD->>S0: pg_dump schema only
    PD->>NEW: apply all DDL with idempotent restore
    PD->>S0: create durable replication slot at L0
    loop each omni table
        PD->>S0: temp slot with snapshot, COPY
        PD->>NEW: rows consistent at Lt after L0
    end
    PD->>S0: start replication from L0
    loop until lag under threshold
        S0-->>PD: WAL omni changes
        PD->>NEW: apply with upsert and per table LSN dedup
        App->>ALL: traffic flows normally throughout
    end
    Note over PD: status awaiting cutover - WAL retained on shard 0 while parked

    Op->>PD: CUTOVER SHARD prod 2
    PD->>PD: pause local omni writes
    PD->>NEW: publish armed state
    Fleet-->>NEW: see armed, pause omni writes, ack
    Note over App,Fleet: omni writes park everywhere - reads and sharded traffic keep flowing
    S0-->>PD: in flight omni writes drain to zero lag
    PD->>S0: stop stream and drop slots
    PD->>PD: POINT OF NO RETURN - provisioning flag flips off, registry reloads with 3 shards
    PD->>PD: resume local omni writes, persist config if enabled
    PD->>ALL: refresh pgdog.config rows on every shard
    PD->>NEW: publish activated state
    Fleet-->>NEW: see activated, activate shard 2, resume omni writes, ack
    PD->>S0: drop publication, also on new shard
    App->>ALL: omni writes now reach all three shards
    Note over Op,App: MANUAL - remove the provisioning flag from the config source, app assigns tenants to shard 2
```

## Config as state

The `provisioning = true` flag makes the config declarative rather than
stateful. The entry already describes the end state, and activation only
deletes one line. When the config comes from a manifest or template
(GitOps), you remove the flag at the source at your leisure.
`cutover_save_config` covers file-managed setups.

The database records the truth either way: the cutover writes
`(shard = N, shards = N + 1)` into the new shard's own `pgdog.config`, and
convergence reads it back at startup (before the listener opens) and after
every RELOAD. A restart with a stale manifest re-activates the shard
before serving a single query.

## Multiple pgdog instances

The cutover coordinates a fleet, using the new shard itself as the shared
medium. Every instance that has the entry can reach it, and it serves
nothing else. The protocol lives in `backend/fleet` next to the registry
and barrier from the previous PR: a coordinator publishes topic-scoped
states (`pgdog.fleet_state`, `pgdog.fleet_acks`, NOTIFY riding each
write) and followers ack them. It ships in this PR rather than the
infrastructure one so its semantics are reviewed beside the only code
that exercises them.

- Every instance heartbeats `pgdog.instances` on shard 0. That table is
  the fleet registry, installed by `SETUP SCHEMA` and by the topology
  tasks. Graceful shutdowns (including SIGTERM) deregister the instance,
  so rolling restarts don't disturb a cutover.
- Every instance that sees provisioning entries runs an agent for the
  next declared shard. The agent registers on that shard, subscribes to
  a NOTIFY channel there (every coordinator state write signals it in
  the same statement), and keeps a 5-second fallback poll for missed
  notifications. State changes reach agents within milliseconds.
- The cutover refuses if a live instance hasn't registered on the new
  shard, naming it in the error. That instance is running a config without
  the entry, and the fix is deploying the config everywhere first.
- The protocol is a two-phase handshake with acks. The coordinator arms
  the omni-write barrier locally, publishes `armed`, and waits for every
  peer's ack. It drains to zero, swaps, resumes its own writes, and
  publishes `activated`. Peers activate, resume writes, and ack. Aborts
  publish `released`.
- Fail-safety is structural. The coordinator refreshes the armed row every
  5 seconds for as long as it holds the fleet's barriers, and agents
  resume writes only when that heartbeat goes silent for 30 seconds, which
  means the coordinator actually died. A dead but recently seen peer times
  out the arm phase and the task re-parks for a retry. An instance that
  missed the activation keeps its omni writes parked until it learns the
  outcome, and a crashed one converges at restart before serving. There is
  no path that writes with a stale topology.
- A solo instance coordinates with nobody and behaves exactly as before.

## Failure and crash behavior

Everything before the swap aborts cleanly and reruns from scratch.
Per-table copy slots are temporary. The durable slot and the auto-created
publication are dropped by cleanup guards. The omni-write barrier and the
provisioning cluster (whose session holds the advisory lock) are drop
guards, so even a hard-aborted cancellation releases them. After the swap,
a crash before the `pgdog.config` refresh is recovered with `SETUP SCHEMA`.
Once the marker is written, restarts converge on their own. There is no
rollback after the swap, because once lookups assign a tenant to the new
shard, removing it would strand data.

## Notes for review

- The drain honors `cutover_timeout` and `cutover_timeout_action`. With
  `abort`, the task releases the barriers and re-parks at "awaiting
  cutover" instead of failing, so the cutover can be retried.
- While parked, the durable slot on shard 0 retains WAL. It's visible in
  `SHOW REPLICATION_SLOTS`, so don't park indefinitely.
- Coordination timing constants (agent poll 1s, arm ack timeout 10s,
  liveness window 15s, heartbeat 5s) are code constants for now. Promoting
  them to config is a reasonable follow-up if fleets need it.
- The task is organized as one `run()` that reads as a table of
  contents over phase modules (`guards`, `provision`, `cutover`,
  `schema_only`), each with its contract (entry, exit, failure) in a
  doc comment, and every lock and barrier held through an RAII guard.
  An enum state machine was considered and rejected: the flow is linear
  with one loop, and the failure behavior lives in guard drop order,
  which a step dispatcher would bury in an Option-filled context
  struct.
- The `pgdog.fleet_state` and `pgdog.fleet_acks` tables are created on
  the new shard when a cutover starts and dropped once every instance
  acked. Nothing persists across attempts.
- Integration suite: `integration/add_shard/run.sh` covers a parked
  cutover with a live omni writer, auto cutover, abort mid-task, the
  cross-instance lock, convergence after a restart with a stale flag, a
  two-instance coordinated cutover, a dead-peer refusal, and a
  schema-only run without omnisharded tables.


.cargo/config.toml
```diff
@@ -32,3 +32,10 @@ rustflags = [
 #  thread 'main' (12096) has overflowed its stack
 [target.x86_64-pc-windows-msvc]
 rustflags = ["-C", "link-arg=/STACK:2097152"]
+
+# Multi-step planning walks deep parse trees recursively. The recursion
+# is finite, but debug-build stack frames are large enough that the
+# 2 MiB default test-thread stack overflows on the deepest paths.
+# Release builds use a fraction of this.
```
levkk: This is configurable in `[memory]`: https://docs.pgdog.dev/configuration/pgdog.toml/memory/#stack_size

pgdog-config/src/database.rs
```diff
@@ -204,6 +204,12 @@ pub struct Database {
     /// Used for resharding only; this database will not serve regular traffic.
     #[serde(default)]
     pub resharding_only: bool,
+    /// This entry is a shard being provisioned by `ADD SHARD`: declared in
+    /// its final shape (`name`, `shard`, `host`) but excluded from the
+    /// serving topology until the cutover activates it. Remove the flag
+    /// from the config once the shard is active.
+    #[serde(default)]
+    pub provisioning: bool,
```
levkk, line 212: This is something you won't be able to currently synchronize between multiple PgDog pods easily. It would require careful integration and orchestration. A simpler, alternative way to handle this is to spin up a new PgDog deployment just for provisioning, run the _ADD SHARD_ command and once the shard is ready, add it to the config in the deployment serving queries.

rlittlefield: Yeah, this is a weird one. We only need the "provisioning" flag as a way to tell pgdog instances not to use it yet. Then we use a different mechanism to say "ok, ready" - you don't have to synchronize the removal of the flag, but probably do want to ensure that all the instances have reloaded the config before starting an ADD SHARD.

Once the cutover happens, we use the LISTEN with a polling fallback for each of the instances to know that it occurred. And if instances restart or somehow miss the listen, can ask the shard if it is ready.

We could use _only_ a marker in the shard's pgdog schema, but then there is a weird case if the shard is unavailable, we can't tell if it ever was fully provisioned and that might have some implications on how to handle down shards?

I hadn't considered the provisioning only deployment. I don't know how that will impact the live cutover and stuff. It would still have to pause writes to the omnisharded tables if we want to keep this a live cutover. I'll need to think about this a bit.

rlittlefield:
Before anyone spends any more time reviewing this, I'm going to ask for a quick pause while I split this into some smaller chunks that are more easily reviewed, then do a cleanup pass on it!

rlittlefield:
Closing in favor of layer-by-layer review: pgdogdev#1279 (sharding key lookups with fixed shard placement), then rlittlefield/pgdog#1 (fleet visibility and the write barrier), then rlittlefield/pgdog#2 (ADD SHARD). Each shows exactly its own diff. An upstream PR for the ADD SHARD layer will follow once its bases have merged; the earlier review comments here still apply and were addressed in the restructure.

PR #1301, opened 2026-08-02
[Example] Add grafana+prometheus example

KennanHunter:
Adds a example runnable with a single `docker compose up` that connects a grafana dashboard.

Dependent on #1293


levkk:
Looks great, thanks for contributing this. One thing I would remove is the synthetic app: we generally try not to include example apps that are not part of the integration suite, just because it would add to the compile time of the project and probably won't be used. But everything else, incl. the docker-compose for running prometheus/grafana is very helpful!

KennanHunter:
Makes sense to me, removed it. Let me know if there's anything else you'd like to see.

levkk (approved):
Amazing, thank you so much!

PR #1303, opened 2026-08-03
Wait for shutdown instead of assuming it took under a second

IgorOhrimenko:
## Problem

`integration/complex/shutdown.sh` sleeps for one second after asking PgDog to shut down, then fails the job if the process is still around. Draining the pools takes as long as it takes, and on a busy runner that is sometimes longer than a second, so the job fails with nothing actually wrong.

From a recent run — every server connection is closed, the check fires one second later, and the process exits right after:

```
07:29:18.48  closing server connection: state=idle, reason=pool offline  (last one)
07:29:19.49  6749
07:29:19.49  Shutdown failed
07:29:19.49  🐕 PgDog is shutting down immediately [SIGTERM]
```

It has failed this way on `main` as well (`e37160da`), and it flips between runs of the same commit.

## Fix

Poll for up to ten seconds instead of guessing. A pooler that genuinely doesn't shut down still fails the test — it just no longer depends on how loaded the runner is.

Verified against a process that exits immediately (returns at once), one that exits after 2.5s (previously a failure, now passes), and one that never exits (still fails, after ~10s).

### DIFF
--- a/integration/complex/shutdown.sh
+++ b/integration/complex/shutdown.sh
@@ -3,6 +3,21 @@ set -e
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 source ${SCRIPT_DIR}/../common.sh
 
+# Draining the pools takes as long as it takes, and a busy runner needs more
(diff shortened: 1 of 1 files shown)


levkk:
Thank you!

PR #1304, opened 2026-08-03
Stop testing implementation detail of exact transaction count

sgrif:
This test is purely trying to test that 2pc works. If the data was inserted correctly, appears where we expect it to be, and the 2pc stats are correct, we don't need to concern ourselves with exactly how many health checks and/or schema queries ran during that time.

### DIFF
--- a/integration/go/go_pgx/pg_tests_test.go
+++ b/integration/go/go_pgx/pg_tests_test.go
@@ -19,9 +19,9 @@ func assertShowField(t *testing.T, query string, field string, value int64, user
 	assert.Equal(t, value, actualValue, fmt.Sprintf("\"%s\" in %s is not %d", field, query, value))
 }
 
-func assertShowFieldGt(t *testing.T, query string, field string, value int64, user string, database string, shard int64, role string) {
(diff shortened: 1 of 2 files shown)


integration/go/go_pgx/pg_tests_test.go
```diff
@@ -19,9 +19,9 @@ func assertShowField(t *testing.T, query string, field string, value int64, user
 	assert.Equal(t, value, actualValue, fmt.Sprintf("\"%s\" in %s is not %d", field, query, value))
 }
 
-func assertShowFieldOneOf(t *testing.T, query string, field string, values []int64, user string, database string, shard int64, role string) {
+func assertShowFieldGt(t *testing.T, query string, field string, value int64, user string, database string, shard int64, role string) {
 	actualValue := showField(t, query, field, user, database, shard, role)
-	assert.Contains(t, values, actualValue, fmt.Sprintf("\"%s\" in %s is not one of %v", field, query, values))
+	assert.Greater(t, value, actualValue, fmt.Sprintf("\"%s\" in %s is not greater than %d", field, query, value))
```
levkk: Greater or equal

levkk (approved):
One flaky test down!

PR #1305, opened 2026-08-03
Don't require RELOAD to keep exactly the same connection count

sgrif:
The goal of this test to ensure that the existing connections are reused when `RELOAD` is issued. However, the exact number of connections needed to perform schema loading might differ slightly each time depending on the exact timing of when the queries complete. It's fine if the connection count increases slightly, we mostly want to make sure it doesn't double. 10% should give us enough wiggle room for this test to test what it's trying to test without breaking as implementation details change

### DIFF
--- a/integration/rust/tests/integration/reload.rs
+++ b/integration/rust/tests/integration/reload.rs
@@ -60,9 +60,9 @@ async fn test_reload_connection_count_stable() {
         .unwrap()
         .get(0);
 
-    assert_eq!(
-        before, after,
-        "connection count changed after RELOAD: before={before}, after={after}"
+    assert!(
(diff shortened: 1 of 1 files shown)


levkk (approved):
Disagree on premise, but agree that having a flaky test is disruptive and not worth the noise.

The _idea_ of this test is very important: we need to be sure that `RELOAD` works adequately, i.e., connection count remains the same. I think the way to test this reliably is to have the pool at full capacity during reload, so no new connections can be created.

That way, if we see 10% more, that's definitely a bug.

Anyway, this is good as is, just keep it in the back of your mind for when you have a second. We might have unit test coverage for this as well, so this is completely fine to merge as-is.

PR #1310, opened 2026-08-04
test: elixir test suite

brentjanderson:
I'm exploring PGDog on Elixir, and noticed that #1067 mentioned running Elixir/Postgrex through PGDog for regression tests. #464 never closed out.

This PR was vibed up with Claude, but I've reviewed it and it seems a pretty reasonable approach overall. If you have any issues with how it's constructed, let me know and I'm happy to address them directly.


levkk:
I don't know how, but you literally read my mind. We just discussed adding these today.

integration/elixir/common.sh
```diff
@@ -0,0 +1,30 @@
...
+#
+ELIXIR_COMMON_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+source "${ELIXIR_COMMON_DIR}/../common.sh"
+
+# Fetch deps and run the ExUnit suite.
+#
+# Elixir/Erlang have to already be on PATH: CI installs them with
+# erlef/setup-beam, local dev uses whatever the developer has (asdf, mise,
+# Homebrew, ...).
+function dev_suite() {
```
levkk: Could you move this into `dev.sh` inside the elexir suite?

levkk (approved):
Awesome, thanks! If you have a second, would be cool to move the elixir runner into its own `dev.sh`, but otherwise, this is good to go.

integration/elixir/common.sh
brentjanderson: Done. I think that's closer to how other suites run it, but I can iterate here if it's still not quite there.

PR #1312, opened 2026-08-05
Feat/normalize config identifiers

### LINKED ISSUE #1238: Normalize PostgreSQL identifiers in pgdog.toml
PgDog treats configured identifiers as literal, case-sensitive strings. It should follow PostgreSQL semantics
- lowercase unquoted identifiers
- preserve quoted identifiers
ensuring configuration matches parsed SQL consistently.

example:
- MyFunction() → myfunction
- "MyFunction"() → MyFunction

as per [comment](https://github.com/pgdogdev/pgdog/pull/1236/changes#r3614084716)

nsavvide:
This PR introduces a helper method to normalize the identifiers of config files. This follows the same procedure as query parsing per the protocol. 

The consumption of the helper is done at `resolve_sharded_table` on `name`, `schema` and `column`.

Tests have been added aswell :test_tube: 

This PR closes #1238 and helps to reduce the complexity in the following PR: #1236

### DIFF
--- a/pgdog-config/src/util.rs
+++ b/pgdog-config/src/util.rs
@@ -53,6 +53,26 @@ pub fn random_string(n: usize) -> String {
         .collect()
 }
 
+/// Fold a configured identifier the way PostgreSQL's parser folds
+/// identifiers in SQL, so values from `pgdog.toml` compare equal to
+/// the ones the query parser produces.
+///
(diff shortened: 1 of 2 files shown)


levkk (approved):
Awesome, thank you!

PR #1318, opened 2026-08-06
fix: don't hand clients a connection Postgres already closed

### LINKED ISSUE #614: [Healthchecks] Dont give connection to client that's been shut down
### Description

Detect connection that's been closed by Postgres (`terminating connection due to administrator command`) and attempt to fetch another from the connection pool.

nsavvide:
Healthchecks only run every `healthcheck_interval` (30s by default), so a backend terminated by `pg_terminate_backend` or an `admin` restart stayed in the pool and was handed to clients unchecked until the interval elapsed.

Adds a non-blocking liveness peek on every checkout. An idle pooled connection should have nothing waiting to be read, so pending bytes or a closed socket disqualify it: the connection is force-closed and the existing retry loop fetches another.

Pending data is treated as fatal without parsing it. That also discards connections that merely received an asynchronous `NoticeResponse`, which costs a reconnect but never correctness, and keeps the check cheap enough to run on every checkout.

Uses a distinct `Error::ServerClosed` rather than HealthcheckError so a dead connection does not toggle pool health. The check runs far more often than a healthcheck, and one terminated backend must not ban the pool or trigger failover.

Closes #614

### DIFF
--- a/pgdog/src/backend/pool/pool_impl.rs
+++ b/pgdog/src/backend/pool/pool_impl.rs
@@ -17,7 +17,7 @@ use crate::backend::pool::LsnStats;
 use crate::backend::{ConnectReason, DisconnectReason, Server, ServerOptions};
 use crate::config::PoolerMode;
 use crate::net::messages::{BackendPid, FrontendPid};
-use crate::net::{Parameter, Parameters};
+use crate::net::{Liveness, Parameter, Parameters};
 
 use super::inner::CheckInResult;
 use super::{
@@ -193,6 +193,7 @@ impl Pool {
                 Ok(conn) => return Ok(conn),
                 // Try another connection.
                 Err(Error::HealthcheckError) => continue,
+                Err(Error::ServerClosed) => continue,
                 Err(err) => return Err(err),
             }
         }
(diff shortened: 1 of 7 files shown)


levkk:
Fantastic! Let me benchmark this locally to double check that this doesn't have a performance impact!

levkk:
Strangely I see absolutely no performance regression here..which is good of course, but I expected this to make an extra syscall on each checkout! I'm going to wait for GH actions to come back to life and probably merge this as-is. Great work!

pgdog/src/backend/pool/pool_impl.rs
```diff
@@ -217,6 +218,13 @@ impl Pool {
         healthcheck_interval: Duration,
         now: Instant,
     ) -> Result<Guard, Error> {
+        if conn.liveness() != Liveness::Clean {
+            conn.stats_mut().state(crate::state::State::ForceClose);
+            conn.disconnect_reason(DisconnectReason::ServerClosed);
+            drop(conn);
```
levkk: nit: this is redundant, `conn` is dropped because it's owned by this function and we don't return it.

nsavvide: Resolved: b12f604362d31a6fca14b7cb44f40ed1484bd6fe

nsavvide:
> Strangely I see absolutely no performance regression here..which is good of course, but I expected this to make an extra syscall on each checkout! I'm going to wait for GH actions to come back to life and probably merge this as-is. Great work!

Makes sense, `now_or_never()` polls the peek once, and tokio checks cached readiness first, so on a healthy idle connection it returns Pending without ever issuing `recv(MSG_PEEK)`. The syscall only happens when the `fd` is already flagged readable, i.e. when there's genuinely something pending.

levkk (approved):

PR #1321, opened 2026-08-07
Fix interval text parsing for negative values and add tests

### LINKED ISSUE #1316: Negative `INTERVAL` values lose their sign in the text decoder, returning incorrect cross-shard `SUM` and `ORDER BY` values
**PgDog Version**

v0.1.50, PostgreSQL 18, 2 shards.

**Description**

The text decoder for `INTERVAL` parses each colon-separated component of the
time-of-day portion as an independently signed integer. For example, `-00:45:00` will parse as +45 minutes because the leading negative sign will be dropped on parsing when converting to an integer. Rows are stored correctly on each shard but computed results for `SUM` and `ORDER BY` are returned incorrectly.

Only clients that request text result format are affected (libpq, psql, psycopg2, node-postgres, etc.)

The relevant code is in `pgdog-postgres-types/src/interval.rs`, in the `Format::Text` arm of `FromDataType::decode`:

```rust
let mut value = value.split(":");   // "-00:45:00" -> ["-00", "45", "00"]
result.hours   = bigint(hours)?;    // bigint("-00") == 0  sign lost here
result.minutes = bigint(minutes)?;  // sign never applied to minutes
result.seconds = bigint(seconds)?;
```

**Configuration**

Requires a sharded table. Minimum configuration:

```toml
[[sharded_tables]]
database = "pgdog"
name = "time_entries"
column = "user_id"
```

**Reproduction**

```sql
CREATE TABLE time_entries (
    id BIGINT PRIMARY KEY,
    user_id BIGINT NOT NULL,      -- sharding key
    duration INTERVAL NOT NULL
);
```

Insert two rows that hash to different shards.

```sql
INSERT INTO time_entries (id, user_id, duration) VALUES (101, 1, '30 minutes');
INSERT INTO time_entries (id, user_id, duration) VALUES (102, 3, '-45 minutes');
```

Values are stored correctly:

```
shard 0: 00:30:00
shard 1: -00:45:00
```

**Incorrect Sum**

```sql
SELECT SUM(duration) FROM time_entries;
```

expected: `-00:15:00`

returned: `0 years 0 mons 0 days 01:15:00.0`

`-45 minutes` is added as `+45 minutes`, so the total comes back positive.

**Incorrect Ordering**

```sql
SELECT id, duration FROM time_entries ORDER BY duration ASC;
```

expected: `102` (-45 min), then `101` (+30 min)

returned: `101`, then `102`

Note the sort column has to be in the `SELECT` list. `SELECT id FROM time_entries ORDER BY duration` returns rows in shard order regardless of the sign bug, which masks this symptom.

Happy to submit a PR if this looks right

bryanmehall:
Negative INTERVAL values lose their sign in the text decoder, returning incorrect cross-shard SUM and ORDER BY values.
Fixes #1316

### DIFF
--- a/pgdog-postgres-types/src/interval.rs
+++ b/pgdog-postgres-types/src/interval.rs
@@ -141,7 +141,14 @@ impl FromDataType for Interval {
                             _ => (),
                         }
                     } else {
-                        let mut value = value.split(":");
+                        // Postgres writes one sign for the whole time-of-day
+                        // group: "-01:02:03" is -(1h 2m 3s)
+                        // Parse the components as magnitudes and apply the sign
+                        // to all of them
+                        let negative = value.starts_with('-');
+                        let time = value.strip_prefix(['-', '+']).unwrap_or(value);
+
+                        let mut value = time.split(":");
                         let hours = value.next();
                         if let Some(hours) = hours {
                             result.hours = bigint(hours)?;
@@ -163,6 +170,13 @@ impl FromDataType for Interval {
                             if let Some(millis) = millis {
                                 result.micros = parse_fractional_micros(millis)?;
                             }
+                        }
+
+                        if negative {
+                            result.hours = -result.hours;
+                            result.minutes = -result.minutes;
+                            result.seconds = -result.seconds;
+                            result.micros = -result.micros;
                         }
                     }
                 }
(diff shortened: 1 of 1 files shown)


levkk (approved):
Awesome, thank you!

PR #1322, opened 2026-08-07
feat: add rows inserted, deleted and completed metrics

murex971:
## Summary
- Count rows affected by `INSERT` / `UPDATE` / `DELETE` from Postgres `CommandComplete` tags

## Test
- [x] Unit tests for CommandComplete tag/rows parsing
- [x] Unit tests for Count Add/Sub/Div including new fields

### DIFF
--- a/pgdog/src/backend/stats.rs
+++ b/pgdog/src/backend/stats.rs
@@ -14,7 +14,7 @@ use crate::{
     config::Memory,
     net::{
         Parameters,
-        messages::{BackendPid, FrontendPid},
+        messages::{BackendPid, CommandComplete, FrontendPid},
     },
     state::State,
 };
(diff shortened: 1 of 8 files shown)


pgdog/src/backend/stats.rs
```diff
@@ -194,6 +194,25 @@ impl Stats {
         self.local.last_checkout.bind += 1;
     }
 
+    /// Record rows affected from a Postgres CommandComplete tag.
+    pub fn rows_affected(&mut self, tag: &str, rows: usize) {
```
levkk: nit: could you pass `&CommandComplete` here instead? This preserves the type information and makes the code more readable I think: we are basically saying that the row count can only be gotten from a `CommandComplete` message.

levkk (approved):
Looks great, just one small function signature change and we should be good to merge.

pgdog/src/backend/stats.rs
murex971: addressed

murex971:
@levkk ready :stuck_out_tongue:

levkk (approved):
Amazing, thank you!

PR #1323, opened 2026-08-07
lb: default auto mode pools to replica

### LINKED ISSUE #1308: [LB] Default to replica in auto mode
Set all pools to `role=replica` until primary election takes place. This allows traffic on the cluster faster and doesn't block read traffic if one of the databases isn't reachable.

murex971:
## Summary
- In auto mode, start all pools as `replica` so reads can flow before primary election finishes
- Wait for a primary to be elected on write checkouts (instead of waiting for “no Auto roles left”), so writes stay correct during election

closes #1308


levkk (approved):
This is excellent. Thank you!

PR #1329, opened 2026-08-10
Feat/unix sockets

### LINKED ISSUE #824: Support Unix domain socket
It is also "peer" mode in Postgres. User can connect to Postgres without having to remember password.

hhamud:
This PR is still in its preliminary stage but the basic function works with unix sockets on the backend. 

* Closes #824


integration/unix_sockets/run.sh
meskill, line 1: this have to be added to the ci suite as new entry otherwise it won't run

integration/unix_sockets/run.sh
```diff
@@ -0,0 +1,97 @@
...
+[general]
+auth_type = "trust"
+
+[[databases]]
+name = "pgdog"
+host = "$SOCKET_DIR"
+port = $PG_PORT
+database_name = "pgdog"
+user = "pgdog"
+EOF
```
meskill: let's just create the config file like most other tests do, I think we use just default static values there

integration/unix_sockets/run.sh
```diff
@@ -0,0 +1,97 @@
...
+    fail "pg_isready not found (is Postgres client installed?)"
+fi
+if ! pg_isready -h "$SOCKET_DIR" -p "$PG_PORT" >/dev/null 2>&1; then
+    fail "Postgres is not listening on unix socket $SOCKET_DIR:$PG_PORT"
+fi
+pass "Postgres is listening on unix socket $SOCKET_DIR (port $PG_PORT)"
+
+# --- 1. Build pgdog -----------------------------------------------------------
+echo "==> building pgdog"
+cargo build --manifest-path "$ROOT/Cargo.toml" --bin pgdog
```
meskill: verify the shared logic for integration tests - we have common scripts that will run and wait the pgdog available

pgdog/src/backend/auth/rds_iam.rs
```diff
@@ -37,7 +37,8 @@ fn resolve_region(addr: &Address) -> Result<String, Error> {
         return Ok(region.clone());
     }
 
-    infer_region_from_rds_host(&addr.host).ok_or_else(|| {
+    let host = addr.host.tcp().expect("Host must be a TCP address");
```
meskill: avoid panics in the runtime code until it's 100% justified

integration/unix_sockets/run.sh
hhamud: Do you mean explicitly to be added to the `ci.yaml` file as another script to run?

meskill: yes

PR #1331, opened 2026-08-10
RESET PREPARED: clear the cache with default config, never reuse statement names

IgorOhrimenko:
Two related defects in the prepared statements cache admin path, split out of #1311 so they can be reviewed on their own.

## RESET PREPARED does nothing with the default config

`ResetPrepared::execute()` passed `prepared_statements_limit` into `close_unused()`, and the default limit is unlimited (`i64::MAX`) — `len() - capacity` saturates to 0 and the command removes nothing. It now passes `0` explicitly: drop everything not in use, whatever the configured limit is.

## close_unused(0) wiped in-use statements and reused global names

`close_unused(0)` called `reset()`: it dropped statements clients still hold and rolled the global name counter back to zero. After that, a server connection that still has an old `__pgdog_N` statement prepared can be handed a *different* query under the same name — the server-side cache check compares names only. The same path ran once a second from the maintenance task for anyone who sets `prepared_statements_limit = 0`.

`close_unused(0)` now removes only unused statements and never touches the counter; `reset()` is `#[cfg(test)]`.

## Testing

- `close_unused(0)` keeps statements in use, removes released ones, and the next insert gets a fresh global name (fails on main: everything is wiped and the name is reused).
- `RESET PREPARED` clears released statements under the default config (fails on main: no-op).
- Parser test for `RESET PREPARED` — `RESET QUERY_CACHE` had one, this command didn't.

### DIFF
--- a/pgdog/src/admin/parser.rs
+++ b/pgdog/src/admin/parser.rs
@@ -256,6 +256,12 @@ mod tests {
     }
 
     #[test]
+    fn parses_reset_prepared_command() {
+        let result = Parser::parse("RESET PREPARED");
+        assert!(matches!(result, Ok(ParseResult::ResetPrepared(_))));
+    }
+
+    #[test]
     fn rejects_unknown_admin_command() {
         let result = Parser::parse("FOO BAR");
         assert!(matches!(result, Err(Error::Syntax)));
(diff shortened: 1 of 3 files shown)


levkk (approved):
Nice!

PR #1332, opened 2026-08-10
Keep portal open when H is last with B or E in the sequence

### LINKED ISSUE #1330: Transaction mode loses portals when a client Flushes before Execute (SQLSTATE 34000) — breaks node-postgres cursors/streaming
**Version:** reproduced on v0.1.48, v0.1.51 and v0.1.52 (latest). `pooler_mode = "session"` is unaffected.

## Summary

In `pooler_mode = "transaction"`, a client that terminates an extended-query batch with `Flush` before sending `Execute` — a legal protocol pattern for reading the `RowDescription` before deciding whether to execute — loses its portal. The subsequent `Execute` lands on a different backend and fails:

```
ERROR:  portal "C_1" does not exist
SQLSTATE: 34000
```

Both **named and unnamed** portals are affected.

## Real-world impact: node-postgres cursors and streaming

This is not only reachable by hand-rolled protocol code. The standard node-postgres paging/streaming stack opens with `Parse/Bind/Describe/Flush` — no `Execute` in the first batch (visible in PgDog's own debug logs as `['P','B','D','H']`) — and is deterministically broken through transaction mode:

| Client | Result |
|---|---|
| `pg-cursor` 2.22 | `34000 portal "C_1" does not exist` — fails on the **first** page, 0 rows |
| `pg-query-stream` 4.17 | `34000` |
| `knex` 3.3 `.stream()` | `34000 portal "C_2"` |

Anything layered on `pg-query-stream` (e.g. TypeORM/Sequelize streaming) inherits this. All three return every row when pointed directly at the same PostgreSQL instance, and all three work through PgDog if the cursor is wrapped in an explicit `BEGIN`/`COMMIT` (which pins the backend for the transaction) — that, or session mode, are the workarounds I found.

For contrast, the default prepared-statement paths of node-postgres, psycopg3, pgx, pgjdbc, Npgsql, libpq and asyncpg all send `Parse/Bind/Describe/Execute/Sync` as one Sync-terminated batch and are unaffected (Npgsql and pgjdbc cannot emit a protocol Flush at all).

## Reproduction (wire level)

Two batches, identical statement, same PgDog, same backend.

**Works** — single Sync-terminated batch:

```
Parse / Bind / Describe / Execute / Sync
-> row returned
```

**Fails** — Flush before Execute:

```
Parse / Bind / Describe / Flush
   <- read ParseComplete, BindComplete, RowDescription
Execute / Sync
-> ERROR 34000 portal "" does not exist
```

Deterministic at inter-batch gaps of **0, 0.05, 0.5 and 3.0 seconds**, so this is not a timing race. It reproduces on a **freshly started** PgDog against an untouched backend — no reload, no maintenance mode, no configuration change involved. Identical over TLS and plaintext.

Note for anyone attempting to reproduce: `psql \bind`, `pgbench -M prepared` and libpq-based clients will come back clean (`pgbench` ran 127,332 transactions with 0 failures here), because libpq sends a single Sync-terminated batch. Use `pg-cursor`, or split the batch at the protocol level.

## Mechanism

`ClientRequest::is_complete()` (`pgdog/src/frontend/client_request.rs:99-119` at v0.1.52) treats `'H'` (Flush) as a dispatch boundary identically to `'S'` (Sync):

```rust
if matches!(message.code(), 'H' | 'S' | 'Q' | 'c' | 'f' | 'F') {
    return true;
}
```

The backend is then eligible for release via `Server::done()` (`pgdog/src/backend/server.rs:769-771` at v0.1.52):

```rust
pub fn done(&self) -> bool {
    self.prepared_statements.done() && !self.in_transaction() && !self.statement_executed
}
```

`statement_executed` is set true only on `CommandComplete` and `PortalSuspended` (`server.rs:660,662`). A batch that has Bound but not yet Executed leaves it `false`, so `done()` returns true and `cleanup_backend` releases the server — taking the portal with it. The client's subsequent `Execute` is routed to a different backend, which has no such portal.

Since PgDog keeps no portal state of its own, a `34000` here is a reliable signal that Bind and Execute were served by different backends. (Named *prepared statements* survive this fine — the prepared-statement cache re-Parses them on the new backend; portals have no equivalent.)

## Suggested direction

An open portal — Bound but neither Executed nor Closed — should make `Server::done()` return false, so `cleanup_backend` cannot release the connection at a Flush boundary. Equivalently, Flush could be a dispatch boundary for *writing* purposes without being a release boundary for *pooling* purposes; the two concerns appear to be conflated in `is_complete()`.

## Environment

- PgDog `ghcr.io/pgdogdev/pgdog:v0.1.48`, re-checked against v0.1.51 and v0.1.52
- `pooler_mode = "transaction"`, PostgreSQL 17 backend
- Wire-level client: hand-rolled PostgreSQL v3 protocol implementation (SCRAM-SHA-256), validated against a direct PostgreSQL connection first (both batch shapes return a row) before being pointed at PgDog
- Node cases: `pg` 8.x with `pg-cursor` 2.22, `pg-query-stream` 4.17, `knex` 3.3 on Node 22

nsavvide:
## Problem

In `pooler_mode = "transaction"`, a client that ends an extended-query batch with `Flush` before `Execute` loses its portal:

```txt
  ERROR:  portal "C_1" does not exist
  SQLSTATE: 34000
```
  
This breaks the node-postgres cursor/streaming stack deterministically; `pg-cursor`, `pg-query-stream`, `knex .stream()`, and anything layered on them (TypeORM/Sequelize streaming). They open with `Parse/Bind/Describe/Flush` and no `Execute`, which is legal: it reads the `RowDescription` before deciding to pull rows.

Closes #1330

## Root cause

`Flush` and `Sync` were treated as the same boundary. They aren't:

- `Flush` = push pending output; the batch continues.
- `Sync`  = the batch is over; close the implicit transaction, drop portals, send `ReadyForQuery`.

## Fix

Reuse the existing pin mechanism (advisory locks, manual pin) with a third reason: the client is mid extended-protocol unit.

## Behavior changes

- A client holding an open portal now holds its backend until `Sync`. Same exposure as a client that sends `BEGIN` and goes idle; `client_idle_in_transaction_timeout` governs it (default infinite, unchanged).
- Cursor traffic now appears in `sv_locked`. Accurate, but new in that metric.

## Tests

Unit;  `opens_portal()`: `P B D H` -> true, `P D H` -> false, `E H` -> true, `B E S` -> false, `S` -> false.

Integration (`frontend::client::test`), red before this change with the reported `34000`:

- `test_flush_after_bind_keeps_portal` — the pg-cursor opening sequence
- `test_flush_portal_paging_stays_pinned` — `Execute/Flush` paging; pin survives several batches
- `test_flush_portal_released_on_terminate` — client walks away mid-batch, backend returns to the pool
- `test_parse_describe_flush_bind_execute_close_sync` — now asserts `server_assignment_count == 2`, locking in that the Parse-only path still releases

### DIFF
--- a/pgdog/src/frontend/client/query_engine/lock.rs
+++ b/pgdog/src/frontend/client/query_engine/lock.rs
@@ -4,9 +4,9 @@ impl QueryEngine {
     /// Check if we need to lock the backend to this client, and do so
     /// if needed.
     pub(super) fn check_lock(&mut self) {
-        // The presence of advisory locks or manual pin or open portals
+        // The presence of advisory locks or manual pin
         // indicates we cannot release the backend.
-        let locked = self.advisory_locks.locked() || self.manual_lock || self.portal_open;
+        let locked = self.advisory_locks.locked() || self.manual_lock;
 
         self.backend.lock(locked);
         self.stats.locked(locked);
--- a/pgdog/src/frontend/client/query_engine/lock.rs
+++ b/pgdog/src/frontend/client/query_engine/lock.rs
@@ -4,9 +4,9 @@ impl QueryEngine {
     /// Check if we need to lock the backend to this client, and do so
     /// if needed.
     pub(super) fn check_lock(&mut self) {
-        // The presence of advisory locks or manual pin
+        // The presence of advisory locks or manual pin or open portals
         // indicates we cannot release the backend.
-        let locked = self.advisory_locks.locked() || self.manual_lock;
+        let locked = self.advisory_locks.locked() || self.manual_lock || self.portal_open;
 
         self.backend.lock(locked);
         self.stats.locked(locked);
--- a/integration/js/pg_tests/package-lock.json
+++ b/integration/js/pg_tests/package-lock.json
@@ -11,6 +11,7 @@
       "dependencies": {
         "@prisma/client": "^6.5.0",
         "pg": "^8.14.1",
+        "pg-cursor": "^2.22.0",
         "pg-hstore": "^2.3.4",
         "postgres": "^3.4.7",
         "sequelize": "^6.32.1"
@@ -1932,6 +1933,15 @@
       "resolved": "https://registry.npmjs.org/pg-connection-string/-/pg-connection-string-2.7.0.tgz",
       "integrity": "sha512-PI2W9mv53rXJQEOb8xNR8lH7Hr+EKa6oJa38zsK0S/ky2er16ios1wLKhZyxzD7jUReiWokc9WK5nxSnC7W1TA==",
       "license": "MIT"
+    },
+    "node_modules/pg-cursor": {
+      "version": "2.22.0",
+      "resolved": "https://registry.npmjs.org/pg-cursor/-/pg-cursor-2.22.0.tgz",
+      "integrity": "sha512-knzXLKqarTjOvb3qDSW0JiGsazmxwEKXrqHfWRte7XUsOYccQRafn3BLnQobWwInkzFJSyOej8y8cQRh2z3kGw==",
+      "license": "MIT",
+      "peerDependencies": {
+        "pg": "^8"
+      }
     },
     "node_modules/pg-hstore": {
       "version": "2.3.4",
--- a/integration/js/pg_tests/package.json
+++ b/integration/js/pg_tests/package.json
@@ -14,8 +14,9 @@
   "dependencies": {
     "@prisma/client": "^6.5.0",
     "pg": "^8.14.1",
-    "postgres": "^3.4.7",
+    "pg-cursor": "^2.22.0",
     "pg-hstore": "^2.3.4",
+    "postgres": "^3.4.7",
     "sequelize": "^6.32.1"
   },
   "devDependencies": {
--- a/integration/js/pg_tests/test/cursor.js
+++ b/integration/js/pg_tests/test/cursor.js
@@ -1,0 +1,98 @@
+import pg from "pg";
+import Cursor from "pg-cursor";
+import assert from "assert";
+const { Client } = pg;
+
+// https://github.com/pgdogdev/pgdog/issues/1330
+//
+// pg-cursor opens with Parse/Bind/Describe/Flush and no Execute. In
+// transaction mode that used to release the backend before the portal was
+// used, and the follow-up Execute failed with 34000.
+//
+// pg-query-stream and knex .stream() are built on pg-cursor.
+
+const URL = "postgres://pgdog:pgdog@127.0.0.1:6432/pgdog";
+
+// Always end the client, otherwise a failure leaves mocha hanging on an
+// open handle instead of reporting.
+async function withClient(fn) {
+  const client = new Client(URL);
+  await client.connect();
+  try {
+    return await fn(client);
+  } finally {
+    await client.end();
+  }
+}
+
+it("reads the first page of a cursor", async () => {
+  const rows = await withClient(async (client) => {
+    const cursor = client.query(new Cursor("SELECT $1::bigint AS one", [1]));
+    const page = await cursor.read(10);
+    await cursor.close();
+    return page;
+  });
+
+  assert.equal(rows.length, 1);
+  assert.equal(rows[0].one, "1");
+});
+
+it("pages through a cursor without losing or repeating rows", async () => {
+  const seen = await withClient(async (client) => {
+    const cursor = client.query(
+      new Cursor("SELECT i FROM generate_series(1, 10) AS i ORDER BY i"),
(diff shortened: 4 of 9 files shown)


levkk:
Nice. Could you add some node tests as well to make sure this works end-to-end?

nsavvide:
> Nice. Could you add some node tests as well to make sure this works end-to-end?

`integration/js/pg_tests/test/cursor.js`, added to the existing mocha suite. Uses `pg-cursor` 2.22; the same version
 reported, which is what `pg-query-stream` and `knex .stream()` are built on.

Against a pre-fix binary, 3 of the 4 fail with the reported error, including
the portal naming from the issue:

  1) reads the first page of a cursor:
     error: portal "C_1" does not exist
  2) pages through a cursor without losing or repeating rows:
     error: portal "C_2" does not exist
  3) runs two cursors in sequence on the same client:
     error: portal "C_3" does not exist
     
c3a5a73d60f4c7fe1ff960d2558d609ccc59ea01

pgdog/src/frontend/client/query_engine/lock.rs
```diff
@@ -4,9 +4,9 @@ impl QueryEngine {
     /// Check if we need to lock the backend to this client, and do so
     /// if needed.
     pub(super) fn check_lock(&mut self) {
-        // The presence of advisory locks or manual pin
+        // The presence of advisory locks or manual pin or open portals
         // indicates we cannot release the backend.
-        let locked = self.advisory_locks.locked() || self.manual_lock;
+        let locked = self.advisory_locks.locked() || self.manual_lock || self.portal_open;
```
levkk: This is semantically incorrect I think. We use locked for session state exlusively. I think we're okay not showing cursor locks at all in stats since they are just long transactions. Also, if we use `locked` here, we'll be grabbing the pool mutex quite a bit if the client uses cursors, which could affect performance.

levkk:
Almost good to go, just would be good to remove the `locked` change since it's not actually fit for this particular purpose.

pgdog/src/frontend/client/query_engine/lock.rs
nsavvide: Makes sense, I could switch over to guarding the release
directly:

```rust
// query.rs, cleanup_backend
if self.backend.done() && !self.portal_open {
```

That drops the lock.rs change and the extra check_lock() I'd added before
cleanup_backend (it only existed to re-apply the lock after connect(),
since at handle() time the binding is still NotConnected). sv_locked
stays session-state only, and Connection::done() is untouched, so
can_disconnect()/shutdown behave as before.

One thing I'd like your call on. `lock(true)` also called `dirty()`, so an
abandoned cursor got `RESET ALL`/`DISCARD TEMP` on check-in. Without it: a client that disconnects mid-cursor hits `Cleanup::new` with `in_transaction == false` (no ReadyForQuery was ever seen, so our flag is stale) and
`dirty == false`, and the server goes back into the pool with a live portal
and an open implicit transaction, showing as idle in transaction until whoever gets it next sends a Sync.

**Q: Cheapest fix is one call on the release-with-open-portal path rather than the per-message dirty(). Prefer marking it dirty there, or force-closing?**

nsavvide: > Makes sense, I could switch over to guarding the release directly:
> 
> ```rust
> // query.rs, cleanup_backend
> if self.backend.done() && !self.portal_open {
> ```
> 
> That drops the lock.rs change and the extra check_lock() I'd added before cleanup_backend (it only existed to re-apply the lock after connect(), since at handle() time the binding is still NotConnected). sv_locked stays session-state only, and Connection::done() is untouched, so can_disconnect()/shutdown behave as before.
> 
> One thing I'd like your call on. `lock(true)` also called `dirty()`, so an abandoned cursor got `RESET ALL`/`DISCARD TEMP` on check-in. Without it: a client that disconnects mid-cursor hits `Cleanup::new` with `in_transaction == false` (no ReadyForQuery was ever seen, so our flag is stale) and `dirty == false`, and the server goes back into the pool with a live portal and an open implicit transaction, showing as idle in transaction until whoever gets it next sends a Sync.
> 
> **Q: Cheapest fix is one call on the release-with-open-portal path rather than the per-message dirty(). Prefer marking it dirty there, or force-closing?**

Finding a solution for this right now....

nsavvide: 47a44873ec8bcb1ca7cb1ca7b1f593e86930a659

nsavvide:
@levkk I have taken the alternative route, making the solution simpler really.

levkk:
Aaah! Beautiful.

PR #1333, opened 2026-08-10
fix: include sv_idle_xact in server connections dashboard formula

zacharywelch:
The formula alias already included `sv_idle_xact`, but the underlying formula did not.


levkk (approved):
Good catch, thank you.

PR #1335, opened 2026-08-11
perf(prepared_statements): lower number of allocations

meskill:
Little performance improvements, performance comments and more explicit names for prepared statements.

- avoid allocations on cleaning
- avoid cloning for cache_key (yes, there are Bytes clone that is arc::clone, but still)
- when creating new CacheKey to put into cache make sure we consume only necessary amount of memory and now the memory occupied by String::capacity
- more realistic memory_usage calculation.

The perf improvements are pure vibes - I don't have numbers.

### DIFF
--- a/pgdog/src/frontend/prepared_statements/global_cache.rs
+++ b/pgdog/src/frontend/prepared_statements/global_cache.rs
@@ -22,7 +22,6 @@ pub struct Statement {
     #[allow(dead_code)]
     version: usize,
     cache_key: CacheKey,
-    evict_on_close: bool,
 }
 
 impl MemoryUsage for Statement {
@@ -35,7 +34,6 @@ impl MemoryUsage for Statement {
                 0
             }
             + self.cache_key.memory_usage()
-            + self.evict_on_close.memory_usage()
     }
 }
 
@@ -44,8 +42,8 @@ impl Statement {
         self.parse.query()
     }
 
-    fn cache_key(&self) -> CacheKey {
-        self.cache_key.clone()
+    fn cache_key(&self) -> &CacheKey {
+        &self.cache_key
     }
 }
 
@@ -66,8 +64,8 @@ pub struct CacheKey {
 impl MemoryUsage for CacheKey {
     #[inline]
     fn memory_usage(&self) -> usize {
-        // Bytes refer to memory allocated by someone else.
-        std::mem::size_of::<Bytes>() * 2 + self.version.memory_usage()
+        // The Bytes alias the Parse in Statement, which counts them via Parse::len.
+        std::mem::size_of::<Self>()
     }
 }
 
@@ -124,7 +122,7 @@ impl MemoryUsage for GlobalCache {
             + self.names.memory_usage()
             + self.counter.memory_usage()
             + self.versions.memory_usage()
-            + self.unused.len() * std::mem::size_of::<usize>()
+            + self.unused.capacity() * 1usize.memory_usage()
     }
 }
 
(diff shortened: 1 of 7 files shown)


levkk (approved):

pgdog/src/frontend/prepared_statements/global_cache.rs
```diff
@@ -317,27 +318,42 @@ impl GlobalCache {
...
+        let removed = unused
+            .extract_if(|counter| {
+                // PERF: the global_name always allocates
+                // do we need to actually store this by String or
+                // can we use buffer
+                self.remove(&global_name(*counter));
+
+                true
+            })
+            .take(over)
```
levkk, line 340: Pretty interesting use of an iterator, took me a minute to understand. I guess that's technically legal, but it's interesting to see iterators having side-effects and us relying on only some of it to be consumed.

pgdog/src/frontend/prepared_statements/global_cache.rs
```diff
@@ -150,7 +148,12 @@ impl GlobalCache {
         } else {
             self.counter += 1;
             let name = global_name(self.counter);
-            let parse = parse.rename(&name);
+            // PERF: we explicitly create the new parse with renamed
+            // to reallocate the data and not to hold original buffer
+            // from pgdog/src/frontend/client/mod.rs
+            // smaller memory footprints and smaller allocations, since
```
levkk, line 154: That's correct. Additionally, I would mention buffer fragmentation. The client will continue to send more messages later and which will monotonically grow the buffer since it won't be able to de-allocate memory in the middle of it.

pgdog/src/frontend/prepared_statements/global_cache.rs
```diff
@@ -317,27 +318,42 @@ impl GlobalCache {
...
+        // PERF: the remove though removes once at a time in the loop, that could
+        // defeat this optimization actually
+        let mut unused = std::mem::take(&mut self.unused);
+
+        let removed = unused
+            .extract_if(|counter| {
+                // PERF: the global_name always allocates
+                // do we need to actually store this by String or
+                // can we use buffer
+                self.remove(&global_name(*counter));
```
levkk, line 336: I saw a trick for this in our new lookup cache: https://github.com/pgdogdev/pgdog/blob/bafec81f1898e96d4a419d4dfe711b7d090b5566/pgdog/src/frontend/router/sharding/lookup.rs#L82

pgdog/src/frontend/prepared_statements/global_cache.rs
meskill: the [extract_if](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.extract_if) is one big side-effect 😄

pgdog/src/frontend/prepared_statements/global_cache.rs
meskill: oh, interesting.
yeah, I also have been doing something like this here https://github.com/pgdogdev/pgdog/blob/f2ae21ce7239be7b7f4208b1750bae1b29f83b62/pgdog-config/src/sharding.rs#L254-L275 

only without the traits with one explicit ref type

meskill: but, here is another case - we don't have this string anywhere and it's created based on counter.

I was thinking:
- do we actually need to store the string here, since it's could be inferred anyway, but probably it'll just move the allocations somewhere else where we'll need the string
- create a buffer that will allocate once and rewrite it in the loop to create key

levkk: Yeah I agree, I think if we use the counter as key for the `names` cache, that would be great and we could switch it to `fnv::HashMap` instead (which is much faster than hashbrown). However, we would need to modify all callers, e.g.: 

- `insert_row_description`: https://github.com/pgdogdev/pgdog/blob/bafec81f1898e96d4a419d4dfe711b7d090b5566/pgdog/src/frontend/prepared_statements/global_cache.rs#L227
- `rewrite`: https://github.com/pgdogdev/pgdog/blob/bafec81f1898e96d4a419d4dfe711b7d090b5566/pgdog/src/frontend/prepared_statements/global_cache.rs#L219

and make them carry the counter info. I don't think that's too difficult: add it to `Parse`, that should be in scope everywhere this matters.

pgdog/src/frontend/prepared_statements/global_cache.rs
meskill: updated the comment

pgdog/src/frontend/prepared_statements/global_cache.rs
meskill: I'll try to experiment with this, but it's more places than this - we use the global_name for renaming and than we use this name to get the info from global_cache. I think we should check both the names and the rename allocations and try to optimize it.

PR #1336, opened 2026-08-11
fix: bump pg_raw_parse for macOS link error, add macOS to CI

jkaczman:
Following up on https://github.com/pgdogdev/pg_raw_parse/pull/49, bumping the pg_raw_parse rev here to fix the macOS link error.

Also added a simple CI job that runs cargo build to prevent regressions. It uses macos-26 instead of macos-14 (used by release CI) [since that image will be fully unsupported soon](https://github.com/actions/runner-images/issues/13518).  Follow up PR: https://github.com/pgdogdev/pgdog/pull/1337

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -1,0 +1,16 @@
+name: macos
+on:
+  push:
+    branches:
+      - main
+  pull_request:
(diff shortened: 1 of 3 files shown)


levkk (approved):
Nice!

PR #1337, opened 2026-08-11
ci: bump macOS release runner to macos-26 (deprecation in macos-14)

jkaczman:
macOS 14 runners, which we use in our CI release workflow, [will be fully unsupported in a few months](https://github.com/actions/runner-images/issues/13518). Bumped to use macOS 26 runners.

I also got rid of an unnecessary macOS install dependency step: we don't build openssl-sys, runner images come with pkgconf, and the build uses Xcode's libclang (not llvm).

Verified macOS 26 build success in the release workflow in my fork (running on top of https://github.com/pgdogdev/pgdog/pull/1336 fix): https://github.com/jkaczman/pgdog/actions/runs/31486516587/job/93763057440


levkk (approved):
Cool!

PR #1340, opened 2026-08-11
test(prepared_statements): avoid corrupting PS after setting number of PS to 0

meskill:
it was fixed already by https://github.com/pgdogdev/pgdog/pull/1331 but anyway at-least here is integration-like test for this with the helpers to reuse later

### DIFF
--- a/pgdog/src/backend/server.rs
+++ b/pgdog/src/backend/server.rs
@@ -2809,6 +2809,35 @@ pub mod test {
         }
     }
 
+    /// Names Postgres itself thinks are prepared on this connection.
+    pub(crate) async fn prepared_in_postgres(server: &mut Server) -> Vec<String> {
+        server
+            .fetch_all::<String>("SELECT name FROM pg_prepared_statements")
+            .await
+            .unwrap()
+    }
+
+    /// Run a named statement the way a client would, leaving it to the
+    /// connection to prepare it first if it isn't already.
+    pub(crate) async fn execute_prepared(
+        server: &mut Server,
+        name: &str,
+        param: &[u8],
+    ) -> Vec<i64> {
+        use crate::net::bind::Parameter;
+
+        let request = ServerRequest {
+            messages: vec![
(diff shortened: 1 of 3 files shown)


levkk (approved):

pgdog/src/backend/server.rs
```diff
@@ -2803,6 +2803,31 @@ pub mod test {
         }
     }
 
+    /// Names Postgres itself thinks are prepared on this connection.
+    pub(crate) async fn prepared_in_postgres(server: &mut Server) -> Vec<String> {
```
levkk, line 2813: Let's mark these with `#[cfg(test)]` to avoid confusion.

pgdog/src/backend/server.rs
```diff
@@ -2803,6 +2803,31 @@ pub mod test {
...
+        server
+            .fetch_all::<String>("SELECT name FROM pg_prepared_statements")
+            .await
+            .unwrap()
+    }
+
+    /// Run a named statement the way a client would, leaving it to the
+    /// connection to prepare it first if it isn't already.
+    pub(crate) async fn execute_prepared(server: &mut Server, name: &str) -> Vec<i64> {
```
levkk: Why is this not taking `self`?  And if we're only using this in tests, let's mark it with `cfg(test)` as well, especially since this is using hardcoded values. Alternatively, give the caller the ability to pass in params - that will make this method usable by callers for real.

pgdog/src/backend/server.rs
meskill: it's already inside the mod test

levkk: Ah my bad

pgdog/src/backend/server.rs
meskill: that is also inside the mod test, no self to not mess with the usual impl, and I added the single param to the function so it could be used in more context if needed.

overall, this function is to run the existing prepared when the actual prepared was managed outside, so it's possible to mess around with actual caches etc.

PR #1342, opened 2026-08-11
fix(prepared_statements): make the old PREPARE unused and allow memory reclaim

meskill:
Make sure the overwrite the PREPARED doesn't leak the statement in global cache. 

And other refactors of global_cache to make it more explicit how it works for simple queries

### DIFF
--- a/pgdog/src/frontend/prepared_statements/global_cache.rs
+++ b/pgdog/src/frontend/prepared_statements/global_cache.rs
@@ -8,19 +8,20 @@ use std::{collections::hash_map::HashMap, str::from_utf8};
 
 use fnv::FnvHashSet as HashSet;
 
+/// Identity of a prepared statement inside the global cache.
+pub type Counter = usize;
+
 // Format the globally unique prepared statement
 // name based on the counter.
-fn global_name(counter: usize) -> String {
+fn global_name(counter: Counter) -> String {
     format!("__pgdog_{}", counter)
 }
 
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Clone)]
 pub struct Statement {
     parse: Parse,
     rewrite: Option<Parse>,
     row_description: Option<RowDescription>,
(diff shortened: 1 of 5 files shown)


levkk (approved):

pgdog/src/frontend/prepared_statements/global_cache.rs
```diff
@@ -184,21 +197,29 @@ impl GlobalCache {
 
     /// Insert a prepared statement into the global cache ignoring
     /// duplicate check.
+    ///
+    /// SQL `PREPARE` gets a key of its own, so it is never handed to
+    /// a second client. It is tracked and evicted like any other statement.
     pub fn insert_anyway(&mut self, parse: &Parse) -> String {
```
levkk: You may want to rename this method accordingly. `insert_anyway` implies bypassing the cache, it does not imply handling prepared statements sent over the simple protocol only.

pgdog/src/frontend/prepared_statements/global_cache.rs
```diff
@@ -101,6 +114,8 @@ impl CachedStmt {
 ///
 /// 1. Mapping between unique prepared statement identifiers (query and result data types),
 ///    and the global unique prepared statement name used in all server connections.
+///    Statements created by SQL `PREPARE` carry a key of their own, so they are
```
levkk, line 117: This is not a great implementation on our part. We should make those re-usable between clients.

pgdog/src/frontend/prepared_statements/global_cache.rs
meskill: renamed to `insert_prepared`

pgdog/src/frontend/prepared_statements/global_cache.rs
meskill: https://github.com/pgdogdev/pgdog/issues/1350

PR #1347, opened 2026-08-12
fix: restore REPLICA IDENTITY after indexes in the schema sync

rlittlefield:
`ALTER TABLE ... REPLICA IDENTITY ...` statements from pg_dump fall
into the schema sync's catch-all arm, which restores them in the
pre-data step. The `USING INDEX` form references an index that is only
created in the post-data step, so the statement fails, and a restore
running with errors ignored swallows the failure. The table lands on
the destination with its default replica identity and nothing reports
that the schema diverged.

We hit this on a cluster where a migration had set
`REPLICA IDENTITY USING INDEX` on every shard: a shard provisioned
through the schema sync silently came up without it, and only showed up
later when logical replication needed the identity columns.

The fix restores replica identity statements in the post-data step,
after indexes exist. pg_dump emits `REPLICA IDENTITY USING INDEX`
right after the index it references and statement order is preserved
within a step, so the index is always in place first. The `FULL`,
`NOTHING` and `DEFAULT` forms don't reference an index and work in
either step, so they move along with it.

The new test covers both forms: no replica identity statement restores
in pre-data, both restore in post-data, and `USING INDEX` restores
after its index.

### DIFF
--- a/pgdog/src/backend/schema/sync/pg_dump.rs
+++ b/pgdog/src/backend/schema/sync/pg_dump.rs
@@ -770,6 +770,20 @@ impl PgDumpOutput {
                                 }
 
                                 nodes::AlterTableType::AT_AddIdentity => (),
+
+                                // REPLICA IDENTITY USING INDEX references an
+                                // index, which is created in the post-data
+                                // step. In pre-data it fails, and a restore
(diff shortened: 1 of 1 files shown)


levkk (approved):
Excellent thank you

PR #1348, opened 2026-08-12
fix: return every selected column in a cross-shard GROUP BY

### LINKED ISSUE #1172: [Query engine] Aggregate functions mixed with columns don't work
**PgDog version**
`v0.1.48`

**Description**
Cross-shard aggregates are not returned correctly when aggregate functions are mixed with columns, e.g.:

```sql
SELECT org_id, id, extra, max(val) FROM t GROUP BY org_id, id;
```

Which returns multiple rows:

```
org_id | id | extra | max
      2 |  1 |       |  20
      3 |  1 |       |  30
      1 |  1 |       |  10
```

**Logs**
Attach trace logs, if the issue can be reproduced locally. To get trace logs, set `RUST_LOG=trace`, e.g.:

```bash
2026-07-09T20:18:11.984124Z DEBUG SELECT org_id, id, extra, max(val) FROM t GROUP BY org_id, id;
2026-07-09T20:18:11.984145Z DEBUG table is sharded, but no sharding key detected
2026-07-09T20:18:11.984151Z DEBUG query router decision: Query(
    Route {
        shard: ShardWithPriority {
            source: Table(
                Sharded,
            ),
            shard: All,
        },
        read: true,
        order_by: [],
        aggregate: Aggregate {
            targets: [
                AggregateTarget {
                    column: 3,
                    function: Max,
                    distinct: false,
                },
            ],
            group_by: [
                0,
                1,
            ],
        },
        limit: Limit {
            limit: None,
            offset: None,
        },
        advisory_locks: AdvisoryLocks {
            locks: {},
        },
        distinct: None,
        rewrite_plan: AggregateRewritePlan {
            helpers: [],
        },
        explain: None,
        rollback_savepoint: false,
        search_path_driven: false,
        schema_changed: false,
        omnisharded: false,
    },
)
```

**Configuration**

Sharded config with matching tables.

jkaczman:
Previously, columns that were neither grouped nor aggregated were nulled in cross-shard `GROUP BY` results. #1172 has an example case. When merging results in `aggregate.rs`, each row was re-built from two kinds of columns: grouping keys and accumulators. Columns that didn't belong to one of those two were left null.

Postgres allows a third kind of column in a grouped query, specifically one that is functionally dependent on the grouping. It [recognizes this ONLY when the table's primary key belongs to `GROUP BY`](https://www.postgresql.org/docs/current/sql-select.html). Which means every row in a group carries the same value for that column, so there's no need to merge them.

Modified to have these extra columns go along with the accumulators within a `GroupState` store, which is mapped against `Grouping`, so that all columns in a row can be accessed when re-constructing. 

A column that isn't functionally dependent (& that doesn't belong to those two cases) can't reach this logic. Postgres will reject the query outright.

Added a regression case (SQL integration).

Fixes #1172.

### DIFF
--- a/integration/sql/cases/944_aggregate_ungrouped_column_case.sql
+++ b/integration/sql/cases/944_aggregate_ungrouped_column_case.sql
@@ -1,0 +1,23 @@
+-- description: Columns that are neither grouped/aggregated will still be included in a cross-shard GROUP BY query
+-- tags: standard sharded
+-- transactional: true
+
+-- The `extra` column isn't grouped nor is it aggregated.
+-- Postgres permits it because the primary key is grouped.
+-- Thus, that makes it functionally dependent, and it must still be returned.
+SELECT org_id, id, extra, max(val) AS max_val
+FROM sql_regression_samples
+GROUP BY org_id, id
+ORDER BY id;
+
+-- Do groups spanning shards still merge into one row with correct aggregates?
+SELECT org_id, min(id) AS first_id, count(*) AS occurrences, max(val) AS max_val
+FROM sql_regression_samples
+GROUP BY org_id
+ORDER BY org_id;
+
+-- What about if we don't use an aggregate at all?
+SELECT org_id, id, extra
+FROM sql_regression_samples
+GROUP BY org_id, id, extra
+ORDER BY id;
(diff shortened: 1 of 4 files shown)


levkk (approved):
Really cool!

PR #1353, opened 2026-08-12
Minor cleanup of TargetHealth

sgrif:
Fixed visibility, removed dead code, replaced cloned value with an accessor method since we always have access to the original


levkk (approved):
Noice. Ignore CI, github actions have been on fire for weeks with intermittent network issues.

PR #1356, opened 2026-08-13
fix(state): don't wait for a ReadyForQuery the server will never send

### LINKED ISSUE #1314: Savepoint mixed with opened portal breaks state on portal execution error
**PgDog version**
`v0.1.49`, `v0.1.51` & `main`

**Description**
When running Elixir's Postgrex and using its `mode: :savepoint` feature, Elixir side times out while PgDog side leaves the connection in a `wait_event='ClientRead'` transaction.

---

Here's the simplest repro, you can run it with `elixir repro.exs`:

```elixir
Mix.install([{:postgrex, "~> 0.19"}])

{:ok, conn} =
  Postgrex.start_link(
    hostname: "127.0.0.1",
    port: 6432,
    username: "postgres",
    password: "postgres",
    database: "postgres"
  )

result =
	Postgrex.transaction(
	  conn,
	  fn c -> Postgrex.query(c, "SELECT 1/0", [], mode: :savepoint) end,
	  timeout: 5_000
	)

dbg(result)
```

then verify the hanging transaction:

```sql
SELECT pid, state, wait_event, query FROM pg_stat_activity WHERE wait_event = 'ClientRead'
```

---

Postgrex is doing 
```
Query("BEGIN")
Query("SAVEPOINT postgrex_query")
Bind
Execute
Query("RELEASE SAVEPOINT postgrex_query")
```

it doesn't do `Sync`, which seems to be what trips PgDog here, as `ErrorResponse` handling re-queues `ReadyForQuery` regardless if `Sync` was sent. Best I can tell, what Postgrex does is unique but legal.

**Logs**

[pgdog-trace-full.log](https://github.com/user-attachments/files/30754508/pgdog-trace-full.log)

**Configuration**

The example Docker compose configuration from the repo. Tested in both sharded and unsharded modes just in case.

jkaczman:
A client using Postgrex's `mode: :savepoint` hangs on any query that errors. The error is returned to the client, but the client never gets the `ReadyForQuery` that lets it show the error to the caller. It hangs until reaching timeout.

Postgrex sends `Bind` (extended), `Execute` (extended), and a **simple** `Query("RELEASE SAVEPOINT ...")` as one batch. It doesn't include a Sync **in that batch**, it does afterwards though, as part of error recovery.

When the `Execute` errors, the server gets rid of all messages until a Sync comes. This includes the RELEASE query previously mentioned. However, due to the mixed nature of the batch (extended, extended, simple), PgDog expects a ReadyForQuery back...

PgDog queued the same expectation (`Z`) for **both** Sync-owed and Query-owed RFQ, so our error branch in `state.rs's action` method couldn't tell them apart. Thus, it re-queued the trailing RFQ unconditionally, and that being the only entry that survived kept `has_more_messages()` true in `server.rs`, so PgDog stayed infinitely stuck on the server socket, and never read the Sync the client had already sent.

Essentially a Deadlock: client waits on PgDog, PgDog waits on server, server waits on a Sync (which is sitting unread in PgDog's client socket).

PgDog 1314

---

What I changed:
- Differentiated a Query-owed vs Sync-owed RFQ into two variants: `ReadyForQuery` and `ReadyForQuerySync`. After an extended-protocol error, the server only ever answers the Sync, so we need a way to distinguish them.
- Modified the error branch to classify by the front queue entry, which belongs to the request the server was processing. This is the same test the server itself uses to decide whether or not to skip to a Sync.
- On an extended error, expectations are dropped up to the first Sync-owed RFQ. If no Sync is queued yet (this bug!), the queue drains and PgDog goes back to the client, where it can now receive the Sync.
- Removed the sticky extended flag that represented the connection seeing extended traffic, not that the error occurred in an extended message.
- Split old `'C' | 's' | 'I'` lump into `CommandComplete` and `ExecutionCompleted`. `CommandComplete` can come from a simple Query or from an extended Execute. An Execute one means error skips to a sync, so we need to differentiate there too.

---

Tests: added two state.rs regression tests, three new Elixir integration tests, did some small updates to some state.rs tests to update to the new variants.

Fixes #1314

### DIFF
--- a/integration/elixir/test/savepoint_test.exs
+++ b/integration/elixir/test/savepoint_test.exs
@@ -1,0 +1,50 @@
+defmodule Pgdog.SavepointTest do
+  use ExUnit.Case, async: false
+
+  # Regression tests for https://github.com/pgdogdev/pgdog/issues/1314.
+  #
+  # Postgrex's mode: :savepoint wraps each query in a savepoint by sending
+  # Bind, Execute and a simple Query("RELEASE SAVEPOINT postgrex_query") in one batch (no Sync in that batch)
+  #
+  # When that Execute errors, the server discards everything until a Sync arrives,
+  # so the Query's ReadyForQuery is never produced.
+  #
+  # PgDog used to keep waiting for it and never read the Sync the client had already sent
+
+  setup do
+    %{conn: Pgdog.connect()}
+  end
+
+  test "query error under mode: :savepoint returns instead of hanging", %{conn: conn} do
+    {:ok, {:error, error}} =
+      Postgrex.transaction(
+        conn,
+        fn c -> Postgrex.query(c, "SELECT 1/0", [], mode: :savepoint) end,
+        timeout: 5_000
+      )
+
+    assert %Postgrex.Error{postgres: %{code: :division_by_zero}} = error
+  end
+
+  test "connection survives a savepoint-mode error", %{conn: conn} do
+    {:ok, {:error, _}} =
+      Postgrex.transaction(
+        conn,
+        fn c -> Postgrex.query(c, "SELECT 1/0", [], mode: :savepoint) end,
+        timeout: 5_000
+      )
+
+    assert Pgdog.one(Postgrex.query!(conn, "SELECT 1::bigint", [])) == 1
+  end
+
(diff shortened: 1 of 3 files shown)


levkk (approved):
Very cool!

PR #1357, opened 2026-08-13
Remove uneccessary AST copy

sgrif:
We were previously parsing an AST only to immediatley copy it onto a new memory arena and throw the original away. I've added a function on the parser side to allow us to skip this step. We can start using it now


levkk (approved):
Oh neat! This might even register on a benchmark

sgrif:
It's probably pretty minor, but yeah cutting ~50% of the allocations should be a visible change

levkk:
The trick is testing it with a big query: 


image

levkk:
I'm going to merge this optimization to include it in today's release. Let me know if there is anything else needed here.

PR #1363, opened 2026-08-14
Skip replica-identity validation for sync-only data-sync

bhargavtheertham-cb:
`data_sync` validates that every published table has a usable replica identity (a primary key or a `REPLICA IDENTITY` index) before it starts copying. This prevents all data syncs, including `--sync-only`.

That requirement only applies to the streaming phase, which uses the replica identity to build the per-row `UPDATE`/`DELETE` filters. The initial `COPY` never uses it, and a `--sync-only` run never streams.

## Change

Add a `require_replica_identity` flag to `Publisher::data_sync` (threaded through `Orchestrator::data_sync` and `CopyDataTask`). `ReshardTask` sets it to `!sync_only`:

- **Online** (default): unchanged — tables are still validated up front, before the potentially multi-hour copy.
- **Sync-only**: validation is skipped, so tables without a primary key or replica identity can be copied.

### DIFF
--- a/pgdog/src/api/copy_data.rs
+++ b/pgdog/src/api/copy_data.rs
@@ -15,6 +15,9 @@ use crate::backend::replication::logical::orchestrator::Orchestrator;
 #[display("copy_data {orchestrator}")]
 pub(crate) struct CopyDataTask {
     pub orchestrator: Orchestrator,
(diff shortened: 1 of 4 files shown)


levkk (approved):
Perfect, thank you!

PR #1364, opened 2026-08-14
fix(pool): extended pipeline mixed messages for multi-shard

meskill:
Fix the issues with multi-shard messages aggregation when multiple parse/describes commands sent in the pipeline.

- proper handle for RowDescription counter to reset state and forward it to the client
- reorganize saving of RowDescription in state to pick the proper RD for validation and decoder
- make sure the decoder gets proper RowDescription and Bind codes when processing the response rows


meskill:
The unit tests and go test should showcase the issue better. The issue appears on multi-shard configuration for requests that try to aggregate things (pgdog/src/backend/pool/connection/multi_shard/mod.rs). 

I have doubts about quality of the fix and the code overall in multi_shard and I believe there are other possible issues with it (like ErrorResponse handling, portals handling, anonymous commands), but I hope to fix some of the problems rn and maybe revisit and overhaul the whole flow completely.

levkk (approved):
Very nice find

PR #1365, opened 2026-08-14
fix(cross_shard_disabled): provide more specific error msg when dealing with unknown sharding key

### LINKED ISSUE #1163: Warn about unknown sharding key rather than assume a cross-shard query
This concerns the behavior when `cross_shard_disabled = true`.

If an unknown sharding key is used (with SET for example), the parser will throw the same error as if no sharding key was entered at all. It's not critical, but it would be nice if the error message thrown would indicate that the sharding key does not exist, to differentiate the two scenarios.

jkaczman:
Provides a specific error message "unknown sharding key was specified" as opposed to "cross-shard queries are disabled" when a query:
1. has an unknown sharding key set (list-based / range-based)
2. cross_shard_disabled=true 
3. would be cross-shard if not stopped

Fixes #1163

### DIFF
--- a/pgdog/src/frontend/client/query_engine/query.rs
+++ b/pgdog/src/frontend/client/query_engine/query.rs
@@ -388,21 +388,11 @@ impl QueryEngine {
         // should be cross-shard (e.g. BEGIN, COMMIT) but aren't really.
         if connected_shards == 0 || connected_shards > 1 {
             let query = context.client_request.query()?;
-
-            let error = if let Some(sharding_key) = context.params.get("pgdog.sharding_key")
-                && let Some(sharding_key_value) = sharding_key.as_str()
-            {
-                // The user specified a sharding key, which was determined to be unknown.
-                // This was done in list-based/range-based sharding.
-                // If not stopped, the query would be cross-shard (due to the unknown key).
-                // Rather than give them a generic 'cross shard disabled' error,
-                // tell them the sharding key is unknown.
-                ErrorResponse::unknown_sharding_key_in_cross_shard_disabled(sharding_key_value)
-            } else {
-                ErrorResponse::cross_shard_disabled(query.as_ref().map(|q| q.query()))
-            };
-
-            self.error_response(context, error).await?;
+            self.error_response(
+                context,
+                ErrorResponse::cross_shard_disabled(query.as_ref().map(|q| q.query())),
+            )
+            .await?;
 
             if self.backend.connected() && self.backend.done() {
                 self.backend.disconnect();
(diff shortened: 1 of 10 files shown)


pgdog/src/net/messages/error_response.rs
```diff
@@ -86,6 +86,21 @@ impl ErrorResponse {
...
 
+    // Cross-shard queries are disabled.
+    // User specified an unknown sharding key in list-based/range-based sharding,
+    // and, if not stopped, the query would be cross-shard.
+    pub fn unknown_sharding_key_in_cross_shard_disabled(sharding_key: &str) -> ErrorResponse {
+        ErrorResponse {
+            severity: "ERROR".into(),
+            code: "58000".into(),
+            message: "unknown sharding key was specified".into(),
+            detail: Some(format!("sharding key {} is unknown", sharding_key)),
```
levkk:
```suggestion
            detail: Some(format!("sharding key '{}' is not mapped", sharding_key)),
```

pgdog/src/frontend/client/query_engine/query.rs
```diff
@@ -389,7 +388,19 @@ impl QueryEngine {
         // should be cross-shard (e.g. BEGIN, COMMIT) but aren't really.
         if connected_shards == 0 || connected_shards > 1 {
             let query = context.client_request.query()?;
-            let error = ErrorResponse::cross_shard_disabled(query.as_ref().map(|q| q.query()));
+
+            let error = if let Some(sharding_key) = context.params.get("pgdog.sharding_key")
```
levkk: I think this is fine for now. A better solution could be to check this inside the parser:

1. If it returns `Shard::All`
2. and `pgdog.sharding_key` is set in params (those are available in the context, see https://github.com/pgdogdev/pgdog/blob/4bafac91f75c89a00c85cdf65f78865dd2bbe8e2/pgdog/src/frontend/router/parameter_hints.rs#L29)
3. and cross-shard queries are disabled (this may be currently missing from the context but can be easily added I think: https://github.com/pgdogdev/pgdog/blob/4bafac91f75c89a00c85cdf65f78865dd2bbe8e2/pgdog/src/frontend/router/parser/context.rs#L21)

then the parser can return this error, which we can catch and send to the client.

The reason this is better is because it avoids connecting to Postgres when we know we can't serve that query because of the missing sharding key mapping.

levkk: Another reason why the parser could be a better place for this is because it'll also catch the comment directive:

```sql
/* pgdog_sharding_key: 25 */ SELECT * FROM users;
```

while this check will only catch the `pgdog.sharding_key` SET parameter.

levkk (approved):
Awesome! We are going to need to organize the checks a little bit next week; it's getting a bit complex. But this looks good!

PR #1368, opened 2026-08-16
feat: partial sharding support for PREPARE and EXECUTE

levkk:
### Prepared statements

Correctly store statements sent with `PREPARE` in the global cache. We use the whole query string with data type specifications as key, while anonymizing the statement name. This is equivalent to using `Parse` as cache key.

```sql
PREPARE __stmt_1(int) AS SELECT $1;
```

becomes

```sql
PREPARE __pgdog_template_name(int) AS SELECT $1;
```

It's then dynamically rewritten to

```sql
PREPARE __pgdog_1(int) AS SELECT $1
```

at execution time.

This is because it's kinda hard to parse the data types out of the string and we need a unique key identifying the prepared statement in the cache.

### Sharding

Handle `EXECUTE` by extracting parameters and passing them to our query router. This works with I think most of our sharding features, i.e., direct-to-shard and cross-shard queries.

#### TODOs

1. Comment-based routing isn't supported. This is because we parse and deparse the original statement to put it in the prepared statement cache, which erases the comment.
2. Query rewrites, e.g., insert split, don't work yet.

### Notes for reviewers

1. Most of the diff in `client/prepared_statements` module is from moving stuff around. There is some diff for handling `PREPARE`, that's where I would focus the review.
2. Simulated messages sent by `Server` now get processed through the server state machine; this was necessary for correct state management for handling `PREPARE` statements; seems to have no bad side effects.


pgdog/src/net/messages/prepare.rs
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    fn message(&self) -> Result<Message, Error> {
+        Ok(Message::new(self.to_bytes()).frontend())
+    }
+
+    fn streaming(&self) -> bool {
+        false
+    }
+}
```
meskill, line 61: is this actually needed?

pgdog/src/net/messages/prepare.rs
```diff
@@ -0,0 +1,74 @@
...
+        let query = self.query();
+        let name = self.name();
+        // This is safe because the statement looks like this:
+        // PREPARE __pgdog_template_name AS [...]
+        // so the template name will always match first.
+        let query = query.replacen(PREPARE_TEMPLATE_NAME, name, 1);
+
+        Query::new(query).to_bytes()
+    }
+}
```
meskill, line 74: nit: I understand why we need the "fake" messages, but I don't like much that we imitate that it's actual message that has byte representation. I'd prefer if we separate this kind of message in the parent enum to make it explicit and do the conversion to the some other real ProtocolMessage to use it's implementation. I.e. instead of ToBytes implement into_query() and call to_bytes on it

pgdog/src/frontend/router/parser/params.rs
```diff
@@ -0,0 +1,132 @@
...
+
+                    Some(ConstValue::Boolean(bool)) => {
+                        let data = Bytes::from((if bool { "t" } else { "f" }).to_string());
+                        BindParameter {
+                            len: data.len() as i32,
+                            data,
+                        }
+                    }
+
+                    _ => BindParameter::new_null(),
```
meskill, line 57: I think we should handle TypeCast as well and not only ConstValue, so `"445-45::uuid"` works and the `compute_shard_for_table` won't trigger broadcasting.
I think there also could FuncCalls and this probably should be revisited later on routing sharding, erroring etc., but anyway null is too destructive for this and misleading

pgdog/src/frontend/prepared_statements/cache_key.rs
```diff
@@ -0,0 +1,53 @@
...
+///
+/// If two `Extended` keys match, it's effectively the same statement.
+/// If they don't, e.g. client sent the same query but
+/// with different data types, we can't re-use it and
+/// need to plan a new one.
+///
+/// A `Simple` key comes from SQL `PREPARE` and matches nothing but itself.
+/// Its declared argument types are not captured, so two of those
+/// statements are never known to be the same.
+///
```
meskill, line 15: that's not true anymore

pgdog/src/frontend/router/parser/query/prepare.rs
```diff
@@ -0,0 +1,80 @@
...
+        &mut self,
+        stmt: &ExecuteStmt,
+        context: &mut QueryParserContext<'_>,
+    ) -> Result<Command, Error> {
+        // Extract parameters from `EXECUTE` statement.
+        let params = ExecuteParams::new(stmt);
+        let stmt_params = StatementParameters::Execute(&params);
+
+        // Create new parser context.
+        let mut context = context.clone();
```
meskill, line 46: that's looks fishy - we clone to mutate and then after mutation we drop this. I don't know maybe it's covered somehow or not harmful, but that is very brittle. I noticed only because the derive(Clone) appeared on the Context

pgdog/src/frontend/router/parser/rewrite/statement/offset.rs
meskill, line 147: oh, I guess that won't work for execute queries - we don't rewrite the PREPARE this way and the offset won't work properly

pgdog/src/frontend/prepared_statements/global_cache.rs
```diff
@@ -440,9 +352,9 @@ mod test {
         assert_eq!(entry.used, 0);
         assert!(cache.unused.contains(&1)); // __pgdog_1
 
-        let name = cache.insert_prepare(&parse);
-        cache.close(&name);
-        assert!(cache.unused.contains(&2)); // __pgdog_2
+        // let (_, name) = cache.insert_prepare(&parse);
+        // cache.close(&name);
+        // assert!(cache.unused.contains(&2)); // __pgdog_2
```
meskill, line 357: should we restore?

pgdog/src/frontend/prepared_statements/global_cache.rs
```diff
@@ -453,28 +365,30 @@ mod test {
     #[test]
     fn test_simple_prepared_is_never_shared() {
```
meskill, line 366: the name is now wrong, since we check it's shared

pgdog/src/backend/prepared_statements.rs
```diff
@@ -302,15 +311,43 @@ impl PreparedStatements {
...
+                if self.contains(prepare.name()) {
                     return Ok(HandleResult::Drop);
                 } else {
-                    self.parses.push_back(name.clone());
+                    self.parses.push_back(prepare.name().to_string());
                     self.state.add_ignore('C');
 
                     // Prepare turns into a Simple Query ('Q') so it expects a regular RFQ back.
                     self.state.add_ignore(ExecutionCode::ReadyForQuery);
+                    self.parses.push_back(prepare.name().to_owned());
```
meskill, line 350: we pushing name twice in this block

pgdog/src/frontend/prepared_statements/mod.rs
```diff
@@ -49,94 +54,117 @@ impl Default for PreparedStatements {
...
             self.global.write().close(&old_value);
         } else {
             // New entry.
-            self.memory_used += str_mem(key) + str_mem(&name);
+            self.memory_used += str_mem(local) + str_mem(global);
         }
+    }
 
-        parse.rename(&name)
+    /// Insert statement into the cache bypassing duplicate checks.
```
meskill, line 113: so, now, we do the deduplication

pgdog/src/frontend/prepared_statements/global_cache.rs
```diff
@@ -145,107 +50,78 @@ impl GlobalCache {
...
-            self.counter += 1;
-            let name = global_name(self.counter);
-            // PERF: we explicitly create the new parse with renamed
-            // to reallocate the data and not to hold original buffer
-            // from pgdog/src/frontend/client/mod.rs
-            // smaller memory footprints and smaller allocations, since
-            // the client buffer is generally bigger than the query.
-            // Holding onto it would also fragment that buffer: the client
-            // keeps appending messages to it and can't free the middle,
-            // so the buffer grows monotonically.
```
meskill, line 171: this comment is still meaningful I believe - the perf notes on renamed doesn't explain anything what pin we avoid here and without the knowledge about what do we free with this it's appealing to drop the new CacheKey and reuse the first one at the start of the function

pgdog/src/frontend/prepared_statements/global_cache.rs
```diff
@@ -145,107 +50,78 @@ impl GlobalCache {
...
-        let name = global_name(self.counter);
-        let parse = parse.renamed(&name);
-        // insert_anyway is used for the simple query PREPARE call
-        // and here the `unique` field based on counter defines
-        // that this statement won't be reused with other clients
-        // i.e. it'll always have `used <= 1` and will be closed
-        // only by the specific client that created it.
-        // The close happens when the client re-uses the same PREPARE
-        // name, or on client disconnect in the close_all call.
-        // TODO: a direct DEALLOCATE won't close it yet.
```
meskill, line 218: we haven't added the DEALLOCATE handling for PREPARE, right? and it'll be closed only when client is disconnected?

pgdog/src/frontend/router/parser/rewrite/statement/mod.rs
```diff
@@ -104,32 +104,41 @@ impl<'a> StatementRewrite<'a> {
...
-            | Node::DeleteStmt(_) => walk::walk(stmt.stmt(), |node| {
-                if let Node::ParamRef(param) = node {
-                    plan.params = plan.params.max(param.number as u16)
-                }
-            }),
-            Node::PrepareStmt(_) | Node::ExecuteStmt(_) | Node::ExplainStmt(_) => {}
+            | Node::DeleteStmt(_) => Some(node),
+            Node::PrepareStmt(prepare) => {
+                // Will use parameters for replacing args, not materialize values.
+                self.extended = true;
```
meskill, line 115: won't this cause any issues with the actual extended protocol? yeah, like using `PREPARE` inside Parse 😄

pgdog/src/frontend/router/parser/error.rs
```diff
@@ -111,4 +111,10 @@ pub enum Error {
 
     #[error("unmapped sharding key was specified")]
     UnmappedShardKey(String),
+
+    #[error("prepare statement can only be DML")]
+    PrepareNotDml,
```
meskill, line 116: unused

pgdog/src/backend/prepared_statements.rs
```diff
@@ -302,15 +311,43 @@ impl PreparedStatements {
                     self.state.add('3');
                 }
             }
-            ProtocolMessage::Prepare { name, .. } => {
-                if self.contains(name) {
+            ProtocolMessage::PrepareFromClient(prepare) => {
+                use crate::net::{CommandComplete, ReadyForQuery};
+                if self.contains(prepare.name()) {
```
meskill, line 316: I think we need to use check_prepared at some extend - the ttl functionality and parses deduplication is not working for prepared rn

pgdog/src/frontend/router/parser/rewrite/statement/mod.rs
```diff
@@ -162,6 +171,13 @@ impl<'a> StatementRewrite<'a> {
             self.limit_offset(&select, &mut plan);
         }
```
meskill, line 172: well, should we do something specific for the prepare/execute here?

pgdog/src/backend/server.rs
```diff
@@ -569,7 +569,7 @@ impl Server {
             if let Some(message) = self.prepared_statements.state_mut().get_simulated() {
                 // INVARIANT: omni dedup in multi_shard relies on this being process-unique;
                 // never substitute a non-unique value here.
-                return Ok(message.backend(self.id));
+                break message.backend(self.id);
```
meskill, line 572: why this change? that seems like simulated prepared will now be processed and setting `sync_prepared=true` that will cause more work on check-in

pgdog/src/frontend/router/parser/rewrite/statement/mod.rs
levkk: Oh god I hope Postgres triggers an error when that happens....

pgdog/src/backend/prepared_statements.rs
levkk: Good call!

pgdog/src/frontend/router/parser/rewrite/statement/mod.rs
levkk: Ah yeah, probably doesn't work currently.

pgdog/src/backend/server.rs
levkk: Because we need `CommandComplete` and `ReadyForQuery` to reset the server state back to `Idle` when we prepare a statement. Otherwise, it gets stuck in `reading data`. Preparing a statement using `Query` basically executes a separate query before the client's request, so the state management is a bit more complex.

We just push whatever message we "simulate" through the state manager though and it seems to work.

pgdog/src/frontend/prepared_statements/global_cache.rs
levkk: We kinda do: we ignore it. https://github.com/levkk/pgdog/blob/8e7eae9f4aab0766ad0d0452b6cf5c092c6c48df/pgdog/src/frontend/client/query_engine/deallocate.rs#L6

When client prepares a new statement, we just replace it in the client cache mapping.

pgdog/src/backend/prepared_statements.rs
levkk: https://github.com/pgdogdev/pgdog/issues/1382

pgdog/src/frontend/router/parser/rewrite/statement/mod.rs
levkk: https://github.com/pgdogdev/pgdog/issues/1383

pgdog/src/frontend/prepared_statements/global_cache.rs
levkk: We have a comment about this on `Parse::renamed`: https://github.com/levkk/pgdog/blob/8e7eae9f4aab0766ad0d0452b6cf5c092c6c48df/pgdog/src/net/messages/parse.rs#L87

meskill: as I mentioned - this comment just answers how it works, and removed comment answers why and what

pgdog/src/frontend/prepared_statements/global_cache.rs
meskill: yes, but are there any reasons to not clean it up since we do it on Close message?

pgdog/src/frontend/router/parser/rewrite/statement/mod.rs
meskill: this is exaggeration, but messing up with flags is always... risky

pgdog/src/backend/server.rs
meskill: yeah, I think there are also other side-effects that I was worrying about like `sync_prepared` always set to true

pgdog/src/net/messages/prepare.rs
levkk: I think we do convert `to_bytes()` somewhere to send it over the wire, yeah.

pgdog/src/net/messages/prepare.rs
levkk: Yeah, that could be cool.

pgdog/src/frontend/router/parser/params.rs
levkk: Good find: https://github.com/pgdogdev/pgdog/issues/1401

pgdog/src/frontend/router/parser/query/prepare.rs
levkk: Totally fine I think. It just copies the references!

pgdog/src/frontend/prepared_statements/global_cache.rs
levkk: We should yes..

pgdog/src/backend/prepared_statements.rs
levkk: Fixed in a follow up PR.

pgdog/src/frontend/prepared_statements/global_cache.rs
levkk: Yeah fair enough. Maybe we restore it back to the function name; just felt a bit like context overload

pgdog/src/frontend/prepared_statements/global_cache.rs
levkk: Good call, yeah, let's handle it: https://github.com/pgdogdev/pgdog/issues/1402

pgdog/src/net/messages/prepare.rs
meskill: so, only to_bytes and not the highlighted code

pgdog/src/frontend/router/parser/query/prepare.rs
meskill: ehm, no!

you can't clone the &mut and there are reasons the context is mut here. There are Vecs in context that are mutated in the inner calls and that info is just dropped since it's mutations on cloned data that is dropped when function ends

PR #1369, opened 2026-08-16
feat(openmetrics): make host configurable

### LINKED ISSUE #1349: Add possibility to bind OpenMetrics HTTP endpoint to a specific IP address
Currently, the **openmetrics_port** configuration parameter only allows specifying the port for the OpenMetrics HTTP endpoint. pgDog always binds the endpoint to 0.0.0.0, making it listen on all network interfaces.

Add support for binding the OpenMetrics HTTP endpoint to a specific network interface, similar to how the existing **host** parameter works.

For example, introduce an **openmetrics_host** parameter that specifies the interface/address on which the OpenMetrics endpoint should listen.

Alternatively, the OpenMetrics endpoint could automatically bind to the same network interface/address specified by the existing **host** parameter.

### Use Case

Running multiple pgDog instances on the same host, where each instance uses its own virtual IP address.

Each pgDog instance should be able to expose its OpenMetrics endpoint only on its own virtual IP, rather than having all instances bind to 0.0.0.0 and potentially conflict with each other.

murex971:
- Add `openmetrics_host` (default `0.0.0.0`, env `PGDOG_OPENMETRICS_HOST`) 
- Unit test for default + env override (`test_env_openmetrics_host`)

closes #1349

### DIFF
--- a/.schema/pgdog.schema.json
+++ b/.schema/pgdog.schema.json
@@ -76,6 +76,7 @@
         "mirror_exposure": 1.0,
         "mirror_queue": 128,
         "omnisharded_sticky": false,
+        "openmetrics_host": "0.0.0.0",
         "openmetrics_namespace": null,
         "openmetrics_port": null,
         "passthrough_auth": "disabled",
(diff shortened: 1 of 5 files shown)


levkk (approved):
🙏

PR #1370, opened 2026-08-16
Perf trial

ygxio:


ygxio:
@levkk could you try these changes in your machine, i see almost 1.2% improvement on my machine but i also think it is due to thermals of my machine that i am getting inconsistent results

levkk:
Thanks! Will take a look shortly.

ygxio:
Ok

sgrif:
I'd like to address these all at once and add a lint instead of doing it piecemeal

ygxio:
Right

PR #1371, opened 2026-08-17
Remove the library target, only build binary target

sgrif:
We have a metric ton of dead code in the codebase stemming from the overuse of `pub`. In theory it'd be easy to find all of it by removing all uses of `pub` from the crate root, but because the consumer is actually `main.rs`, not a well definted public API, this results in the whole crate being seen as dead code.

By just moving the crate root to `main.rs`, it'll be much easier for us to find and remove any dead code. This does mean that we can no longer have integration tests or benches, as these require a library crate. Most of our integration tests are *actual* integration tests that involve running the server, so the single one we had could just be moved in-crate. Benchmarks are much more difficult. There's not really any way to have benchmarks which depend on modules of a bin crate. For now I've just removed the single benchmark as it's not run anywhere. Once we've removed the dead code and cleaned up the module structure, if we really want that benchmark back, we can add a separate `lib.rs` purely as a library target for that benchmark which exposes what it needs.


levkk:
I don't see any warnings printed by the compiler [here](https://github.com/pgdogdev/pgdog/actions/runs/32046100720/job/95434265809?pr=1371) about dead code. Is that expected?

sgrif:
Yes, I didn't wanted this to be easier to review without the noise of dead code removal in the same PR.

PR #1372, opened 2026-08-17
Remove a batch of dead code

sgrif:


pgdog/src/backend/replication/logical/subscriber/parallel_connection.rs
```diff
@@ -235,15 +218,11 @@ mod test {
         let mut parallel = ParallelConnection::new(server).unwrap();
 
         parallel
-            .send(
```
levkk, line 238: I don't remember if we do anything relevant in `send` but this may not be a no-op.

pgdog/src/backend/replication/logical/subscriber/stream.rs
```diff
@@ -125,7 +125,7 @@ pub struct StreamSubscriber {
     committed_lsn: i64,
     // Working position in the stream (advances on Begin for deduplication).
     lsn: i64,
-    lsn_changed: bool,
+    pub(super) lsn_changed: bool,
```
levkk: This gives mutable access to this attribute, so not semantically identical.

pgdog/src/backend/schema/sync/progress.rs
```diff
@@ -66,30 +52,17 @@ impl Default for Item {
...
             Self::Other { sql } => write!(f, "\"{}\"", no_comments(sql)),
-            Self::TableDump { schema, name } => write!(f, "table \"{}\".\"{}\"", schema, name),
         }
     }
 }
 
 impl Item {
     fn action(&self) -> &str {
         match self {
-            Self::Index { .. } => "creating",
```
levkk, line 89: That's surprising. We definitely create indexes. Maybe we log them differently.

levkk (approved):
Neat

pgdog/src/backend/replication/logical/subscriber/parallel_connection.rs
sgrif: I verified it is when I changed this

pgdog/src/backend/replication/logical/subscriber/stream.rs
sgrif: There's actually a lang change to give control over this in flight.

PR #1373, opened 2026-08-17
feat(lb): route specific users to replicas

### LINKED ISSUE #979: Route specific users to replica
Would it be easy to add a user level configuration to route the user only to the read replica?

We have a use case where it's not possible for us to configure correctly the options in the connection string, becuse of the third party system connection.

jkaczman:
When the `read_only` flag in users.toml is set true for a given `User`, this will now route that `User` strictly to replicas, assuming that there's replicas available. When we have no replicas, the primary will still be used. This change only affects using `include_primary` or `include_primary_if_replica_banned` as the read_write_split option.

Also: I found what I suspect to be a bug where we previously weren't checking for `&& !target.ban.banned()` in the LoadBalancer for `exclude_primary`.

Fixes #979.

### DIFF
--- a/pgdog/src/backend/pool/request.rs
+++ b/pgdog/src/backend/pool/request.rs
@@ -8,14 +8,19 @@ pub struct Request {
     pub id: FrontendPid,
     pub created_at: Instant,
     pub read: bool,
+
+    // Load balancer uses this to determine if primary should be allowed to read.
+    // Propagated from `User.read_only` setting.
+    pub read_only: bool,
 }
 
 impl Request {
-    pub fn new(id: FrontendPid, read: bool) -> Self {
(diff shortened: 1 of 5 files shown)


pgdog/src/backend/pool/request.rs
```diff
@@ -8,14 +8,19 @@ pub struct Request {
     pub id: FrontendPid,
     pub created_at: Instant,
     pub read: bool,
+
+    // Load balancer uses this to determine if primary should be allowed to read.
+    // Propagated from `User.read_only` setting.
+    pub read_only: bool,
```
sgrif:
```suggestion
    pub(crate) read_only: bool,
```

sgrif (approved):

levkk (approved):
🚀

PR #1360, opened 2026-08-13
perf: smaller future size

ygxio:
closes #1359

PR #1376, opened 2026-08-17
Box our largest futures

sgrif:
Async code needs to have a much lower stack size threshold to justify boxing vs sync code. The compiler's coroutine implementation can generate extremely [not ideal
code](https://github.com/rust-lang/rust/issues/69826), and in general async types are more likely be unable to have move memcpys elided.

This boxes any future that is greater than 10k in size across our codebase. This mostly addresses the root cause #1360 was targeting, though I've left that change in place as the doubling in size with the previous code was legit and this appears in enough hot paths to justify it.

### DIFF
--- a/pgdog/src/frontend/client/query_engine/query.rs
+++ b/pgdog/src/frontend/client/query_engine/query.rs
@@ -60,7 +60,7 @@ impl QueryEngine {
 
         match safe_timeout(
             context.timeouts.query_timeout(&State::Active),
-            self.client_server_exchange(context),
+            Box::pin(self.client_server_exchange(context)),
         )
         .await
         {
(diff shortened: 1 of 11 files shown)


pgdog/src/frontend/client/query_engine/query.rs
```diff
@@ -60,7 +60,7 @@ impl QueryEngine {
 
         match safe_timeout(
             context.timeouts.query_timeout(&State::Active),
-            self.client_server_exchange(context),
+            Box::pin(self.client_server_exchange(context)),
```
levkk, line 63: This is very hot. Very curious about the performance gains we get from this one, would be cool to see a bench

levkk (approved):

pgdog/src/frontend/client/query_engine/query.rs
sgrif: image

PR #1377, opened 2026-08-17
WIP: Don't allow traffic on database until healthcheck passes

### LINKED ISSUE #1309: [LB] Don't allow traffic on database until healthcheck passes
**PgDog version**
`v0.1.51`

**Description**
When databases are added to `pgdog.toml`, PgDog creates the pool immediately and adds it to the load balancer, assuming those databases are reachable. If they are not, the clients will return errors temporarily while the load balancer banning logic kicks in. We can avoid this by running a health check _first_ and only allow the database to serve traffic once it passes the initial check.

**Configuration**
Multiple databases, with `role = "replica"` or `role = "auto"`.

sgrif:
Fixes #1309

### DIFF
--- a/.schema/pgdog.schema.json
+++ b/.schema/pgdog.schema.json
@@ -100,6 +100,7 @@
         "read_write_strategy": "conservative",
         "regex_parser_limit": 1000,
         "reload_schema_on_ddl": true,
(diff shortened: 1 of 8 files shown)


levkk (approved):
Nice!

PR #1380, opened 2026-08-18
Fix connection stats handling and add RESET STATS command

### LINKED ISSUE #1281: RELOAD clears connection pool and topology statistics
PgDog supports hot-reloading its configuration at runtime by sending SIGHUP to the process, or by connecting to the admin database and issuing RELOAD. This is intended to let operators tweak settings (routing, load balancing, pool sizes, etc.) without dropping active client or server connections. PgDog also exposes real-time internal statistics via the admin console (SHOW STATS, SHOW CLIENTS, SHOW POOLS, etc.) . 
Currently, any reload, whether triggered by a no-op config re-apply or by an actual topology change (e.g., Patroni failover promoting a new primary, a replica being added/removed), resets all pool and topology statistics to zero. This is overly aggressive: even when the topology changes, most existing nodes (shards, replicas, primary) typically remain the same between reloads. Only the delta (added/removed/promoted nodes) should be treated as new or torn-down pools.

Impact / Use Cases:

1. HA failover monitoring: In Patroni-managed clusters, a primary/replica swap is common and expected; PgDog should treat the promoted node's pool as a continuation (matched by node identity, not by role) rather than wiping its stats.

2. Elastic scaling scenarios: When adding a new read replica or shard, existing pools shouldn't lose history just because the overall topology map changed, only the new node's pool should start fresh.

3. Incident correlation across failovers: Engineers debugging an outage that spans a failover event need pre- and post-failover stats on the same underlying node to see whether a problem followed the connection or the role.

4. Capacity planning accuracy: Long-running trend analysis (e.g., total queries per node, peak wait times) becomes far more reliable if minor topology churn doesn't erase months of accumulated counters for nodes that never actually left the cluster.

5. Reduced noise in dashboards: Grafana/Prometheus dashboards built on PgDog metrics would stop showing false "zero" dips at every reload, reserving real drops for actual node removal/restart events.

rodneyosodo:
### Problem
Every `RELOAD` (SIGHUP or admin command) zeroed all pool and cluster statistics. Operators currently misuse `RELOAD` to reset counters, @ykonrad asked for a proper `RESET STATS` command.

### What's changed
- **Pool stats survive reload**: `Pool::move_conns_to` now copies the source pool's `stats` into the destination pool (the one-line fix at the spot @levkk pointed to).
- **Cluster metrics survive reload**: `Cluster::move_conns_to` carries `mirror` counts and accumulates lookup-stats into the new cluster's Arc.
- **`RESET STATS` admin command**: zeros pool stats, cluster mirror counts, and lookup stats across all databases/mirrors. Symmetric with what reload preserves.
- Supporting helpers: `LookupStats::reset/accumulate`, `ClusterMetrics::reset`, `Shard/Databases::reset_stats`.

### Verification
Manual run through pgdog confirms: queries counted → RELOAD preserved → RESET zeroed → counting resumes.

Fixes https://github.com/pgdogdev/pgdog/issues/1281


sgrif:
Would you change anything marked as `pub` to either `pub(crate)` or `pub(super)` depending on where it's used, please?

pgdog/src/backend/pool/cluster.rs
```diff
@@ -472,13 +472,35 @@ impl Cluster {
...
     pub(crate) fn move_conns_to(&self, other: &Cluster) -> Result<(), Error> {
+        // Carry cluster-level statistics over so a reload doesn't reset
+        // mirror and lookup counters. The lookup counters are accumulated
+        // into the new cluster's own stats Arc (unless it's the same one),
+        // so lookups served by the new cluster keep counting.
+        {
+            let from = self.stats.lock();
+            let mut to = other.stats.lock();
+            to.mirror = from.mirror;
+            if !Arc::ptr_eq(&to.lookup, &from.lookup) {
```
levkk, line 467: I would put this assertion higher, before we take the lock, to avoid deadlocks. Basically, we want to make sure that `self` and `other` are never the same cluster object.

PR #1381, opened 2026-08-18
fix: do not generate node_id if NODE_ID env is not set

meskill:
The test is not flaky - the math if flaky

### DIFF
--- a/pgdog/src/util.rs
+++ b/pgdog/src/util.rs
@@ -8,7 +8,7 @@ use once_cell::sync::Lazy;
 use rand::{Rng, distr::Alphanumeric};
 use std::ops::ControlFlow;
 use std::panic::Location;
-use std::{env, future::Future, future::pending, num::ParseIntError, time::Duration};
+use std::{env, future::Future, future::pending, time::Duration};
 use tokio::time::Interval;
 use tracing::warn;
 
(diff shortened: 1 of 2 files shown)


pgdog/src/util.rs
meskill, line 133: 00000023

levkk:
Sorry I don't follow, could you explain the bug and the fix?

pgdog/src/util.rs
```diff
@@ -533,10 +538,21 @@ mod test {
...
+
+    #[test]
+    fn test_node_id_not_a_number() {
+        let _guard = set_env_var("NODE_ID", "pgdog-abc");
+        assert_eq!(node_id(), None);
+    }
+
+    #[test]
+    fn test_node_id_bare_number() {
+        let _guard = set_env_var("NODE_ID", "7");
```
levkk, line 554: The acutal node id we get in kube is gonna be `name-deployment-<number>`. We do test it below, so that's good, but this test doesn't cover the actual documented use case for this: https://docs.pgdog.dev/features/sharding/unique-ids/#configuration

meskill: the test documents current behavior and it worked like this before, so if anyone setting it explicitly...

meskill:
> Sorry I don't follow, could you explain the bug and the fix?

ok, you see the code I commented? The old implementation without NODE_ID was using instance_id that was randomly generated as hex string. And the random hex string is a proper base10 number sometimes - that's why node_id was returning something occasionally in flaky test

levkk:
> > Sorry I don't follow, could you explain the bug and the fix?
> 
> ok, you see the code I commented? The old implementation without NODE_ID was using instance_id that was randomly generated as hex string. And the random hex string is a proper base10 number sometimes - that's why node_id was returning something occasionally in flaky test

image

levkk (approved):

PR #1384, opened 2026-08-18
fix(schema_sharding): check for previous SearchPath and Set/Comment in cross_shard_omni_check

### LINKED ISSUE #1374: [Query engine] schema sharding broken due to cross-shard omni check
**PgDog version**
`main`

**Description**
When using schema sharding, our cross-shard omni table write checker is triggering false positives for catalog tables. This breaks schema sharding since people can't execute some standard queries.

**Logs**
```
psycopg.errors.SystemError: cannot write to an omnisharded table in a direct-to-shard transaction
DETAIL:  SELECT
    typname AS name, oid, typarray AS array_oid,
    oid::regtype::text AS regtype, typdelim AS delimiter
FROM pg_type t
WHERE t.oid = to_regtype($1)
ORDER BY t.oid
```

**Configuration**
```toml
[general]
cross_shard_disabled = true

[[databases]]
host = "db0"
name = "pgdog"
shard = 0

[[databases]]
host = "db1"
name = "pgdog"
shard = 1

[[sharded_schemas]]
database = "pgdog"
name = "customer_a"
shard = 0

[[sharded_schemas]]
database = "pgdog"
name = "customer_b"
shard = 1
```

jkaczman:
Fixed a false-positive from `cross_shard_omni_check` when handling manual routing in schema sharding.

Used to trigger when all these were true:
-  In pgdog.toml, `read_write_strategy `= `conservative` (default). This is why I put the test in integration/complex; the default integration pgdog.toml config uses `aggressive`.
- SET `pgdog.shard` (or comment directive)
- Running within a transaction
- Querying an omnisharded table (SELECT, read)

Resulting in the error: `cannot write to an omnisharded table in a direct-to-shard transaction`

Being that it's a transaction, it's treated as a write due to read_write_strategy setting. SET `pgdog.shard` is higher priority than `SearchPath`, so it overtakes it in `ShardsWithPriority`. Thus, when `peek()` is called in `ShardsWithPriority`, it will show the `Set`, and therefore, when `!is_search_path()` -> `requires_full_shard_coverage()` -> `is_omnishard_unsafe()`, it'll return as true (marking it unsafe). This did not consider the SET (manual routing) in the schema sharding setup, or that it masked the `SearchPath`.

Regression test was implemented, and it now passes.

Fixes https://github.com/pgdogdev/pgdog/issues/1374.

### DIFF
--- a/pgdog/src/frontend/router/parser/route.rs
+++ b/pgdog/src/frontend/router/parser/route.rs
@@ -1,12 +1,11 @@
 use std::{fmt::Display, ops::Deref};
 
-use lazy_static::lazy_static;
-
 use super::{
     Aggregate, DistinctBy, Limit, OrderBy, explain_trace::ExplainTrace,
     rewrite::statement::aggregate::AggregateRewritePlan, statement::AdvisoryLocks,
 };
 use crate::frontend::router::sharding::PendingLookup;
+use lazy_static::lazy_static;
 
 /// The shard destination for a query.
 #[derive(Debug, Clone, PartialEq, PartialOrd, Ord, Eq, Hash, Default)]
@@ -129,6 +128,10 @@ pub struct Route {
     /// The query engine resolves them and routes the query again;
     /// the query doesn't execute while any are unresolved.
     pending_lookups: Vec<PendingLookup>,
+    /// Being routed by `ShardSource::Comment` or `ShardSource::Set`
+    manual_routing: bool,
+    /// Previously (or currently) a `ShardSource::SearchPath` as the source
+    contains_search_path: bool,
 }
 
 impl Display for Route {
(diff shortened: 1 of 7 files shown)


levkk (approved):
Cool!

pgdog/src/frontend/router/parser/route.rs
```diff
@@ -283,12 +286,40 @@ impl Route {
...
+
+    pub fn set_manual_routing(&mut self, manual_routing: bool) {
+        self.manual_routing = manual_routing;
+    }
+
+    /// Returns true if the shard is being routed by `ShardSource::Comment` or `ShardSource::Set`
+    pub fn is_manual_routing(&self) -> bool {
+        self.manual_routing
+    }
+
```
sgrif, line 307: If you're providing both a getter and setter, in general it's preferable to make the field `pub(crate)`

jkaczman: Will do! I'm going to make another PR changing that code soon, so I'll fix that there. 

Should I change all the other getter/setters that were already there to use pub(crate) as well?

sgrif: If it were part of this PR I wouldn't but as a separate pr yeah I'd do the whole struct at once. It's up to you either way.

PR #1390, opened 2026-08-18
refactor: cleanup multi-shard state & decoder

levkk:
All changes are no-ops.

- remove redundant code in multi shard state manager, and move message handling into its own functions
- give methods more intuitive names
- fix a few comments
- remove no longer necessary `debug!`
- panic if row decoder has no row description


sgrif (commented):
Apparently you can't approve a PR after it's been merged anymore but 👍

PR #1392, opened 2026-08-19
refactor: remove unused code

meskill:
If we drop pub from main.rs file we'll see a lot of unused code and imports. This pr does just it and removes all the unreferenced code (with some exceptions). 

The intention is to have less code and not to mess with pub(crate) everywhere just to have the unused warnings


sgrif:
Sorry, I think we ended up working on this in parallel :(

PR #1393, opened 2026-08-19
add application_name_add_host setting

### LINKED ISSUE #253: What is equivalent of "application_name_add_host" to pass the application_name connecting to postgresql?
What is the equivalent to pgbouncer's `application_name_add_host` ? [https://www.pgbouncer.org/config.html]
i have checked the docs [https://docs.pgdog.dev/configuration/pgdog.toml/general/] and can't find any setting for this, **we need this**

murex971:
## Summary

- Add `[general] application_name_add_host` (default `false`, env `PGDOG_APPLICATION_NAME_ADD_HOST`) to append the client address to `application_name` as `{name} - {ip}:{port}`.
- Apply the suffix at login and again when the client changes `application_name` via `SET` / `set_config`, so `pg_stat_activity` can identify the source of queries.
- If the client sends no `application_name`, use an empty prefix (` - 10.0.0.5:1234`) rather than PgBouncer’s `"app"` default.
- Snapshot client settings once per request (`ClientRequestSettings` in `buffer()`) .

closes #253


pgdog-config/src/general.rs
```diff
@@ -606,6 +606,19 @@ pub struct General {
     #[serde(default = "General::log_disconnections")]
     pub log_disconnections: bool,
 
+    /// Add the client host address and port to `application_name` at connection start
+    /// and whenever the client later changes it with `SET` / `set_config`.
```
jkaczman, line 623: If we're trying to mirror the PgBouncer feature, it seems we **shouldn't** be appending the client host address and port if the client later changes it.

> If application_name is later changed with SET, PgBouncer does not change it again.

murex971: yes, it doesn't change, but I wasn't sure if that was a conscious decision or implementation limitation for pg_bouncer.
I thought since we could, maybe we should ?

Do you want me to revert this ?

levkk: I think you're right, this is because pgbouncer can't handle `SET`. We should support this.

pgdog/src/frontend/client/query_engine/query.rs
```diff
@@ -324,9 +325,16 @@ impl QueryEngine {
             // Update client params with values
             // sent from the server using ParameterStatus(B) messages.
             if !changed_params.is_empty() {
+                let add_host = config().config.general.application_name_add_host;
```
levkk: Could you fetch this info here: https://github.com/pgdogdev/pgdog/blob/main/pgdog/src/frontend/client/mod.rs#L638

I think we should also refactor this into something like `ClientRequestSettings`. The idea is to fetch this info _once_ per request to avoid any race conditions around configuration reloading.

murex971: addressed

pgdog/src/frontend/client/query_engine/query.rs
```diff
@@ -320,9 +322,16 @@ impl QueryEngine {
             // Update client params with values
             // sent from the server using ParameterStatus(B) messages.
             if !changed_params.is_empty() {
+                let add_host = context.request_settings.application_name_add_host;
+                let host = context.client_addr.to_string();
```
levkk: This is an unconditional allocation. You may want to push it down into the:

```rust
if add_host && name.eq_ignore_ascii_case("application_name")
```

clause to avoid it on the hot path for default deployments that doesn't enable this setting.

murex971:
@levkk gtg

jkaczman:
@murex971 if you don't mind fixing the conflicts, I can review this again

murex971:
@jkaczman done!

PR #1394, opened 2026-08-19
Remove another batch of dead code

sgrif:
99% of any changed lines are changes to path imports or indentation changes for modules that were `test::test`. The only case of a meaningful code change is in query_engine/two_pc/server_transactions.rs. We were loading a bunch of information and putting them in an enum that was never read. The consuming code only cares about checking that this is non-empty. This could be refactored further to just `SELECT EXISTS` but I chose the smallest possible change that eliminated the dead code instead of a larger refactor.

This commit reveals several public configuration options that were unused. I have not removed them from `pgdog-config` in this commit, as that change feels like it should be discussed in isolation, while this commit is generally purely mechanical in nature.


levkk:
I was gonna say, jealous of those red stats, but...

image

check this out: https://github.com/pgdogdev/pgdog/pull/1378

PR #1396, opened 2026-08-19
perf: Box::pin removed for future returned by `self.client_server_exchange(context)`

### LINKED ISSUE #1395: perf: Box::pin removed for  future returned by `self.client_server_exchange(context)`
Clients lock the whole `RwLock<HashMap>` when updating their entries

so if 10 clients are connected to same shard they unnecessarily wait for lock, 

instead the lock should be on per entry rather than whole Hashmap

proof

Image

Huge chunk is spent waiting for obtaining mutex of the DashMap

This is a potential perf optimisation

ygxio:
closes #1395

### DIFF
--- a/pgdog/src/frontend/client/query_engine/query.rs
+++ b/pgdog/src/frontend/client/query_engine/query.rs
@@ -1,3 +1,5 @@
+use std::pin::pin;
+
 use tracing::{info, trace};
 
 use crate::{
@@ -58,12 +60,21 @@ impl QueryEngine {
             }
         }
 
-        match safe_timeout(
-            context.timeouts.query_timeout(&State::Active),
-            Box::pin(self.client_server_exchange(context)),
-        )
-        .await
-        {
+        // Stack-pinned, not `Box::pin`: this runs once per query, and boxing
+        // costs a malloc, a free and a memcpy of the whole exchange future
+        // every time. `Pin<&mut F>` is a `Future`, so `safe_timeout` takes it.
+        //
+        // The inner block matters: unlike the boxed future, which was moved
+        // into `safe_timeout` and dropped by the `.await`, `pin!` binds a local
+        // that would otherwise hold its borrow of `self` and `context` until
+        // the end of the function, past the arms below.
+        let query_timeout = context.timeouts.query_timeout(&State::Active);
+        let result = {
+            let exchange = pin!(self.client_server_exchange(context));
+            safe_timeout(query_timeout, exchange).await
+        };
+
+        match result {
             Ok(response) => response?,
             Err(err) => {
                 // Close the conn, it could be stuck executing a query
--- a/pgdog/src/frontend/client/query_engine/query.rs
+++ b/pgdog/src/frontend/client/query_engine/query.rs
@@ -1,5 +1,3 @@
-use std::pin::pin;
-
 use tracing::{info, trace};
 
 use crate::{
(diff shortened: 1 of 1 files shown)


pgdog/src/frontend/client/query_engine/query.rs
```diff
@@ -58,12 +60,21 @@ impl QueryEngine {
...
+        // costs a malloc, a free and a memcpy of the whole exchange future
+        // every time. `Pin<&mut F>` is a `Future`, so `safe_timeout` takes it.
+        //
+        // The inner block matters: unlike the boxed future, which was moved
+        // into `safe_timeout` and dropped by the `.await`, `pin!` binds a local
+        // that would otherwise hold its borrow of `self` and `context` until
+        // the end of the function, past the arms below.
+        let query_timeout = context.timeouts.query_timeout(&State::Active);
+        let result = {
+            let exchange = pin!(self.client_server_exchange(context));
```
jkaczman: It seems like Box::pin was used here to put the Future on the heap (introduced 3d ago in https://github.com/pgdogdev/pgdog/pull/1376) because of the large size of the Future (over 10k in size), specifically due to what's mentioned in the clippy error CI is failing on: https://rust-lang.github.io/rust-clippy/rust-1.96.0/index.html#large_futures 

> "Due to the current [unideal implementation](https://github.com/rust-lang/rust/issues/69826) of Coroutine, large size of a Future may cause stack overflows."

Whereas, this puts the Future on the stack. This sounds like it gives us a performance boost, but is it safe to use in production?

ygxio:
Can i get a review here @levkk

jkaczman (commented):
I looked through the issue linked, and it doesn't look like the stats you showed correlate to these exact changes (had some other things included); could you please clarify on the exact improvements here?

ygxio:
yes as per keryl i removed the changes of DashMap and only kept the pinning changes

jkaczman:
> yes as per keryl i removed the changes of DashMap and only kept the pinning changes

I saw you had this chart for DashMap + Box::pin,
image

Could you make one for this one solo?

ygxio:
sure it was same for only pin changes, would you mind testing tps on ur system too?

sgrif:
I'm not sure this is the right move. The whole point of this change was to avoid having it on the stack because the future size becomes too large otherwise. Removing this allocation is just going to end up moving the cost up the stack, not get rid of it

ygxio:
@sgrif yes that is right but what i saw is extra rjmallocs, i donot have screenshot right now but i will get one, will post in this thread.

pgdog/src/frontend/client/query_engine/query.rs
```diff
@@ -58,12 +60,21 @@ impl QueryEngine {
...
-        .await
-        {
+        // Stack-pinned, not `Box::pin`: this runs once per query, and boxing
+        // costs a malloc, a free and a memcpy of the whole exchange future
+        // every time. `Pin<&mut F>` is a `Future`, so `safe_timeout` takes it.
+        //
+        // The inner block matters: unlike the boxed future, which was moved
+        // into `safe_timeout` and dropped by the `.await`, `pin!` binds a local
+        // that would otherwise hold its borrow of `self` and `context` until
+        // the end of the function, past the arms below.
```
sgrif: I'd prefer to see comments like this as part of the commit message rather than on the code itself. These often get out of sync with the code they're talking about and result in the code base being littered with inaccurate statements about why it's written a certain way. I'd only mention the borrow of `self` and `context` if the code fails to compile without the block

sgrif:
> yes that is right but what i saw is extra rjmallocs

Right, but that's not inherently a problem. Replacing an allocation with a 10k stack variable is a tradeoff not an automatic win. The fact that you're seeing memcpy upon boxing implies something else is interfering with the optimizer. The compiler should be able to place the value in the box directly without having to copy it from the stack.

pgdog/src/frontend/client/query_engine/query.rs
ygxio: sure i will take care of it

ygxio:
> > yes that is right but what i saw is extra rjmallocs
> 
> Right, but that's not inherently a problem. Replacing an allocation with a 10k stack variable is a tradeoff not an automatic win. The fact that you're seeing memcpy upon boxing implies something else is interfering with the optimizer. The compiler should be able to place the value in the box directly without having to copy it from the stack.

will dig down more, need to know more on this issue

sgrif:
To be clear, I don't think this change is 100% the wrong choice. I just don't think there's been sufficient investigation into what's going on. While this is a relatively hot path, one extra allocation per request should not be causing huge impacts on our performance.

sgrif:
> will dig down more, need to know more on this issue

Thank you!

ygxio:
starting from the root cause

image

execute occupies almost 5.2% of the samples

out of which have some mallocs and rjems

image

image

image

image

image

image

sgrif:
Alright, I'm sufficiently convinced. If you can update this to pass CI I'll merge it. None of the `pin` stuff is necessary, though. Removing `Box::pin` is the only change required

sgrif:
This is perfect. Thanks for working on this!

PR #1398, opened 2026-08-19
fix(sharded_schemas): during exclusive use, relax cross_shard_omni_check; add mixed use notes

jkaczman:
Follow up on https://github.com/pgdogdev/pgdog/pull/1384 for issue https://github.com/pgdogdev/pgdog/issues/1374,

When `pgdog.toml` is configured to use `sharded_schemas` exclusively (as in, `sharded_tables` are not configured), relax `cross_shard_omni_check`. 

When both `sharded_schemas` and `sharded_tables` are being used, the error will still throw. Added a TODO to defer consideration of multiple sharding functions. I also added a test for that for later.

Additionally fixed the check for `OmniWriteWithDirective` from allowing writes to `omnisharded_tables` in `sharded_tables` when `sharded_schemas` were configured in the database (the check previously didn't make sure that it was solely `sharded_schemas`).

Removed tests that used both sharded_tables (lookup table) and sharded_schemas.


levkk (approved):
looks good i think!

jkaczman:
@levkk it seems the `omni_in_direct_to_shard` error never gets triggered anymore after fixing the `omni_write_with_directive` conditioning to check for **exclusive** `sharded_schemas`. Meaning, the directive one catches the ones that were previously erroring on `omni_in_direct_to_shard`. Does it still look good? I removed the (what seems to be) redundant error.

levkk:
> @levkk it seems the `omni_in_direct_to_shard` error never gets triggered anymore after fixing the `omni_write_with_directive` conditioning to check for **exclusive** `sharded_schemas`. Meaning, the directive one catches the ones that were previously erroring on `omni_in_direct_to_shard`. Does it still look good? I removed the (what seems to be) redundant error.

I suspect this is because all of our unit test harnesses configure schema sharding, which this PR explicitly checks for and bypasses that check. I think if we want to make sure the check still works, configure a test with only column-based sharding and try the bad request with that.

integration/complex/cross_shard_omni_check/run.py
```diff
@@ -55,12 +58,123 @@ async def test_cross_shard_omni() -> None:
             assert type(row) is Record
             assert row['name'] == "int4"
 
+        # Test without using pgdog.shard; just use search_path.
+        # Nothing is abnormal about this (vs the last one); it should work fine.
+        async with conn.transaction():
```
sgrif: Out of curiosity, is there a reason the integration tests for this are in Python? It's not necessarily a problem but I think it'd be good to default to Rust integration tests unless we're specifically testing a regression/behavior of a specific client library.

pgdog/src/frontend/router/parser/route.rs
```diff
@@ -286,40 +284,21 @@ impl Route {
         self.search_path_driven
     }
 
-    pub fn set_contains_search_path(&mut self, contains_search_path: bool) {
-        self.contains_search_path = contains_search_path;
+    pub fn set_sharded_schema_only(&mut self, sharded_schema_only: bool) {
```
sgrif: Since this PR is touching this field anyway, I'd love to see the getter/setter yeeted

jkaczman:
> > @levkk it seems the `omni_in_direct_to_shard` error never gets triggered anymore after fixing the `omni_write_with_directive` conditioning to check for **exclusive** `sharded_schemas`. Meaning, the directive one catches the ones that were previously erroring on `omni_in_direct_to_shard`. Does it still look good? I removed the (what seems to be) redundant error.
> 
> I suspect this is because all of our unit test harnesses configure schema sharding, which this PR explicitly checks for and bypasses that check. I think if we want to make sure the check still works, configure a test with only column-based sharding and try the bad request with that.

It looks like we have these still, which use `new_sharded_client_without_schemas()`
https://github.com/pgdogdev/pgdog/blob/e93eab62dca178b7b01f5fb7f7682f41e4d1cd31/pgdog/src/frontend/client/query_engine/test/test_omnisharded.rs#L208-L235

and

https://github.com/pgdogdev/pgdog/blob/e93eab62dca178b7b01f5fb7f7682f41e4d1cd31/pgdog/src/frontend/client/query_engine/test/test_omnisharded.rs#L237-L262

These check for `omni_write_with_directive`. I wonder if `omni_in_direct_to_shard` was just created because of the missed logic error? Or perhaps for using two sharding functions simultaneously?

integration/complex/cross_shard_omni_check/run.py
jkaczman: Moved to Rust! I used Python because I forgot we could override pgdog.toml `read_write_strategy` with an admin SET in the tests (complex came with custom pgdog.toml files).

pgdog/src/frontend/router/parser/route.rs
jkaczman: Done!

levkk (approved):
Looks good I think! Let's try it!

PR #1400, opened 2026-08-19
feat: split multi-query requests

levkk:
Using our parser, tell the query engine to re-execute the requests as a simple query pipeline.

Refactor the query engine and remove the junk from `client.rs`. Also move request splicing to the query engine. Also, remove multi-set handling, which is no longer necessary.


pgdog/src/frontend/client/query_engine/start_transaction.rs
```diff
@@ -42,6 +42,16 @@ impl QueryEngine {
         Ok(())
     }
 
+    pub(super) async fn start_transaction_multi_query(&mut self) -> Result<(), Error> {
+        if !self.backend.connected() {
+            return Err(Error::ImplicitOnDisconnected);
+        }
+
+        self.backend.execute("BEGIN").await?;
```
sgrif, line 50: If we're rewriting to the extended protocol, we shouldn't need this explicit transaction control. IMO we should avoid needing to handle it this way so we don't have to worry about handling transaction control statements in the query itself.

pgdog/src/frontend/router/parser/cache/ast.rs
```diff
@@ -96,9 +96,15 @@ impl Ast {
         let mut rewrite_plan = Default::default();
         let ast = make::try_owned(|mem| {
             let mut ast = mem.parse(query.query_without_comment)?;
-            if let Some(stmt) = ast.as_mut().into_iter().next() {
+            // Only rewrite if there is one statement in the request.
+            // Otherwise, the parser will refuse to run it and tell the query
+            // engine to re-send statements separately.
+            let only_one = ast.stmts().count() == 1;
+
+            if only_one && let Some(stmt) = ast.as_mut().into_iter().next() {
```
sgrif:
```suggestion
            if let Ok(stmt) = ast.as_mut().into_iter().exactly_one() {
```

pgdog/src/frontend/router/parser/query/split_simple.rs
```diff
@@ -0,0 +1,54 @@
...
+
+        let stmts = ast
+            .ast
+            .stmts()
+            .map(deparse)
+            .collect::<Result<Vec<_>, PgParseRawError>>()?;
+        let queries = stmts
+            .into_iter()
+            .map(|query| query.as_str().to_string())
+            .collect::<Vec<_>>();
```
sgrif, line 50: We don't need the intermediate vec

```suggestion
        let queries = ast
            .ast
            .stmts()
            .map(|stmt| {
                let query = deparse(stmt)?;
                Ok::<_, Error>(query.as_str().to_owned())
            })
            .collect::<Result<Vec<_>, _>>>()?;
```

sgrif (approved):
Would be good to have an integration test that looks like

```
execute("INSERT ...; BEGIN; INSERT ...")
execute("ROLLBACK")

assert_only_first_insert_visible
```

pgdog/src/frontend/router/parser/cache/ast.rs
levkk: Oh that's nice!

pgdog/src/frontend/client/query_engine/start_transaction.rs
levkk: Yeah I agree. It was just more complicated because then we would need to handle `PREPARE` and `EXECUTE` sent via `Parse`, which we currently don't, so the scope increased.

It's actually pretty easy for us right now to start/commit implicit transactions ourselves, we do it already for normal transactions and for 2pc, so this is not a big leap.

But that of course still leaves the problem of a client lib executing `PREPARE` via `Parse` unsolved.

pgdog/src/frontend/router/parser/query/split_simple.rs
levkk: I think I was building something else here and I forgot to remove it

pgdog/src/frontend/client/query_engine/start_transaction.rs
sgrif: My only thought is that fixing PREPARE/EXECUTE via extended protocol is a relatively contained (and I would hope straightforward) bug fix, while not being able to support `BEGIN` mid-query is more structural and harder to address later.

But I don't feel strongly about this either way

levkk: I'm starting to come around to your point of view here. Ive been debugging transaction state for a couple days now and it has a lot of interactions with savepoints, 2pc and maybe supporting prepare here isn't worth it ...

levkk: initially*

sgrif: Happy to do the work of supporting it separately from this PR :)

levkk:
Done in #1432

PR #1405, opened 2026-08-20
feat(tls): server certificate SAN can have a trailing dot

levkk:


pgdog/src/net/tls.rs
```diff
@@ -485,6 +494,109 @@ pub fn connector_with_verify_mode(
     Ok(connector)
 }
 
+/// Certificate verifier that treats one trailing dot in a DNS SAN as optional.
+#[derive(Debug)]
+struct TrailingDotVerifier {
+    webpki_verifier: Arc<dyn ServerCertVerifier>,
+}
+
+impl TrailingDotVerifier {
```
sgrif, line 503: I'm not familiar with the rustls API but this seems like a lot of code for a trailing dot. (And it wouldn't surprise me at all that the TLS library API actually makes it this hard)

sgrif (approved):

pgdog/src/net/tls.rs
levkk: Yeah... I'm not sure if this is how this is supposed to be done. I'm waiting for a review from Ryan.

rlittlefield:
It looks like curl strips trailing dots automatically, but the default in go `crypto/x509`, python's ssl, and `rustls-webpki` all mismatch this, apparently by design. From what I can tell, its _probably_ a malformed cert from google.

Here is the python GCP cloud sql issue:
https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/issues/1087

It looks like the GCP Go connector does the same thing this PR does, so that's a good sign: https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/main/internal/cloudsql/tls_verify.go#L112-L118

I guess I could see not wanting to fall back to custom code for strict verification, so if that's a concern I suppose it could be a configuration option to strip the dot?

pgdog/src/net/tls.rs
```diff
@@ -485,6 +494,109 @@ pub fn connector_with_verify_mode(
...
+            intermediates,
+            server_name,
+            ocsp_response,
+            now,
+        );
+
+        if matches!(
+            &result,
+            Err(rustls::Error::InvalidCertificate(
+                rustls::CertificateError::NotValidForNameContext { .. }
```
rlittlefield, line 535: Would this make sense as
```
Err(rustls::Error::InvalidCertificate(
    rustls::CertificateError::NotValidForName
    | rustls::CertificateError::NotValidForNameContext { .. },
))
```
It sounds like rustls considers it okay to change enum variants, so this might help avoid a silent behavior change if I'm reading this right.

PR #1415, opened 2026-08-22
perf: replace ToString with AsRef

meskill:
Avoid allocations when possible for different net messages by avoiding ToString trait and using AsRef<str>.

before -> after
image

before
image

after 
image

I have to come up with very specific shape of benchmark sql to see the good performance boost, otherwise it was not hurting performance anyway in most cases, but better to overprovision.


levkk (approved):
Awesome

PR #1417, opened 2026-08-24
refactor(sharding_key_update): remove SELECT from old shard; add RETURNING * to DELETE

jkaczman:
Removes the unnecessary `SELECT` query from the old shard at the beginning of sharding key updates. Moves the row fetch to a `RETURNING *` clause in the `DELETE` achieving equivalent functionality with one less query.

Co-authored-by: Sage <sagetheprogrammer.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2969,7 +2969,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 [[package]]
 name = "pg_raw_parse"
 version = "0.1.0"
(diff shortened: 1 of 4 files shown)


sgrif (approved):
Want to wait to see what @levkk says before we merge this in case there's something we forgot about that we don't have test coverage for

levkk (approved):
LGTM, as long as you're cool with doing unconditional row moves. The edge case covered by fetching the row first is somewhat unlikely so that's probably fine.

sgrif:
Yeah, we figured both that this is sufficiently rare to not worry about, and that the cost shouldn't be much higher if we already had to query all shards to determine the shard the row previously lived on.

PR #1418, opened 2026-08-24
refactor(query_engine_context): remove rw result from context field

jkaczman:
Starting to restructure RewriteResult to become a "Query Planner";

As part of our goal to support more "multi-query" queries such as subselects and CTEs, as well as supporting better EXPLAIN output, we need an intermediate step between parsing and execution that creates a proper "Query Plan".

After going through the code and poking it a few different ways, we determined that the best place to live is what is currently called RewriteResult, and what we expect to eventually call `QueryPlanner`. Our long term goal is to have this take the query engine context as an argument, and return the abstract steps required, those steps being the query to be run (with subqueries replaced with placeholders), and the route it is going to be run on.

As we're promoting this to a "thing that does the thing", it no longer makes conceptual sense to have it live on QueryEngineContext, as it's not context. Additionally, we intend to eventually have this start to hold onto borrowed data, and adding an additional lifetime to QueryEngineContext will affect much more code than what is necessary for this refactor.

Co-authored-by: Sage <sagetheprogrammer.com>


levkk:
Might also want to look at combining it with `Route`. That thing has a lot of context that you will care about, .e.g., aggregate handling, sorting, limit/offset + routing information. All of this will be needed for the `EXPLAIN` 2.0.

PR #1420, opened 2026-08-24
refactor: deprecation warning for explicit set to 'pg_query_protobuf'; remove QueryParserEngine references in pgdog

jkaczman:
Removes internal `QueryParserEngine` fields/references that were no longer being used. Emits a deprecation warning if the user has `query_parser_engine = "pg_query_protobuf"` explicitly set in their pgdog.toml file.


levkk (approved):
Sweet!

PR #1421, opened 2026-08-24
fix(lb): use CancellationToken reliable shutdown signal

meskill:
Replace pack of Notify with CancellationToken to implement the shutdown signal.

One of the issue I faced is the `lb/monitor` that was not shutting down sometimes gracefully. Also, I think this is the reason the complex integration test fails sometime.


levkk (approved):
Sweet

sgrif (approved):
Heck yeah! -1 reimplementing tokio primitives ourselves

PR #1422, opened 2026-08-24
feat(pg_raw_parse): use our own version of normalize

jkaczman:
Updates pg_raw_parse rev to https://github.com/pgdogdev/pg_raw_parse/pull/41 to use our own version of `normalize`.

### DIFF
--- a/pgdog/src/frontend/router/parser/cache/cache_impl.rs
+++ b/pgdog/src/frontend/router/parser/cache/cache_impl.rs
@@ -5,6 +5,7 @@ use std::collections::HashMap;
 use std::time::Duration;
 
 use parking_lot::Mutex;
+use pg_raw_parse::nodes;
 use std::sync::Arc;
 use tracing::debug;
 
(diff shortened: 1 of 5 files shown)


pgdog/src/frontend/router/parser/cache/cache_impl.rs
```diff
@@ -212,19 +213,20 @@ impl Cache {
     /// Used by dry run mode to keep stats on what queries are routed correctly,
     /// and which are not.
     ///
-    pub fn record_normalized(&self, query: &str, route: &Route) -> Result<(), Error> {
-        let normalized = normalize(query)?;
+    pub fn record_normalized(&self, query: &nodes::RawStmt, route: &Route) -> Result<(), Error> {
+        let normalized = normalize(query);
+        let normalized = normalized.stmt().as_str().ok_or(Error::EmptyQuery)?;
```
sgrif: You want to use `deparse`. `Node::as_str` checks if it's a `String` node and extracts the inner value. It's not for turning queries into SQL

sgrif: (which is why you're seeing `EmptyQuery` errors in CI. `RawStmt::stmt` will never be a `Node::String`, so `Node::as_str` will always return `None` here.)

sgrif (approved):

PR #1427, opened 2026-08-25
fix(lb): avoid NoPrimary errors while waiting for primary election

meskill:
Fix the issue with the subscription to primary update on load-balancer when the primary temporarily unavailable. 

Please see the test and comment on it for context. 

The issue could generate unnecessary NoPrimary errors instead of actually waiting the new primary to be identified due to the misuse of Notify that always generated +1 permit effectively removing any subscription after first wait.

### DIFF
--- a/pgdog/src/backend/pool/lb/mod.rs
+++ b/pgdog/src/backend/pool/lb/mod.rs
@@ -9,7 +9,7 @@ use std::{
 };
 
 use rand::seq::SliceRandom;
-use tokio::sync::Notify;
+use tokio::sync::watch;
 use tokio_util::sync::CancellationToken;
 use tracing::warn;
 
@@ -95,8 +95,8 @@ pub struct LoadBalancer {
     pub(super) lb_strategy: LoadBalancingStrategy,
     /// Shutdown signal for the replicas monitor.
     pub(super) maintenance: CancellationToken,
-    /// Role detection waiter.
-    pub(super) role_detection: Arc<Notify>,
+    /// Pool elected as primary by automatic role detection.
+    pub(super) elected_primary: Arc<watch::Sender<Option<Pool>>>,
     /// Read/write split.
     pub(super) rw_split: ReadWriteSplit,
 }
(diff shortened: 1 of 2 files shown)


pgdog/src/backend/pool/lb/test.rs
```diff
@@ -1513,16 +1518,86 @@ async fn test_auto_mode_primary_election_releases_writes() {
...
+
+    set_lsn_stats(&lb.targets[1], false, 200);
+    assert!(lb.redetect_roles());
+    assert!(!lb.redetect_roles());
     assert!(lb.primary().is_some());
+
+    // primary should be resolved to second target
+    let primary = spawned.await.unwrap().unwrap();
+    assert_eq!(primary.pool.addr().host, "localhost");
+    drop(primary);
```
meskill, line 1598: for the original code here spawned were returned immediately with the NoPrimary error though it should've waited for primary as expected. 

`.notified().await` always triggered and the error raised when trying to actually get the primary

pgdog/src/backend/pool/lb/mod.rs
```diff
@@ -324,37 +324,30 @@ impl LoadBalancer {
...
-        Ok(())
-    }
+        if !self.role_detection_enabled() {
+            return Err(Error::NoPrimary);
+        }
 
-    pub(super) async fn get_primary(&self, request: &Request) -> Result<Guard, Error> {
-        self.get_primary_internal(request).await
-    }
+        let mut receiver = self.elected_primary.subscribe();
```
levkk: What if `redetect_roles` sends a new primary just before you subscribe to this channel? I think if you wanted to avoid this, you would subscribe before `self.primary()` call above, but that would involve polling the `RwLock` inside the `watch` channel for each write query.

meskill: Good point!
This thing is covered by `watch_for` method that is called for current and future changes in the channel, so the value will be still visible to it. I added the test for this to verify that subscribing after the `redetect_roles` still resolves the new primary

levkk (approved):
Nice, I think this is good.

PR #1432, opened 2026-08-25
feat: partial multi-query support

levkk:
## What

Support multi-statement queries sent via the simple protocol if they can be safely executed without us having to do anything fancy with transactions.

## Preface

A multi-statement query looks something like this:

```sql
SELECT 1; SELECT 2;
```

They are sent inside a single `Query` message and Postgres is expected to process all commands sequentially and inside a single, implicit transaction.

This presents an issue for sharded deployments, since queries might need to go to different shards. So, what can we do about this?

## Split and execute

Since we have the Postgres parser handy, we can split the query and execute each statement separately. One caveat here is we **won't** start a transaction inside Postgres automatically (at least, not yet), so we have to be careful about which statements we can execute safely.

### `SET` mixed with _one_ query

This is totally safe (and quite common):

```sql
SET statement_timeout TO 0; SELECT * FROM users;
```

We can split these up, execute them separately, without breaking transactional guarantees. Only one statement actually reads data, while PgDog handles `SET` internally. This type of query will, as of this PR, be executed normally. Yay!

Conversely, the following cannot be split safely (yet) and will continue to return an error from us:

```sql
SET statement_timeout TO 0; SELECT * FROM users; SELECT * FROM orders;
```

You can set as many things as you want:

```sql
SET statement_timeout TO 0;
SET lock_timeout TO 0;
SELECT * FROM pg_type;
SHOW lock_timeout;
RESET statement_timeout;
RESET lock_timeout;
```

This is safe to execute because only _one_ query actually cares about data, and the rest is handled by PgDog or is inconsequential if executed outside a transaction.

### DDL

```sql
CREATE TABLE x [...];
CREATE TABLE y [...];
```

This is quite common (e.g., in migration scripts) and is quite safe to execute without splitting at all. We send this query to all shards concurrently. No problems.

### Manual transactions

This happens sometimes and now works correctly:

```sql
BEGIN; SELECT * FROM users WHERE id = 1; SELECT * FROM orders WHERE user_id = 2; COMMIT;
```

This is safe to split and execute because client starts a transaction and we can handle those no problem. If the transaction returns an error, the remaining statements are not executed; same behavior as regular Postgres.

### Checks

Other things we check for and block now:

1. Unclosed transactions, e.g., `BEGIN; SELECT 1;`. Bad! Blocked.
2. Multi-statement queries outside transactions, bad, blocked!
3. Mix of transactions and not, bad, blocked!

This is only relevant for a single `Query` message. Normal query flow is not affected by this change.

### DIFF
--- a/integration/rust/tests/integration/multi_set.rs
+++ b/integration/rust/tests/integration/multi_set.rs
@@ -48,8 +48,7 @@ async fn test_multi_set_with_timezone_interval() {
 #[tokio::test]
 async fn test_multi_set_mixed_works() {
     for conn in connections_tokio().await {
-        let _ = conn
-            .batch_execute("SET statement_timeout TO '10s'; SELECT 1")
+        conn.batch_execute("SET statement_timeout TO '10s'; SELECT 1")
             .await
             .unwrap();
     }
--- a/integration/rust/tests/integration/multi_set.rs
+++ b/integration/rust/tests/integration/multi_set.rs
@@ -46,17 +46,11 @@ async fn test_multi_set_with_timezone_interval() {
 }
 
 #[tokio::test]
-async fn test_multi_set_mixed_returns_error() {
+async fn test_multi_set_mixed_works() {
     for conn in connections_tokio().await {
-        let err = conn
+        let _ = conn
             .batch_execute("SET statement_timeout TO '10s'; SELECT 1")
             .await
-            .unwrap_err();
-        let db_err = err.as_db_error().expect("Expected a DbError");
-        let msg = db_err.message();
-        assert!(
-            msg.contains("multi-statement queries cannot mix SET with other commands"),
-            "unexpected error: {msg}",
-        );
+            .unwrap();
     }
 }
--- a/pgdog/src/frontend/client/query_engine/split.rs
+++ b/pgdog/src/frontend/client/query_engine/split.rs
@@ -45,7 +45,7 @@ impl QueryEngine {
     ///
     /// Caller is expected to abort the request and return the result back to the caller
     /// for resubmission.
-    pub(super) fn check_extended_request_split(
+    pub(super) fn check_extended_pipeline_rewrite(
         request: &ClientRequest,
     ) -> Result<Option<QueryEngineResult>, Error> {
         if request.is_multi_exec() {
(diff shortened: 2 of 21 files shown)


integration/rust/tests/integration/multi_set.rs
```diff
@@ -46,17 +46,10 @@ async fn test_multi_set_with_timezone_interval() {
 }
 
 #[tokio::test]
-async fn test_multi_set_mixed_returns_error() {
+async fn test_multi_set_mixed_works() {
```
sgrif, line 49: Love this

pgdog/src/frontend/client/query_engine/split.rs
```diff
@@ -24,9 +66,29 @@ impl QueryEngine {
...
         self.backend.out_of_sync()
             && !context.client_request.is_sync_only()
-            && context.extended_pipeline_requests_left > 0
+            && !context.pipeline.is_done()
+    }
+
+    /// Return true if we should ignore this query because
+    /// the simple query pipeline is in an error state, i.e., inside a failed
+    /// transaction.
+    pub(super) fn simple_pipeline_check(&self, context: &QueryEngineContext<'_>) -> bool {
```
sgrif: What do you think about making this a bit more descriptive?

```suggestion
    pub(super) fn in_errored_simple_pipeline(&self, context: &QueryEngineContext<'_>) -> bool {
```

pgdog/src/frontend/client/query_engine/split.rs
```diff
@@ -1,5 +1,44 @@
...
 use super::*;
-use crate::frontend::ClientRequest;
+use crate::{frontend::ClientRequest, net::Query};
+
+/// Query engine pipeline state.
+pub(crate) enum Pipeline {
+    Extended { requests_left: usize },
+    Simple { requests_left: usize },
+    None,
+}
```
sgrif, line 11: I like this structure. Long term I think it'd be nice to have more of the state bundled here (not asking for that to be changed in this PR)

sgrif (approved):
The transaction_safe check feels a little overboard to me, but this seems fine since we intend to move to implicit transactions down the line. 

With the changes to how planning works that John and I are working on, we're thinking we may end up having communication with the shards *always* use the extended protocol if the query parser is enabled, which will make auto-transactions much easier both here and with any other 1->N rewrites we do going forward.

PR #1434, opened 2026-08-26
fix(query_engine): unwrap type casts in EXECUTE sharding params

### LINKED ISSUE #1401: [Query engine] EXECUTE rewriting doesn't handle type casts
**PgDog version**
`main`

**Description**
#1368 doesn't handle typecast extraction when extracting the sharding key from `EXECUTE` statements:

```sql
EXECUTE stmt1 (1::bigint);
```

**Configuration**

```toml
[general]
prepared_statements = "full"
```

murex971:
- Peel `TypeCast` wrappers when extracting EXECUTE arguments for sharding key routing (e.g. `EXECUTE stmt(1::bigint)`).
- Without this, casted args were treated as NULL and queries were incorrectly broadcast to all shards.
- Fixes #1401

### DIFF
--- a/pgdog/src/frontend/router/parser/params.rs
+++ b/pgdog/src/frontend/router/parser/params.rs
@@ -165,13 +165,7 @@ mod test {
         assert_eq!(params.parameter(14).unwrap().unwrap().text().unwrap(), "3");
         assert_eq!(params.parameter(15).unwrap().unwrap().text().unwrap(), "1");
         assert_eq!(
-            params
-                .parameter(16)
-                .unwrap()
-                .unwrap()
-                .uuid()
-                .unwrap()
-                .to_string(),
+            params.parameter(16).unwrap().unwrap().text().unwrap(),
             "123e4567-e89b-12d3-a456-426614174000"
         );
         assert_eq!(params.parameter(17).unwrap().unwrap().bigint().unwrap(), 5);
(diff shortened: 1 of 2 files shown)


pgdog/src/frontend/router/parser/params.rs
```diff
@@ -117,16 +121,41 @@ mod test {
...
-                assert_eq!(first, 1);
-            }
-
-            _ => panic!("not an execute stmt"),
-        }
+        let stmt = pg_raw_parse::parse(
+            "EXECUTE __pgdog_1 (
+                1, 'test', 1.5, true, false, NULL, B'101', now(),
+                2::bigint, 'x'::text, 3.5::float8, false::bool,
+                NULL::int, (4::int)::bigint
```
meskill: we should test also the cases where the actual typecast happens, like `'3'::int, 1::bool, `13245-4564..`::uuid` it's probably fine right now while we use Format::Text, but that could the be issue if we decide to change it to binary.

also maybe we need to cover more verbose syntax `CAST(value AS type)` for typecast as well

murex971: @meskill added cast coverage, extended tests.

murex971: agreed that binary encoding would need separate handling if parameter format changes :grimacing:

levkk (approved):
This LGTM! @meskill will let you do a final pass and merge :bow:

meskill (approved):

PR #1436, opened 2026-08-26
refactor(api/tasks): add typed definition and state to tasks

meskill:
Make the explicit types for task definitions and state that should be manageable to send as state to the control plane. 

And other improvements and refactors: tasks storage traversal, consistent names and better tasks handling


levkk (approved):
Sure 😆 

Edit: Oh I see what's going on. Yup, LGTM.

PR #1437, opened 2026-08-26
refactor(config): extract pool config creation

meskill:
Move the Pool initialization factory to the pgdog-config crate to be able to reuse e.g. in control plane.

The core of changes is the new pool.rs in the pgdog-config and the remaining is the refactoring for this moving and some types renaming and simplification


levkk (approved):
Nice! Makes things so much easier.

PR #1438, opened 2026-08-26
fix(admin): preserve casing for admin command string

meskill:
The original `Parser::parse(&query.query().to_lowercase())` can cause issue if we rely on what is inside the query. So, move the normalization inside the parser and not the caller side

### DIFF
--- a/pgdog/src/admin/parser.rs
+++ b/pgdog/src/admin/parser.rs
@@ -157,13 +157,18 @@ pub(crate) struct Parser;
 impl Parser {
     /// Parse the query and return a command we can execute.
     pub(crate) fn parse(sql: &str) -> Result<ParseResult, Error> {
+        let sql = sql.trim();
+
         // Handle SET separately because
         // we're about to clobber valid SQL syntax below.
         if is_set_statement(sql) {
-            return Ok(ParseResult::Set(Set::parse(sql)?));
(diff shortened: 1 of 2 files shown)


pgdog/src/admin/parser.rs
```diff
@@ -157,13 +157,18 @@ pub struct Parser;
 impl Parser {
     /// Parse the query and return a command we can execute.
     pub fn parse(sql: &str) -> Result<ParseResult, Error> {
+        let sql = sql.trim();
```
sgrif, line 160: What do you think about making having `is_set_statement` and `get_show_variable` work on lowercased strings so we can just have the single `to_lowercase` up here?

sgrif (approved):

pgdog/src/admin/parser.rs
sgrif: Or actually, what do you think about having `let normalized_sql = sql.trim().to_lowercase()` to make it more clear whether we're using the modified or the preserved original and have to worry less about what modifications were run when?

meskill: just don't want to mess up this again - if we do normalization upfront we can reuse it where it's not needed or it'll be harder on ee path to undo it or embed the ee specific code somewhere in case we'll need the original query

PR #1441, opened 2026-08-26
Perform the great un-pub-ening

sgrif:
And thus, as it was fortold, the acolytes of encapsulation did gather. And on the twenty-sixth day of the eight month, they recited that most sacred of incantations:

`find pgdog/src -type f -name '*.rs' | xargs sed -i 's/\<pub /pub(crate) /g' && cargo fmt`

And the spirits of encapsulation heard their prayer, and thus, their world was `pub` no longer.

And they rejoiced, as the lords of compilation would once again see their dead (code).

And they rejoiced, for future generations will begin to consider API boundaries more explicitly.

And they rejoiced, for their jouney to reduce intermingling between `frontend` and `backend` could begin.

But for today, they knew they could look upon their works and smile.


levkk (approved):
Image

PR #1442, opened 2026-08-26
fix: use keyless cosign attestations

dkarter:
# Motivation

Publish verifiable image signatures and SLSA provenance without depending on GitHub's artifact-attestation service, which requires GitHub Enterprise Cloud for private repositories. This provides a reference that can also be used by the private Enterprise and Control image workflows.

# Summary of Changes

- Add a reusable composite action that keylessly signs image digests with Cosign.
- Publish SLSA v1 provenance for architecture manifests and the final multi-architecture index.
- Remove the GitHub-specific `attestations: write` permission while retaining OIDC and package permissions.
- Pin Cosign Installer to the verified v4.1.2 commit and Cosign to v3.0.6.

# Testing

- YAML parsing and `git diff --check` pass locally.
- After a workflow run, verify the final digest with `cosign verify` and `cosign verify-attestation --type slsaprovenance1`, constrained to the package workflow identity and GitHub Actions OIDC issuer.

# Dependencies/Special Considerations

Keyless signing records the repository and workflow identity in Sigstore's public transparency log. This is harmless for this public repository, but should be an explicit decision before copying the action into a private repository.

### DIFF
--- a/.github/actions/sign-and-attest-image/action.yml
+++ b/.github/actions/sign-and-attest-image/action.yml
@@ -1,0 +1,67 @@
+name: Sign and attest image
+description: Keylessly sign an image digest and publish SLSA v1 provenance
+
+inputs:
+  image:
+    description: Fully qualified image digest
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Install Cosign
+      uses: sigstore/cosign-installer@6f9f17788090df1f26f669e9d70d6ae9567deba6 # v4.1.2
+      with:
+        cosign-release: v3.0.6
+
+    - name: Sign and attest image
+      env:
+        IMAGE: ${{ inputs.image }}
+        BUILDER_ID: https://github.com/${{ github.workflow_ref }}
+        INVOCATION_ID: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}/attempts/${{ github.run_attempt }}
+        REPOSITORY_ID: ${{ github.repository_id }}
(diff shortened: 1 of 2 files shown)


.github/actions/sign-and-attest-image/action.yml
```diff
@@ -0,0 +1,67 @@
...
+inputs:
+  image:
+    description: Fully qualified image digest
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Install Cosign
+      uses: sigstore/cosign-installer@6f9f17788090df1f26f669e9d70d6ae9567deba6 # v4.1.2
```
dkarter, line 13: they don't use immutable releases on github (which are literally a checkbox) 🤦🏼‍♂️ so i pinned to the sha for this release:
https://github.com/sigstore/cosign-installer/releases/tag/v4.1.2

levkk: I wonder if we need to check that box too... :laughing:

dkarter: yea it's probably a great idea 😄, especially for a database pooler (due to the security surface area).

although most people that install pgdog will likely be doing it via a docker image (so for us that's more important). For github actions that's super important IMO.

i think it should be the default for releases honestly.. 

I haven't looked thoroughly at your release process - but it might require making a change:
- will first publish the release as draft
- attach the assets
- then mark the released published

PR #1443, opened 2026-08-27
feat(stats): add banned gauge to pool metrics

### LINKED ISSUE #1424: Metric to indicate the banned status of a database
I'm looking to write an alert for when a database is banned. Initially I tried using 
```
sv_active + sv_idle + sv_idle_xact + sv_locked == 0
```
though I found these metrics track the sockets providing the database connections and that, depending upon the exact error, a socket may still be in an established state even though the health check is failing. Is there an existing metric for detecting a banned database? If not, could one be added?

ChrisJr404:
Closes #1424. Adds a `banned` gauge to the per-pool OpenMetrics output so you can alert on a database being banned instead of inferring it from the socket counts, which don't reliably drop to zero for every kind of failure.

It reuses the same ban state that `SHOW BANS` reports (switched the metrics loop to `pools_with_roles_and_bans`), emitting one measurement per pool with the usual database/user/host/shard/role labels: 1 when the pool is banned, 0 when it's available. Added a render test next to the existing pool metric tests.

### DIFF
--- a/pgdog/src/stats/pools.rs
+++ b/pgdog/src/stats/pools.rs
@@ -52,6 +52,7 @@ impl Pools {
         let mut maxwait = vec![];
         let mut errors = vec![];
         let mut out_of_sync = vec![];
+        let mut banned = vec![];
         let mut total_xact_count = vec![];
         let mut total_xact_2pc_count = vec![];
         let mut avg_xact_count = vec![];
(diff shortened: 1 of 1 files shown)


levkk (approved):
Cool! Happy to merge if you sign the CLA.

PR #1444, opened 2026-08-27
Fix role leakage between connections

### LINKED ISSUE #1341: Transaction pooling leaks `SET ROLE` between connections
**PgDog version**

`0.1.52` (`dd37c05`, transaction pool mode)

**Description**

In transaction pooling, a frontend can acquire a session-level advisory lock, execute `SET ROLE`, and disconnect. PgDog returns the physical backend to the pool, but the next frontend can inherit the previous frontend's `current_user`.

The session-level advisory lock pins the source backend. PgDog marks a pinned binding dirty [`connection/mod.rs`](https://github.com/pgdogdev/pgdog/blob/dd37c05/pgdog/src/backend/pool/connection/mod.rs#L430-L437), which marks its server dirty [`binding.rs`](https://github.com/pgdogdev/pgdog/blob/dd37c05/pgdog/src/backend/pool/connection/binding.rs#L484-L490). Closing the frontend therefore sends that backend through dirty cleanup.

For a dirty server, PgDog runs `RESET ALL`, `pg_advisory_unlock_all()`, and `DISCARD TEMP` [`cleanup.rs`](https://github.com/pgdogdev/pgdog/blob/dd37c05/pgdog/src/backend/pool/cleanup.rs#L14-L20). PostgreSQL marks `role` with `GUC_NO_RESET_ALL` [`guc_tables.c`](https://github.com/postgres/postgres/blob/REL_16_9/src/backend/utils/misc/guc_tables.c#L4157-L4166), and `ResetAllOptions()` skips settings with that flag [`guc.c`](https://github.com/postgres/postgres/blob/REL_16_9/src/backend/utils/misc/guc.c#L1981-L1987). The backend therefore returns to the pool with `SET ROLE` still active, allowing the next frontend to inherit the previous `current_user`.

**Reproduce**

1. Start PostgreSQL 16 and PgDog in transaction pool mode.
2. Create the target role and grant membership to `app`:

```sql
CREATE ROLE scm_session_role NOLOGIN;
GRANT scm_session_role TO app;
```

3. Run the following with the direct and PgDog DSNs supplied by the environment:

```python
#!/usr/bin/env python3
import psycopg


    try:
        with psycopg.connect(dsn, autocommit=True) as source:
            with source.cursor() as cur:
                cur.execute("SELECT pg_advisory_lock(707519)")
                cur.execute("SET ROLE scm_session_role")
                cur.execute("SELECT pg_backend_pid(), session_user, current_user")
                before = cur.fetchone()

        with psycopg.connect(dsn, autocommit=True) as peer:
            with peer.cursor() as cur:
                cur.execute("SELECT pg_backend_pid(), session_user, current_user")
                after = cur.fetchone()

        print(f"{label}: source={before[1:]!r} peer={after[1:]!r} pid_same={before[0] == after[0]}")
    except Exception as exc:


run("direct", DIRECT_DSN)
run("pgdog", PGDOG_DSN)
```

**Expected behavior**

Both peer connections should report `('app', 'app')`.

**Actual behavior**

Direct PostgreSQL reports a clean peer:

```text
direct: source=('app', 'scm_session_role') peer=('app', 'app') pid_same=False
```

PgDog reuses the backend with its previous identity:

```text
pgdog: source=('app', 'scm_session_role') peer=('app', 'scm_session_role') pid_same=True
```

Replacing the `SET ROLE` line with `SET SESSION AUTHORIZATION scm_session_role` is a related variant: it produces `peer=('scm_session_role', 'scm_session_role')` through PgDog and `peer=('app', 'app')` directly. That variant requires `app` to be a superuser.

nsavvide:
Fixes #1341.

The bug

In transaction pooling, a client that pins its backend and runs SET ROLE leaks its current_user to the next client handed that backend:

```
direct: source=('app','scm_session_role') peer=('app','app')             pid_same=False
pgdog:  source=('app','scm_session_role') peer=('app','scm_session_role') pid_same=True
```

The next client executes queries as a role it never asked for and may not be entitled to.

Root cause, see [scoping](https://github.com/pgdogdev/pgdog/issues/1341#issuecomment-5440638271)

The fix is adding one statement to the `DIRTY` cleanup list:

`Query::new("RESET SESSION AUTHORIZATION")`

On your (@levkk) RESET ROLE suggestion: I implemented and tested that first. RESET SESSION AUTHORIZATION makes all tests green, because per the docs SET SESSION AUTHORIZATION also resets role — so `RESET_ROLE` alongside it is dead weight and I dropped it. Happy to switch to RESET ROLE instead if you'd rather keep the change narrower in intent, but then the SET SESSION AUTHORIZATION variant documented in the issue stays broken: it leaks session_user as well, and session_authorization is in -out path never resets it either.

### DIFF
--- a/pgdog/src/backend/pool/cleanup.rs
+++ b/pgdog/src/backend/pool/cleanup.rs
@@ -14,6 +14,7 @@ static PREPARED: Lazy<Vec<Query>> = Lazy::new(|| vec![Query::new("DEALLOCATE ALL
 static DIRTY: Lazy<Vec<Query>> = Lazy::new(|| {
     vec![
         Query::new("RESET ALL"),                       // Reset all parameters.
+        Query::new("RESET SESSION AUTHORIZATION"),     // Reset all skips session_authorization.
         Query::new("SELECT pg_advisory_unlock_all()"), // Remove all advisory locks.
         Query::new("DISCARD TEMP"),                    // Drop all temporary tables.
     ]
--- a/pgdog/src/backend/pool/guard.rs
+++ b/pgdog/src/backend/pool/guard.rs
@@ -314,6 +314,83 @@ mod test {
         drop(guard);
     }
 
+    /// A client set a role and the connection was returned dirty. The next client
+    /// must not inherit it.
+    ///
+    /// `RESET ALL` does not clear `role`: Postgres flags it `GUC_NO_RESET_ALL` and
+    /// `ResetAllOptions()` skips such settings, so cleanup has to reset it explicitly.
+    ///
+    /// <https://github.com/pgdogdev/pgdog/issues/1341>
+    #[tokio::test]
+    async fn test_cleanup_resets_role() {
+        crate::logger();
+        let pool = pool();
+
+        let mut guard = pool.get(&Request::default()).await.unwrap();
+        let pid_before: Vec<i32> = guard.fetch_all("SELECT pg_backend_pid()").await.unwrap();
+
+        guard.execute_checked("SET ROLE pgdog1").await.unwrap();
+        let role: Vec<String> = guard.fetch_all("SELECT current_user").await.unwrap();
+        assert_eq!(role, vec!["pgdog1".to_string()]);
+
+        guard.mark_dirty(true);
+        drop(guard);
+
+        // Our test pool is only 1 connection, so this is the same backend.
+        let mut guard = pool.get(&Request::default()).await.unwrap();
+        let pid_after: Vec<i32> = guard.fetch_all("SELECT pg_backend_pid()").await.unwrap();
+        assert_eq!(
+            pid_before, pid_after,
+            "1-connection pool should hand back the same backend"
+        );
+
+        let role: Vec<String> = guard.fetch_all("SELECT current_user").await.unwrap();
+        assert_eq!(
+            role,
+            vec!["pgdog".to_string()],
+            "SET ROLE leaked across a dirty check-in"
+        );
+    }
+
+    /// Same as [`test_cleanup_resets_role`], for `SET SESSION AUTHORIZATION`.
+    ///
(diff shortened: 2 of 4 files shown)


nsavvide:
@levkk feel free to review this one, thanks :))!

levkk:
> that pins its backend and runs SET ROLE

This is an edge case I think. The real bug is if a client runs `SET ROLE` in any scenario, and I suspect, in that case, the role will leak to the next client as well. We need to detect `SET ROLE` in the parser and place it in the `Parameters` along with others (or have it handled separately) and make sure we check it on each connection checkout from the pool.

nsavvide:
> > that pins its backend and runs SET ROLE
> 
> This is an edge case I think. The real bug is if a client runs `SET ROLE` in any scenario, and I suspect, in that case, the role will leak to the next client as well. We need to detect `SET ROLE` in the parser and place it in the `Parameters` along with others (or have it handled separately) and make sure we check it on each connection checkout from the pool.

You're right, I reproduced it. SET ROLE issued after a backend is attached leaks with no pin involved:
```
BEGIN;
SELECT 1;          -- backend is attached from here
SET ROLE pgdog1;   -- plain SET, survives COMMIT in Postgres
COMMIT;
```
Next client on that backend gets current_user = pgdog1 fails with this PR's fix applied, so check-in cleanup can't reach it. Worth noting what is already working: role is not in UNTRACKED_PARAMS, the parser already produces Command::Set { name: "role" }, and link_client already emits RESET "role" at check-out. A SET ROLE outside a transaction is therefore already handled on main; I have a passing test for it. The hole is specifically that a SET issued once a server is attached is never written to client_params (populated only at check-out, server.rs:731), so the next check-out has nothing to reset.

That's the same hole #1299 is fixing for parameters generally.

So: how would you like this split?

1. Land this PR for the dirty-check-in case, cover the attached-server case in #1299.
2. Close this, fold role into #1299.
3. I extend #1299 to cover role and session_authorization.

PR #1449, opened 2026-08-27
feat: re-prepare statements after schema changes

### LINKED ISSUE #943: Re-prepare statements after schema changes

jkaczman:
Previously, when we encountered a statement that was previously `Parse`'d (and put into global cache), and that was invalidated due to a schema change, we always returned an error to the Client, even in recoverable situations.

This introduces functionality that internally re-prepares when we're not in a transaction.

It covers both situations where we may come across this error, being [`Bind`, `Execute`] and [`Parse`, `Describe`] pairs, and after validating all conditions are met:
- fetches the `Parse`, 
- sends a Close with the statement to clear the invalidated cache.
- sends a [`Parse`] to the relevant `Route`, caching statement again.
- retries the Client's original messages again 

All cases should be tested (`Parse` + `Describe`, multi shard, regular) in the integration test.

Fixes https://github.com/pgdogdev/pgdog/issues/943.


pgdog/src/frontend/client/query_engine/query.rs
```diff
@@ -156,6 +152,35 @@ impl QueryEngine {
             self.pending_explain = None;
         }
 
+        // NOTE: This code must be ordered to run before context.transaction is modified
+        if message.code() == 'E'
```
levkk: We already do this here: https://github.com/pgdogdev/pgdog/blob/5d8152273c066f7cd36003c79477fa58c7c350d0/pgdog/src/backend/server.rs#L640

You should just propagate the `schema_changed` flag up from `self.backend`.

pgdog/src/frontend/client/query_engine/query.rs
```diff
@@ -268,6 +296,177 @@ impl QueryEngine {
...
 
+    /// If we encounter a cache invalidated error & we're either:
+    ///
+    /// 1. In a transaction, but we've not encountered a Bind & Execute before this statement,
+    ///    we haven't encountered any simple protocol messages,
+    ///    and we don't have parameters local to the transaction
+    /// 2. Not in a transaction
+    ///
+    /// Then, handle it on our end instead of forwarding the error to the client.
+    pub(crate) async fn retry_statement(
```
levkk: There is a lot of duplicated logic here.

It's a lot simpler if you only do this if you're not inside a transaction only. That should cover a lot of use cases.

Also, you don't need to invalidate the statement, you just need to re-prepare it on the server, which can be achieved with a sequence of:

```
Close [1]
Parse
```

1: https://github.com/pgdogdev/pgdog/blob/5d8152273c066f7cd36003c79477fa58c7c350d0/pgdog/src/net/messages/close.rs#L1

Just send it to all connected servers by calling `close_many` (which bypasses server check that ignores all client-sent `Close` messages): https://github.com/pgdogdev/pgdog/blob/5d8152273c066f7cd36003c79477fa58c7c350d0/pgdog/src/backend/server.rs#L1045

pgdog/src/frontend/client/transaction_type.rs
```diff
@@ -1,5 +1,57 @@
+use crate::frontend::ClientRequest;
+use crate::net::Protocol;
+
+/// Represents the state we hold for the current Transaction;
+/// the type and if it's reversible (in its current state)
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Transaction {
+    transaction_type: TransactionType,
+    reversible: bool,
```
levkk: Not entirely sure this is worth the trouble, since this won't cover the vast majority of transactions. I could be convinced otherwise though.

pgdog/src/frontend/client/query_engine/query.rs
levkk: Make sure to ignore `ParseComplete` received from the server in this scenario (i.e., send `Parse` and maybe `Flush`, which will force Postgres to execute the command immediately).

jkaczman: Got it. Thanks! Which would you prefer: dropping support for transactions or polishing the code to be less duplicated/cleaner?

levkk: Probably both to get started! Check this out: https://github.com/pgdogdev/pgdog/pull/970/

This was a previous attempt at this. TLDR: it was entirely managed inside `Server` by closing and re-preparing the statement. I think that came pretty close, it just needed to check that we were not inside a transaction and also needed to be a bit cleaner.

jkaczman: I changed it up a bit, please let me know how it looks now. I could also look into moving it into `Server` if you'd like.

pgdog/src/frontend/client/query_engine/query.rs
```diff
@@ -258,16 +268,140 @@ impl QueryEngine {
...
+
+        // Send the Parse and a Flush to cache the new statement
+        {
+            let mut client_request = ClientRequest::default();
+            client_request.route = context.client_request.route.clone();
+
+            client_request.push(ProtocolMessage::from(parse.clone()));
+            client_request.push(ProtocolMessage::from(messages::Flush));
+
+            self.handle_request_and_maybe_forward(client_request, context, false)
```
levkk: I think this can be as easy as:

```rust
self.backend.send([Parse, Flush]).await?;
let parse_complete = self.backend.read().await?;
if parse_complete.code() =! '1' {
  // error, parse and return to client
}
```

I don't think you need to push that through the whole send/recv query engine loop.

pgdog/src/frontend/client/query_engine/query.rs
```diff
@@ -156,6 +157,15 @@ impl QueryEngine {
             self.pending_explain = None;
         }
 
+        if code == 'E'
+            && self.backend.schema_changed()
+            && context.transaction().is_none()
+            && !self.retrying
+            && Box::pin(self.retry_statement(context)).await?
```
levkk: What about returning `QueryEngineResult::Retry` here? You could close and re-prepare the statement
and then just ask the query engine to re-submit the request, while keeping the servers connected. Seems easier than setting a flag on the engine for an edge case.

pgdog/src/frontend/client/mod.rs
```diff
@@ -620,6 +620,9 @@ impl Client {
                     }
                 }
             }
+            QueryEngineResult::Retry => {
+                return Box::pin(self.client_messages(query_engine)).await;
```
levkk: This could potentially cause infinite recursion. I think you probably want to handle it as `Split` above: try to execute once, and if you don't get `Done`, return an error.

pgdog/src/frontend/client/query_engine/query.rs
```diff
@@ -268,6 +280,61 @@ impl QueryEngine {
...
+            .await?;
+
+        // Send the Parse and a Flush to cache the new statement
+        let mut client_request = ClientRequest::default();
+        client_request.route = context.client_request.route.clone();
+
+        client_request.push(ProtocolMessage::from(parse.clone()));
+        client_request.push(ProtocolMessage::from(messages::Flush));
+
+        self.backend.send(&client_request).await?;
```
levkk: Could use `send_ignore` here instead: it faster and was built to re-prepare the statement on a connection: https://github.com/pgdogdev/pgdog/blob/main/pgdog/src/backend/server.rs#L527

pgdog/src/frontend/client/query_engine/query.rs
```diff
@@ -268,6 +280,61 @@ impl QueryEngine {
         Ok(())
     }
 
+    /// If we encounter a cache invalidated error & we're not in a transaction,
+    /// then, handle it on our end instead of forwarding the error to the client.
+    pub(crate) async fn retry_statement(
```
levkk: I would call this `reprepare_statement` probably since we are no logner retrying here.

pgdog/src/backend/pool/connection/binding.rs
```diff
@@ -42,6 +42,58 @@ impl Binding {
...
+        }
+    }
+
+    // Reset the schema changed flag on the server(s) the client is connected to.
+    pub(crate) fn reset_schema_changed(&mut self) {
+        match self {
+            Binding::Direct(shard, _) => shard.reset_schema_changed(),
+            Binding::MultiShard(shards, _) => shards
+                .iter_mut()
+                .for_each(|shard| shard.reset_schema_changed()),
```
sgrif, line 60: It's a minor style nit, but I'd use a for loop here as you do below. 

```suggestion
            Binding::MultiShard(shards, _) => {
                for shard in shards {
                    shard.reset_schema_changed();
                }
            }
```

pgdog/src/frontend/client/query_engine/query.rs
```diff
@@ -250,14 +261,29 @@ impl QueryEngine {
...
+        let prepare_family = matches!(code, '1' | 't' | 'T' | 'n');
+        let duplicate = prepare_family
+            && self
+                .suppress
+                .iter()
+                .position(|c| *c == code)
+                .map(|position| {
+                    self.suppress.remove(position);
+                })
+                .is_some();
```
sgrif, line 274: Can we separate the mutation out from the conditional? I had to stare at this a bit too long to understand what it's doing. I think the density is making it harder to follow.

pgdog/src/frontend/client/query_engine/query.rs
```diff
@@ -156,6 +157,16 @@ impl QueryEngine {
             self.pending_explain = None;
         }
 
+        if code == 'E'
+            && self.backend.schema_changed()
+            && context.transaction().is_none()
+            && Box::pin(self.reprepare_statement(context)).await?
+        {
+            self.suppress = std::mem::take(&mut self.delivered);
```
levkk, line 165: Could you provide an example of messages that are being suppressed here? I'm not entirely clear on why this is necessary. For example, if the client sends `Bind` and `Execute`, the server will respond with `ErrorResponse` on `Bind` immediately, and the client won't receive any messages except that.

I believe the same happens with `Describe` and `Flush`, or is that not the case? i.e., something like this would need message supression: `Describe`, `Bind`, `Execute`, `Sync`. `Describe` succeeds but `Bind` fails and we need to retry while suppressing `T` and `t` we received from the server?

pgdog/src/frontend/client/query_engine/mod.rs
```diff
@@ -61,6 +61,11 @@ pub(crate) struct QueryEngine {
     stats: Stats,
     backend: Connection,
     streaming: bool,
+    // Track ParseComplete / ParameterDescription / RowDescription / NoData
+    // already received for the present ClientRequest
+    delivered: Vec<char>,
+    // Delivered by the previous attempt of a retried request
+    suppress: Vec<char>,
```
sgrif, line 68: Not sure how I feel about tracking these separately since we should generally be sending exactly one of these. I guess if we got multiple `Describe` messages from the client we would be resending the `ParameterDescription`/`RowDescription`/`NoData` multiple times but that feels extremely niche.

...This is actually making me realize why this error even exists in the first place and is making me realize it might not be safe for us to do this at all. When this error occurs, we've already sent a `ParameterDescription` and `RowDescription` to the client (either immediately before this or much earlier and they're caching it), but this error means the value of one of those messages will have changed, so the client will end up sending the wrong data to their `Bind` or interpreting the results incorrectly when they receive their `DataRow`. @levkk Have you thought about this?

pgdog/src/frontend/error.rs
```diff
@@ -21,6 +21,9 @@ pub(crate) enum Error {
     #[error("{0}")]
     Router(#[from] super::router::Error),
 
+    #[error("prepared statement invalidated again after retry, giving up")]
```
sgrif, line 24: We need to make sure we're handling this error up the stack and returning essentially the original error that we got from PG. This isn't a good user-facing error message.

sgrif (commented):
Other than some style nits this generally looks good, but the course of reviewing this made me realize that this might be tremendously unsafe for us to do at all.

pgdog/src/frontend/client/query_engine/mod.rs
levkk: Yes, I thought the same just now. Can we even safely re-prepare the statement in this case since whatever `RowDescription` the client cached from way back when doesn't match the new table schema anymore?

sgrif: I think the *only* time we could ever safely reprepare it is if the error occurs on the very first `Describe` which is exceptionally rare. We'd essentially only be able to solve this for clients that do no client side prepared statement caching, which are the clients least likely to hit this in the first place

PR #1452, opened 2026-08-28
feat: configurable listener backlog to allow more incoming connection requests

ryderhwang:
## Problem

`Listener::listen` binds via `TcpListener::bind`, which hardcodes `listen(1024)`. The accept queue is therefore capped at 1024 pending connections regardless of `net.core.somaxconn`. Raising the sysctl has no effect past the application's own argument.

For deployments where clients open a connection per request (e.g. Django with `CONN_MAX_AGE=0` behind PgDog), a rolling deploy of a large client fleet reconnects everyone at once. With `tcp_abort_on_overflow=0` (the Linux default), overflowed handshakes are silently dropped.

## Change

New `[general]` setting `listen_backlog`, default to 1024.

The kernel still caps the effective value at `net.core.somaxconn`, so operators raise both together; the doc comment says so.

### DIFF
--- a/.schema/pgdog.schema.json
+++ b/.schema/pgdog.schema.json
@@ -59,6 +59,7 @@
         "idle_healthcheck_delay": 5000,
         "idle_healthcheck_interval": 30000,
         "idle_timeout": 60000,
+        "listen_backlog": 1024,
         "load_balancing_strategy": "random",
         "load_schema": "auto",
         "log_connections": true,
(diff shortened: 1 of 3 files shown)


levkk:
:pray: Ping me when the PR is good to go, I'll run a final CI pass on it.

ryderhwang:
@levkk good to go 🙏 Since the first CI pass: regenerated the JSON schema (`cargo run -p pgdog-jsonschema`), and added unit tests covering the new bind path — v4/v6 bind, lookup failure, and the empty-resolution error (extracted a small `first_addr` helper so that branch is directly testable). Build, tests, fmt, and clippy are clean locally on the pinned toolchain; CLA is signed. Default behavior is unchanged (backlog stays 1024 unless `listen_backlog` / `PGDOG_LISTEN_BACKLOG` is set).

ryderhwang:
One more commit after a self-review pass (6b640e4): restored `TcpListener::bind`'s multi-address fallback — binding by hostname now tries each resolved address in order instead of only the first (matters for e.g. `host = "localhost"` resolving to `::1` then `127.0.0.1`), with tests for the fallback, the exhausted-candidates error, and empty resolution. That's the final change from my side — ready for your CI pass whenever you are.

levkk:
LGTM on the implementation. Quick question, do you think this setting is better put into the `[tcp]` config block? https://github.com/pgdogdev/pgdog/blob/main/pgdog-config/src/networking.rs#L50

ryderhwang:
Considered `[tcp]` initially, but landed on `[general]` for two reasons:

1. Everything in `[tcp]` today is a **per-connection** option — applied by `tweak()` to each accepted client/server socket — so those settings effectively pick up config reloads on new connections. `listen_backlog` configures only the listening socket and can never apply on reload, so in `[tcp]` it'd be the odd one out on both counts.
2. It tunes the same socket that `host`/`port` define, and `[general]` already carries the "cannot be changed at runtime" family (host, port, workers).

No strong feelings though — if you prefer mechanism-grouping under `[tcp]`, it's a one-line move plus schema regen and I'll push it right away.

PR #1457, opened 2026-08-31
feat: support SCRAM-SHA-256-PLUS for client connections

### LINKED ISSUE #1453: Support SCRAM-SHA-256-PLUS (channel binding) for client connections
PgDog advertises a single SASL mechanism, `SCRAM-SHA-256` (`Authentication::scram()`, sent from `frontend/client/mod.rs`), so any client asking for channel binding gets refused:

```
psql: error: connection to server at "..." failed: channel binding required but not supported by server's authentication request
```

Please advertise `SCRAM-SHA-256-PLUS` alongside it on connections PgDog terminates TLS for, using `tls-server-end-point`, and honour clients that require it.

#530 added the gs2 header parsing as "the missing header parsing logic for optional channel binding we don't use", so clients that merely offer binding already connect. This is asking for the rest of it.

### Why

Channel binding gives a pooled client MITM protection with no trust store on its side, and the pooler terminating TLS is the only thing in the path that can offer it.

The usual alternative, telling people to use `sslmode=verify-full`, isn't portable. Measured on macOS against an endpoint whose cert chains to a public CA:

| Client | `require` | `verify-full` | `verify-full` + `sslrootcert=system` |
|---|---|---|---|
| psql / libpq 18.6 | ok | wants `~/.postgresql/root.crt` | ok |
| psycopg2/3 (wheels, libpq 17.0.9) | ok | wants `~/.postgresql/root.crt` | `certificate verify failed`, unless `SSL_CERT_FILE` is set |
| pgjdbc 42.7.9 | ok | wants `~/.postgresql/root.crt` | `Could not open SSL root certificate file system` |
| node-postgres 8.23.0 | ok | ok | `ENOENT: open 'system'` |
| Go pgx v5 | ok | ok | ok |

Two of five verify unaided. `sslrootcert=system` isn't a portable repair either: it needs libpq 16+ (the sentinel appears in `fe-secure-openssl.c` on `REL_16_STABLE` and not on 15), pgjdbc has no such keyword, node-postgres reads it as a path and throws before it opens a socket, and where it does work it only resolves to OpenSSL's default paths, which the psycopg wheels don't aim at the platform's roots.

So on a public endpoint you pick between a string most clients can paste and one that authenticates the server. Channel binding removes that choice, and costs the client one URI parameter.

Neon landed in the same place, for the same cross-platform reasons: "We've now added channel_binding=require (on top of sslmode=require) to the connection strings we show in our console"
https://neon.com/blog/postgres-needs-better-connection-security-defaults

Happy to test a branch and report back

abnegate:
## Summary
- On connections PgDog terminates TLS for, advertise `SCRAM-SHA-256-PLUS` first (then `SCRAM-SHA-256`), matching PostgreSQL.
- Bind PLUS handshakes to `tls-server-end-point` from the leaf cert presented on that connection, snapshotted at accept so a later TLS reload cannot change the hash mid-login.
- Clients that pick `SCRAM-SHA-256` still work. PLUS is not advertised on plaintext, or when the cert's signature algorithm does not name a single hash (Ed25519 / RSASSA-PSS).

Fixes https://github.com/pgdogdev/pgdog/issues/1453

`scram` is pinned to merged `pgdogdev/scram` @ `ee15a47` (https://github.com/pgdogdev/scram/pull/6).

## Test plan
- [x] AuthenticationSASL encodes `SCRAM-SHA-256-PLUS\0SCRAM-SHA-256\0\0` and still encodes SHA-256-only for plaintext
- [x] `tls-server-end-point` for `tests/tls/cert.pem` matches SHA-256 of the DER
- [x] Every named RFC 5929 signature OID maps to the right digest length; Ed25519 / RSASSA-PSS / garbage DER return `None`
- [x] Hashed-password PLUS handshake succeeds / wrong password fails
- [x] `handle()` refuses PLUS with no cbind, unknown mechanisms, GS2 `y`, non-SASL initial, ErrorResponse, and a SASLInitial sent as client-final
- [x] Plain-password PLUS `handle()` accepts the correct password
- [x] Empty AuthenticationSASL mechanism list is `UnexpectedPayload`
- [x] `scram` pin points at `pgdogdev/scram` @ `ee15a47`
- [x] `integration/rust` tokio-postgres client: `sslmode=require` + `channel_binding=require` (plain + hashed users); wrong password fails; `channel_binding=disable` still works over TLS
- [x] `channel_binding=require` against this branch as a TLS PgDog (psql 18.6 / rust-postgres 0.19)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4098,7 +4098,7 @@ checksum = "94143f37725109f92c262ed2cf5e59bce7498c01bcc1502d7b9afe439a4e9f49"
 [[package]]
 name = "scram"
 version = "0.6.0"
-source = "git+https://github.com/abnegate/scram.git?rev=b641be3932a3e4b24705aa3a6848d4aa0e8d2fa9#b641be3932a3e4b24705aa3a6848d4aa0e8d2fa9"
+source = "git+https://github.com/pgdogdev/scram.git?rev=ee15a47f1051dc79589e8db666be7014c1e23fba#ee15a47f1051dc79589e8db666be7014c1e23fba"
 dependencies = [
  "aws-lc-rs",
  "base64 0.13.1",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4098,7 +4098,7 @@ checksum = "94143f37725109f92c262ed2cf5e59bce7498c01bcc1502d7b9afe439a4e9f49"
 [[package]]
 name = "scram"
 version = "0.6.0"
-source = "git+https://github.com/pgdogdev/scram.git?rev=053e108210a35232efe60023ec43288fd601d8c4#053e108210a35232efe60023ec43288fd601d8c4"
+source = "git+https://github.com/abnegate/scram.git?rev=b641be3932a3e4b24705aa3a6848d4aa0e8d2fa9#b641be3932a3e4b24705aa3a6848d4aa0e8d2fa9"
 dependencies = [
  "aws-lc-rs",
  "base64 0.13.1",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2439,8 +2439,10 @@ dependencies = [
  "chrono",
  "futures-util",
  "libc",
+ "native-tls",
  "ordered-float",
  "parking_lot",
+ "postgres-native-tls",
  "rand 0.9.4",
  "reqwest",
  "rust_decimal",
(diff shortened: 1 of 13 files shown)


levkk:
Could you add a client integration test? We could use the Rust client in `integration/rust`.

abnegate:
Added a tokio-postgres client test in `integration/rust/tests/integration/scram_plus.rs`. It connects with `sslmode=require` and `channel_binding=require` (so the rust client will only succeed if we advertised PLUS and the `tls-server-end-point` binding matches), plus the hashed-password user, a wrong-password refusal, and `channel_binding=disable` over TLS so SHA-256 still works when the client opts out.

abnegate:
Ran the last test-plan item against this branch (`feat/scram-sha-256-plus` @ `0cca725e`) as a TLS PgDog (`auth_type=scram`, `integration/tls/cert.pem`).

- rust-postgres 0.19 with `sslmode=require` + `channel_binding=require`: plain user and hashed user both completed `SELECT 1`; wrong password failed; `channel_binding=disable` still worked over TLS
- psql 18.6 (libpq): the same four cases

`channel_binding=require` no longer hits `channel binding required but not supported by server's authentication request`.

PR #1458, opened 2026-08-31
fix: clear prepared statement cache on DISCARD

### LINKED ISSUE #1402: DISCARD does not remove statement from client's cache
**PgDog version**
`main`

**Description**
When client calls `DISCARD`, it doesn't remove the statement from the client's cache or update the global counter. For clients that are connected for a long time, this can cause a slow memory leak.

**Configuration**

```toml
[general]
prepared_statements = "full"
```

murex971:
## Summary
- `DISCARD` was acknowledged without dropping the client's prepared statements or decrementing the global use count, so long-lived clients could leak cache entries.
- Reuse `PreparedStatements::close_all()` on `DISCARD`, the same cleanup already used on disconnect.
- closes #1402 

## Test plan
- [x] `cargo nextest run --profile dev test_discard_clears_prepared_statement_cache`

### DIFF
--- a/integration/ruby/ar_spec.rb
+++ b/integration/ruby/ar_spec.rb
@@ -204,6 +204,9 @@ describe 'active record' do
         end
 
         ActiveRecord::Base.connection.execute 'DISCARD ALL'
+        # DISCARD ALL deallocates prepared statements, so ActiveRecord has to
+        # forget the names it cached.
+        ActiveRecord::Base.connection.clear_cache!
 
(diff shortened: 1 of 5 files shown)


levkk (approved):
According to https://www.postgresql.org/docs/current/sql-discard.html we should also be unlistening on all channels  (`UNLISTEN *`), releasing all advisory locks, and resetting parameters to whatever was in the startup message.

We should implement these as follow ups.

PR #1459, opened 2026-08-31
feat: count Postgres idle-in-transaction timeouts as a pool stat

### LINKED ISSUE #518: Track idle_in_transaction errors
### Description

Track idle-in-transaction errors and present them as a stat.

murex971:
## Summary
- Count server connections terminated by PostgreSQL's `idle_in_transaction_session_timeout` (`25P03`).
- Propagate the count from server stats to pool stats during connection check-in.
- Expose `total_idle_xact_timeouts` and `avg_idle_xact_timeouts` through `SHOW STATS` and OpenMetrics.
- closes #518


levkk:
This is something you can record on the server stats [1] instead which are automatically propagated to pool stats [2].

1: https://github.com/pgdogdev/pgdog/blob/9b15c4431d5343277246dab141be7c4da6a68ebd/pgdog/src/backend/server.rs#L638
2: https://github.com/pgdogdev/pgdog/blob/9b15c4431d5343277246dab141be7c4da6a68ebd/pgdog/src/backend/pool/pool_impl.rs#L254

pgdog/src/frontend/client/mod.rs
```diff
@@ -543,6 +543,11 @@ impl Client {
                         }
 
                     match event {
+                        // Client disconnected after idling in a transaction.
+                        BufferEvent::IdleInTransactionTimeout => {
```
levkk: Oh I see what you're doing now. So, our idle in transaction timeout is not the only source of this error. In fact, this is way more common to be triggered by Postgres, so that's why I mentioned you should handle this in the `async fn read` handler in `server.rs` instead: you will see `25P03` error code returned by the server, and add it to the server counter, which will automatically add it to the pool counter too.

murex971: Ah! I get it now.
will change to postgres `25P03` 
would also rename as client prefix won't make sense

levkk (approved):

PR #1460, opened 2026-08-31
refactor(resharding): migrate schema_sync to ctx state 

meskill:
Prepare the schema_sync task to be explorable by control plane and tasks system.

- add new task wrappers for schema_sync
- drop old progress and tracking and use tasks state
- schema flow and pgdump refactor to drop unused parts and make it more reusable


levkk (approved):
LGTM!

levkk (approved):
🚀

PR #1463, opened 2026-08-31
Pin connections when temporary tables are used

### LINKED ISSUE #1145: Pin connections when temporary tables are used
Disable connection pooling for the client when its creating temporary tables. Re-enable once the tables are dropped or the client disconnects.

This needs to use our parser I think, where we can detect a `CREATE TEMP TABLE` statement and a subsequent `DROP TABLE` statement.

sgrif:
This commit changes the parser to look for `CREATE TEMP TABLE`, locking the connection until all are dropped. Tables can be created with `ON COMMIT DROP` as an option, which causes them to be dropped whenever a transaction is committed. Tables are dropped on rollback regardless of options. They can also be explicitly dropped with `DROP TABLE`.

The behavior here is slightly more complex than that of advisory locks, since temporary tables interact with transactions as normal, so we need to keep track of which tables have been committed and which haven't.

This will, however, incorrectly lock the connection indefinitely if a savepoint is rolled back and doing so would undo the creation of a temp table. Since this same bug affects transaction scope advisory locks, and fixing this bug is non-trivial, I've opted not to handle it here.

Temp tables cannot go in a schema, so we do not need to worry about schema qualification here.

Deallocating temp tables was already handled elsewhere so this commit didn't need to handle it.

Close #1145

### DIFF
--- a/pgdog/src/frontend/client/query_engine/end_transaction.rs
+++ b/pgdog/src/frontend/client/query_engine/end_transaction.rs
@@ -77,6 +77,15 @@ impl QueryEngine {
             && !rollback
             && context.transaction().map(|t| t.write()).unwrap_or(false);
 
+        if rollback {
+            self.temp_tables.retain(|_, state| state.committed);
+        } else {
+            self.temp_tables.retain(|_, state| {
+                state.committed = true;
+                !state.drop_on_commit
+            });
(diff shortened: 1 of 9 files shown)


levkk (approved):
Very nice

Issue #1470, opened 2026-09-01
Observability: login/startup latency histogram, in-flight logins gauge, abandoned-login counter


ryderhwang:
## Summary

When the login path saturates (#1469), PgDog is **completely silent**: TCP accept succeeds, the client gives up mid-startup on its own `connect_timeout`, and nothing appears in logs, `pgdog.errors`, or any OpenMetrics series. During our worst production event (~130k client-side connect timeouts over ~30 min) every PgDog-side signal was green. We could only diagnose it from client-side APM and kernel TCP counters.

## Feature request

Login/startup observability on the OpenMetrics endpoint:

1. **`pgdog_login_duration` histogram** — accept → ReadyForQuery, ideally with a label or separate series splitting the phases (startup read, auth exchange, pool params fetch), so a saturated phase is identifiable.
2. **`pgdog_logins_in_flight` gauge** — connections accepted but not yet ReadyForQuery. This is *the* leading indicator of login-path saturation: it climbs minutes before clients start timing out.
3. **`pgdog_logins_abandoned` counter** — client disconnected (or socket error) before login completed. This is exactly the event that was invisible: today it's indistinguishable from an LB health check (which `handle_client` deliberately ignores on `ConnectionReset`). Even counting them under a single counter, without logging, would have turned a 30-minute mystery into a one-glance diagnosis.
4. Optionally `pgdog_logins_total` (rate of completed logins) — cheap and makes capacity planning against the per-instance login ceiling trivial.

## Why metrics and not logs

At storm rates (thousands/s), per-event logging is unusable and itself a hazard; counters/histograms are the right shape. Log lines for abandoned logins could stay behind `log_connections`.

Same production deployment as #1452; happy to contribute the implementation if the shape above (or some variant) is acceptable.

Issue #1469, opened 2026-09-01
Login path saturates at ~550-700 connects/s per instance regardless of cores (production report)


ryderhwang:
## Summary

In a large production deployment we've measured PgDog's client login path saturating at **~550–700 new client connections/s per instance**, while the instance uses only ~0.5 of its 4 cores (`workers = 4`). Past that rate, TCP accepts succeed but startup/auth lags; clients hit their own `connect_timeout` mid-startup and PgDog logs and reports **nothing** (no errors, no warnings), because from its perspective the connection simply went away before login completed. Adding instances fixes it linearly; adding cores to an instance does not — which suggests a serialization point in the login path rather than CPU exhaustion.

## Environment

- pgdog v0.1.52, `pooler_mode = "transaction"`, `passthrough_auth = "enabled_plain_allow_change"`, no client TLS
- 4 vCPU / `workers = 4` per instance, Kubernetes (arm64), listen backlog raised via sysctls (accept-queue overflow was measured and ruled out as the main cause — ~373 overflow drops vs ~130k client-side connect timeouts during the worst event)
- Client fleet: Django/psycopg2, ~180 app instances, **a fresh connection per request** (no client-side pooling), so a rolling deploy of the app fleet produces reconnect waves of ~7k connects/s fleet-wide

## What we observed

- At 8 instances (~600/s/pod capacity each), a fleet-wide reconnect wave (~7k/s) drove client-side connect p99 to a pinned 5s (their libpq `connect_timeout`), with tens of thousands of failures over ~30 min — while `pgdog.errors` stayed 0, CPU stayed ~12% of the pod, and logs were silent.
- Scaling to 32 and then 80 instances made the same waves invisible (~175/s/pod → connect p99 ~20ms flat).
- Separate load testing measured the per-instance wall at ~700 connects/s against a warm instance and ~550/s under sustained load, roughly independent of core count.

## Reading the code, candidate suspects

(From `frontend/listener.rs` / `frontend/client/mod.rs` / `backend/databases.rs` on main — happy to be corrected.)

1. With passthrough auth, every login calls `databases::add(user)`. The password-matches path looks cheap (read + `constant_time_eq`), but the password-differs / first-seen path takes the global config lock, **clones the entire config**, calls `set()` and `reload_from_existing()` — per login. During a credential rotation with `allow_change`, a reconnect wave would hit this path on every login. We don't believe this is our steady-state wall (our password is stable), but it looks pathological under rotation + storm.
[...]

PR #1471, opened 2026-09-01
Add login path observability: login_duration histogram, in-flight gauge, abandoned counter

ryderhwang:
Implements #1470 (context in #1469).

## What

Four new series on the OpenMetrics endpoint:

- `logins_in_flight` (gauge) — accepted but not yet ReadyForQuery; the leading indicator of login-path saturation
- `logins` (counter) — completed logins
- `logins_abandoned` (counter) — clients that started the Postgres handshake but went away (or hit `client_login_timeout`) before login completed
- `login_duration` (histogram, ms) — accept → ReadyForQuery latency of completed logins

## Design

- A `LoginTimer` RAII guard is created at accept, **engaged** on the first startup message, **success**-marked when `Client::spawn` gets a logged-in client, and **disarmed** when PgDog ends the login with an explicit response (auth failure, shutting down, pool down) or the connection is a cancel request. Dropping it engaged-but-unfinished counts one abandoned login — every exit path is covered without instrumenting each error branch.
- Bare TCP connect/close with **no startup bytes** (LB health checks, k8s TCP probes) is excluded from `logins_abandoned` — those never engage the guard.
- Histogram rendering goes through a new overridable `OpenMetric::render_measurements` (default keeps today's behavior for every existing metric), emitting standard `_bucket`/`_sum`/`_count` with cumulative buckets. Buckets are 1ms–30s; happy to change bounds.
- The three scalar metrics also flow through the OTLP exporter; the histogram is text-endpoint-only for now since the OTLP renderer currently only speaks gauge/sum.

## Verification

- 6 new unit tests (guard semantics incl. probe exclusion + histogram text rendering); `cargo nextest run --profile dev` shows an identical pass/fail set to current `main` on the same machine (all pre-existing/env failures, zero deltas).
- Live smoke test against Postgres 14 with `passthrough_auth = "enabled_plain_allow_change"`, transaction mode: a real psql login renders `logins 1` + one histogram observation; a bare TCP connect/close renders nothing; a client that sends StartupMessage, receives AuthenticationCleartextPassword, and closes renders `logins_abandoned 1`.
- `cargo fmt` + `cargo clippy` clean.

## Why

During our worst production login-path saturation event (~130k client-side connect timeouts over ~30 min, details in #1469), every PgDog-side signal was green — TCP accepted, clients gave up mid-startup, and nothing was logged or counted. With these metrics the event is a one-glance diagnosis: `logins_in_flight` climbing with `login_duration` p99 stretching toward the clients' connect timeout, then `logins_abandoned` ticking up.


ryderhwang:
Validated under load while benchmarking for #1469: with the harness driving **15,000 logins/s** through a single instance (`workers = 1`), `logins` counted **119,068** — exactly matching the generator's success count, zero drift — and the histogram/abandoned/in-flight series behaved as designed (a mid-login client disconnect increments `logins_abandoned`; a bare TCP connect/close increments nothing).

One scope note from that investigation: these metrics see only connections PgDog has *accepted*. Kernel-level pre-accept drops (accept-queue overflow, conntrack) remain invisible by nature — `listen_backlog` (#1452) plus node `TcpExtListenOverflows` monitoring cover that side.

ryderhwang:
Re the codecov report — the 3 uncovered patch lines are all in paths that have no coverage on `main` either, independent of this PR:

- `stats/otel_exporter.rs` (2 lines): `Logins::load()` and `all.extend(logins.iter())` inside the OTLP push loop. `run()` returns early when no `otel.endpoint` is configured, and no CI config sets one, so the whole loop body is at 0% on `main` (file is ~22% for that reason).
- `frontend/listener.rs` (1 line): `login_timer.engage()` in the `Startup::GssEnc` arm. Nothing in the integration suites sends a GSSENCRequest, so the arm's existing `SslReply::No` line is equally uncovered on `main`. The `engage()`/`success()`/`disarm()` semantics are unit-tested in `stats/logins.rs`, and the SSL + plain-startup `engage()` calls, the `Client::spawn` hand-off, and the Cancel `disarm()` are exercised by the integration suites.

Happy to add a raw-socket GSSENC test or factor the exporter's collection step out into a testable fn if you'd like those covered, but I didn't want to restructure existing code for it in this PR.

PR #1473, opened 2026-09-01
feat: allow source and destination clusters to overlap

levkk:
Allow source and destination clusters to overlap. Matching databases in the destination cluster are ignored and data is only routed to shards not specified in the source cluster. This little hack allows us to:

1. Sync omni tables to new shards added to an existing cluster
2. Copy rows from existing nodes to new nodes with new sharding map

### Example

```toml
[[databases]]
name = "source"
host = "10.0.0.0"

[[databases]]
name = "destination"
host = "10.0.0.0" # Same host as source!
shard = 0

[[databases]]
name = "destination"
host = "10.0.0.1"
shard = 1

[[sharded_tables]]
database = "destination"
column = "tenant_id"

[[sharded_tables.mapping]]
values = [1, 2, 3]
shard = 0

[[sharded_tables.mapping]]
values = [4]
shard = 1
```

Using `data-sync`, we can copy:

1) all omni tables
2) all rows matching `tenant_id = 4`

from source to destination, without touching the original node.

### DIFF
--- a/pgdog/src/backend/replication/logical/publisher/publisher_impl.rs
+++ b/pgdog/src/backend/replication/logical/publisher/publisher_impl.rs
@@ -214,7 +214,7 @@ impl Publisher {
             // Each subscriber owns a partition of destination shards for omni-table DML
             // (dest_shard % n_sources == source_shard), preventing cross-subscriber deadlocks.
             let mut stream =
-                StreamSubscriber::new(dest, tables, OmniOwnership::new(number, n_sources));
+                StreamSubscriber::new(source, dest, tables, OmniOwnership::new(number, n_sources));
 
             // Take ownership of the slot for replication.
             let mut slot = self
@@ -813,7 +813,8 @@ mod test {
         let cfg = config();
         let cluster = Cluster::new_test(&cfg);
         cluster.launch();
-        let mut stream = StreamSubscriber::new(&cluster, &[], OmniOwnership::test());
+        let mut stream =
+            StreamSubscriber::new(&Cluster::default(), &cluster, &[], OmniOwnership::test());
         stream.connect().await.unwrap();
 
         let result = stream.handle(begin_copy_data(1)).await;
@@ -833,7 +834,8 @@ mod test {
         let cfg = config();
         let cluster = Cluster::new_test(&cfg);
         cluster.launch();
-        let mut stream = StreamSubscriber::new(&cluster, &[], OmniOwnership::test());
+        let mut stream =
+            StreamSubscriber::new(&Cluster::default(), &cluster, &[], OmniOwnership::test());
         stream.connect().await.unwrap();
 
         let result = stream.handle(commit_copy_data(1)).await;
--- a/pgdog/src/backend/replication/logical/subscriber/copy.rs
+++ b/pgdog/src/backend/replication/logical/subscriber/copy.rs
@@ -3,7 +3,7 @@
 
 use futures::future::join_all;
 use pg_raw_parse::Node;
-use tracing::debug;
+use tracing::{debug, warn};
 
 use crate::frontend::client::query_engine::TwoPcPhase;
 use crate::frontend::client::query_engine::two_pc::{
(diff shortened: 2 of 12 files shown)


levkk:
Next on the TODO:

~1. Add an `--omnisharded-only` flag to `copy-data` (and admin DB command)~ Use `CREATE PUBLICATION ... FOR TABLES` to only sync omni tables instead.
2. Add a way to build a temporary lookup table and atomically swap it during cutover

pgdog/src/backend/replication/logical/subscriber/stream.rs
```diff
@@ -167,31 +174,38 @@ impl StreamSubscriber {
...
+                continue;
+            }
+
+            let primary = shard.primary_standalone(ConnectReason::Replication).await?;
+            conns.push((shard_number, primary));
+        }
+
+        if conns.is_empty() {
+            return Err(Error::SourceDestinationIdentical);
         }
```
meskill, line 203: this is very similar code how it's done in the `copy.rs`, we may consider to put this inside `OverlappingShardsCheck` and reuse

pgdog/src/backend/replication/logical/subscriber/stream.rs
```diff
@@ -167,31 +174,38 @@ impl StreamSubscriber {
...
-    // Connect to all the shards.
-    //
-    // The transaction-control prepare and the omni FULL-identity validation run
-    // synchronously on the raw `Server` connections (both are one-shot,
-    // request/response query flows). Only once they succeed are the connections
-    // moved into their per-shard pipelined tasks for the streaming apply path.
+    // Connect to all shards.
     pub(crate) async fn connect(&mut self) -> Result<(), Error> {
-        let mut conns: Vec<Server> = vec![];
+        let mut conns = vec![];
```
meskill, line 179: oh, when the len of conns do not correspond to the len of destination shards it'll be more error prone if for any reason we'll use the conns index instead of shard_number inside for shard identification. I'll put the exact places right now what I found with this problem, but I wonder if we should cover it somehow on the type level:
- use wrapper structure for conns
- create empty Conn instead that do nothing to have the same len
- maybe operate on Shard or similar instead of passing shard_number along the data

pgdog/src/backend/replication/logical/subscriber/stream.rs
meskill, line 268: here should be the `dest.shards().len()` I believe instead, since this relies on the behavior of shard routing that sees all shards

pgdog/src/backend/replication/logical/subscriber/copy.rs
```diff
@@ -73,20 +76,32 @@ impl CopySubscriber {
...
+        let overlap_check = OverlappingShardsCheck::new(&self.source);
+        let destination_shards = self.dest.shards();
+
+        if destination_shards.is_empty() {
+            return Err(Error::DestinationNoShards);
         }
 
-        self.connections = servers;
+        for (shard_number, shard) in destination_shards.iter().enumerate() {
+            if overlap_check.overlaps(shard)? {
```
meskill, line 88: I wonder is we should use the `Shard` type or similar everywhere instead of tuples (shard_number, smth). There is a number already inside the `Shard` and we could simplify it in places like this and maybe reuse instead of adding shard_number to other structs

pgdog/src/backend/replication/logical/subscriber/copy.rs
```diff
@@ -73,20 +76,32 @@ impl CopySubscriber {
...
+                warn!(
+                    "skipping data sync to {} because it is part of the source cluster",
+                    shard.primary_address()?,
+                );
+                continue;
```
meskill, line 93: some edge case to validate maybe: check if the shard number is different in source/destination cluster, just in case

pgdog/src/backend/replication/logical/publisher/publisher_impl.rs
meskill, line 447: ParallelSync looks like using all the shards for `destination_has_rows` check, so for any error we'll stop retrying since we'll find the entries on the existing shards

meskill:
Sounds like to this to work some invariants should hold, because if we change the number of shards the routing will expect the data be reorganized differently on the destination comparing to source and since we leave the data on source shards as is we should guarantee that the routing will work the same way as it was before.
I wonder if we can check this somehow and prevent issues? That relies on sharded_tables.mapping configuration, but I'm not sure we can use it to do the validation upfront if not only the configuration is exhaustive. Maybe we should abort in case we identified rerouting to another existing shard while doing the copy?

meskill:
This should have some integration test

PR #1474, opened 2026-09-02
fix: omit cross-shard aggregate helpers on single-shard queries

### LINKED ISSUE #1425: [Cross shard] temporary columns for `stddev()` visible in client response
Image

murex971:
## Summary
- Cross-shard `AVG`/`STDDEV`/`VARIANCE` rewrites add `__pgdog_*` helper columns before routing. When the query later hits a single shard, those helpers were still sent to Postgres and shown into the client result.
- Keep the original SELECT as `direct_stmt` and restore it after routing when the route is not cross-shard (simple Query and Parse/prepared statements). Cross-shard queries still use the helper rewrite.
- closes #1425

### DIFF
--- a/pgdog/src/frontend/router/parser/rewrite/statement/mod.rs
+++ b/pgdog/src/frontend/router/parser/rewrite/statement/mod.rs
@@ -166,8 +166,15 @@ impl<'a> StatementRewrite<'a> {
             return Err(err);
         }
 
-        if let NodeMut::SelectStmt(mut select) = stmt.stmt_mut() {
+        if matches!(stmt.stmt(), Node::SelectStmt(_)) {
+            let direct_stmt = pg_raw_parse::deparse(&*stmt)?.as_str().to_owned();
+            let NodeMut::SelectStmt(mut select) = stmt.stmt_mut() else {
+                unreachable!("statement was checked to be SELECT");
+            };
             self.rewrite_aggregates(&mut select, mem, &mut plan, self.db_schema)?;
+            if !plan.aggregates.is_noop() {
+                plan.direct_stmt = Some(direct_stmt);
+            }
             self.limit_offset(&select, &mut plan);
         }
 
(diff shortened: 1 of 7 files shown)


pgdog/src/frontend/router/parser/rewrite/statement/mod.rs
```diff
@@ -166,8 +166,15 @@ impl<'a> StatementRewrite<'a> {
...
         }
 
-        if let NodeMut::SelectStmt(mut select) = stmt.stmt_mut() {
+        if matches!(stmt.stmt(), Node::SelectStmt(_)) {
+            let direct_stmt = pg_raw_parse::deparse(&*stmt)?.as_str().to_owned();
+            let NodeMut::SelectStmt(mut select) = stmt.stmt_mut() else {
+                unreachable!("statement was checked to be SELECT");
+            };
             self.rewrite_aggregates(&mut select, mem, &mut plan, self.db_schema)?;
+            if !plan.aggregates.is_noop() {
```
levkk, line 174: I think it would cool to have a `rollback_aggregates` function (should be pretty easy to remove the helper columns we added) and only then call `deparse`. That way, queries without aggregates bypass the deparse step and remain fast!

pgdog/src/frontend/router/parser/rewrite/statement/plan.rs
```diff
@@ -61,8 +69,36 @@ impl RewriteResult {
     pub(crate) fn apply_after_parser(&self, request: &mut ClientRequest) -> Result<(), Error> {
         match self {
             Self::InPlace {
-                offset: Some(offset),
-            } => offset.apply_after_parser(request),
+                offset,
+                direct_stmt,
+            } => {
+                if request.is_executable()
```
levkk, line 90: I wouldn't mind moving this to `apply_after_parser` (maybe its own function), to keep this code clean. The `apply_after_parser` also has better context on how to rewrite requests correctly (since `apply` is adjacent too), so there is more opportunity for code re-use.

levkk:
I think just a couple of small changes needed. Longer term we would want to not run the aggregate rewriter at all if the query is direct to shard, but that requires a bigger refactor (parse before rewrite, not rewrite before parse like we do now).

murex971:
yeah, ideal architecture would be first decide query route then rewrite, but it would have been a bigger refactor

levkk:
:eyes:

murex971:
@levkk adressed comments

Issue #1475, opened 2026-09-02
Complete session state cleanup for DISCARD ALL


murex971:
PgDog intercepts `DISCARD ALL` and responds without forwarding it to PostgreSQL. Prepared statements are cleared, but other session state documented by PostgreSQL is still retained.
`DISCARD ALL` should also:
- remove all `LISTEN` subscriptions (`UNLISTEN *`);
- release all advisory locks held by the client;
- reset session parameters to the values supplied in the startup message.

This should make a PgDog session behave like a PostgreSQL session after `DISCARD ALL`.
Reference: https://www.postgresql.org/docs/current/sql-discard.html

PR #1478, opened 2026-09-02
fix: separate DISCARD variants and prepared statement cleanup

### LINKED ISSUE #1477: Handle PostgreSQL DISCARD variants independently
Description:

PgDog currently treats `DISCARD ALL`, `DISCARD PLANS`, `DISCARD TEMP`, and `DISCARD SEQUENCES` as the same command. As a result, the client prepared-statement cache is cleared for every variant.

Update the parser and query engine to preserve the requested DISCARD variant and apply the correct behavior:

DISCARD ALL: clear prepared statements and decrement global usage counters.
DISCARD PLANS: invalidate cached plans without removing prepared statements.
DISCARD TEMP: drop temporary tables only.
DISCARD SEQUENCES: clear sequence state only.

murex971:
- Preserve the specific PostgreSQL DISCARD target during parsing.
- Clear client prepared statements only for DISCARD ALL.
- closes #1477

### DIFF
--- a/pgdog/src/frontend/client/query_engine/discard.rs
+++ b/pgdog/src/frontend/client/query_engine/discard.rs
@@ -9,10 +9,13 @@ impl QueryEngine {
     pub(super) async fn discard(
         &mut self,
         context: &mut QueryEngineContext<'_>,
+        target: DiscardTarget,
         extended: bool,
     ) -> Result<(), Error> {
         let _extended = extended;
-        context.prepared_statements.close_all();
+        if target == DiscardTarget::All {
+            context.prepared_statements.close_all();
+        }
         let bytes_sent = context
             .stream
             .send_many(&[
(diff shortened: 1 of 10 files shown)


levkk (approved):
LGTM. A few follow-ups we should address:

`DISCARD ALL` is also equivalent to `UNLISTEN *`, which we currently don't handle. We should add an `unlisten_all` function and call it here as well. Current implementation for reference: https://github.com/pgdogdev/pgdog/blob/9991e8a48e2037dadd26af4439b5ddcab4c4fd39/pgdog/src/frontend/client/query_engine/pub_sub.rs#L37

`DISCARD TEMP` should be forwarded to the Postgres connection & we should track that we dropped all temporary tables. We recently added support for pinning connections if temp tables are created, so this would effectively unpin the client: https://github.com/pgdogdev/pgdog/pull/1463

murex971:
yes, I have opened #1475 for these changes

Issue #1096, opened 2026-06-22
[Subqueries] Handle pl/pgsql invocations in the `FROM` clause by executing them as a subquery


hanjianqiao:
**PgDog version**
ghcr.io/pgdogdev/pgdog:latest

**Description**
When inserting 200 data entries using the `generate_series` function, 200 data entries were generated for each shard, which was intended to distribute these 200 data entries evenly across the shards.

**Logs**

```bash
# using pgdog with two shards
CREATE TABLE orders (order_id bigint, customer_id bigint, amount numeric(10,2), PRIMARY KEY (order_id, customer_id));
INSERT INTO orders SELECT g, g, (g*1.5) FROM generate_series(1,200) g;
```
After previous query is done, via pgdog `select count(*) from orders;` returns 400, and each shard has 200 same records.

**Configuration**
pgdog.shard.toml
```toml
[general]
host = "0.0.0.0"
port = 6432
default_pool_size = 10

[[databases]]
name = "data_shard"
host = "data-shard0"
port = 5432
database_name = "testdb"
shard = 0

[[databases]]
name = "data_shard"
host = "data-shard1"
port = 5432
database_name = "testdb"
shard = 1

[[sharded_tables]]
database = "data_shard"
column = "customer_id"
```

levkk:
Yup, we need to evaluate and expand this function inside PgDog. Currently, we send it as-is to Postgres.

levkk:
Actually, correction, we should handle this as a subquery:

1. Send it to the right shard
2. Collect the result
3. Inject it into the top level query and send it to Postgres

PR #1482, opened 2026-09-02
refactor: consolidate multi-step planning, assembly and execution into QueryPlanner type

jkaczman:
@sgrif and I have been working on #1096 (subqueries), and as the first step to getting there, discussed how we can create an architecture for multi-step that...
- (1) reduces duplication, 
- (2) concentrates all relevant code into one area, 
- (3) clearly articulates query dependencies (as in, we cannot construct a query without receiving the responses of a previous query). An example of being we cannot construct the INSERT for Sharding Key Updates without the DELETE ... RETURNING * `DataRow` returning. Another example being, of course, subqueries.
- (4) doesn't do more than necessary; as in, we only need the `Parse`/`Bind` handled dynamic Steps, we don't need to pre-make requests for each before execution-time,
- (5) removing unnecessary code, e.g. the `InPlace `variant, so that we can just receive the `QueryPlanner` and handle it, and also, making `OffsetPlan` independent of this new struct, as it's not relevant (and can just be passed straight through)
- (6) a nice abstraction for (1) knowing when to forward messages to the Client, and (2) determining how to aggregate them (especially in dynamic situations)
- (7) running checks during planning instead of having a check in the parser and deferring more checks until later, which we can do because we have the `QueryEngineContext` to work with now.
- ... and more 

No tests were removed. Added a couple of tests for a few things I wanted to make sure worked. 

A lot of the diff is from moving pre-existing functions/tests around.
- `pgdog/src/frontend/router/parser/rewrite/statement/update.rs` => `pgdog/src/frontend/client/query_engine/multi_step/test/sharding_key_update.rs`
- `pgdog/src/frontend/router/parser/rewrite/statement/insert.rs` => `pgdog/src/frontend/client/query_engine/multi_step/test/split_insert.rs`

TODO: Cache insert splits / subqueries / etc (relevant benchmark: https://github.com/pgdogdev/pgdog/pull/1484)


pgdog/src/frontend/client/query_engine/multi_step/types.rs
```diff
@@ -0,0 +1,113 @@
...
+
+/// We need to preserve all responses instead of just the ones we plan to look up,
+/// for example, for purposes of constructing a Response based on `CommandComplete`s,
+/// which is why I opted for this structure over a `HashMap`
+impl ResponseHistory {
+    pub(crate) fn push(&mut self, responses: StepResponses) {
+        self.steps.push(responses);
+    }
+
+    pub(crate) fn get(&self, key: &str) -> Option<&StepResponses> {
```
levkk: `key` could be a type instead of static string. That would make it less error-prone.

pgdog/src/frontend/router/parser/rewrite/statement/plan.rs
```diff
@@ -34,40 +36,22 @@ pub(crate) struct RewritePlan {
     /// Each tuple contains (name, statement) for ProtocolMessage::Prepare.
     pub(crate) prepare_rewrites: Vec<PrepareExecute>,
 
-    /// Splitting of multi-tuple INSERT statements into
-    /// multiple queries.
-    pub(crate) insert_split: Vec<InsertSplit>,
```
levkk, line 39: We used to cache the rewrite previously. Now, we're performing it on each request, prepared statement or not. I suspect that would add some performance overhead to this step, worth benchmarking.

PR #1483, opened 2026-09-02
fix: run SCRAM key derivation off the async runtime

### LINKED ISSUE #1468: SCRAM auth blocks Tokio runtime
**PgDog version**
`main`

**Description**
SCRAM is CPU-bound and pretty expensive. It runs on the Tokio worker, blocking the runtime. During client connection storms, this creates backpressure and can cause connections to fail.

https://github.com/pgdogdev/pgdog/blob/f40d5f6a298efbd86753296e16545cd97d128965/pgdog/src/auth/scram/server.rs#L58-L73

**Potential fix**

Modify the `scram` crate and make that call async. Once async, push it to a background thread, i.e., `spawn_blocking`.

mvidaurre:
Fix #1468.

Move hash computation to background thread using `spawn_blocking`. Precompute the hash in advance instead of inside the `scram` crate trait.

### DIFF
--- a/pgdog/src/auth/scram/server.rs
+++ b/pgdog/src/auth/scram/server.rs
@@ -6,6 +6,7 @@ use crate::net::messages::*;
 
 use pgdog_config::users::PasswordKind;
 use scram::server::ClientFinal;
+use tokio::task::spawn_blocking;
 use tracing::error;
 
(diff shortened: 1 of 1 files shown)


levkk (approved):
Cool, thanks!

PR #1486, opened 2026-09-02
feat: per-database TLS overrides for server connections

rlittlefield:
Allows each `[[databases]]` entry to override the upstream TLS settings that were previously global-only: `tls_verify`, `tls_server_ca_certificate`, `tls_server_certificate`, and `tls_server_private_key`.

Motivation: managed Postgres providers with per-instance CAs (e.g. GCP Cloud SQL) issue a different server CA and client certificate for the primary and each replica. A single global client cert cannot authenticate to both, so one PgDog deployment serving primary and replica pools needs per-server TLS settings.

Behavior:
- Each override falls back to its `[general]` counterpart, except the client certificate and key, which act as a keypair: setting either per-database replaces the global pair. Config validation rejects a cert without its key (and vice versa) at load time instead of failing on the first connection.
- The upstream connector cache is now a map keyed by the resolved TLS settings instead of a single slot, so servers with different certs no longer thrash a single cached connector.
- TLS overrides are part of `Address`, so pool identity changes (and connections are re-established) when they change on reload.
- Reload validates every database's resolved TLS settings, rebuilds each in-use connector from disk (picking up certificate rotations at unchanged paths, e.g. re-mounted Kubernetes secrets), and drops connectors the config no longer references. Nothing is swapped in unless the whole config validates.
- `.schema/pgdog.schema.json` regenerated; `example.pgdog.toml` documents the new settings.

Testing: unit tests cover config parsing, validation, override resolution (including the keypair rule), connector cache reuse and eviction, and reload rebuild-from-disk behavior. Integration tests (integration/tls/backend, CI only, same gating as the existing backend-TLS setup in that suite) run against Postgres with clientcert=verify-ca: a per-database client cert authenticates while the certless [general] fallback is rejected, and an in-place certificate rotation at an unchanged path is picked up by RELOAD. `cargo fmt` and `cargo clippy` are clean.


levkk:
Nice, LGTM

rlittlefield:
I'm going to see if we can get the reload to work too. Might be bigger that way so it could go in a different PR if that isn't required.

levkk:
Pretty sure the connector is dynamic and re-built on reach config reload

rlittlefield:
I added a flattened config there to reuse based on your suggestion!

For the reload, build_acceptor is called on each reload, but since the config is loaded and cached by path, it would only reload the actual cert file contents if the path changes. So as long as the path is identical, I think it would fail to get the new contents.

levkk:
That's definitely an oversight on our part. Should be a relatively easy fix I suppose, adds a bit of overhead to the reload, but that's fine.

pgdog-config/src/server_tls.rs
```diff
@@ -0,0 +1,29 @@
...
+/// same name.
+#[derive(
+    Serialize, Deserialize, Debug, Clone, PartialEq, Default, JsonSchema, PartialOrd, Eq, Ord, Hash,
+)]
+#[serde(rename_all = "snake_case", deny_unknown_fields)]
+pub struct ServerTls {
+    /// Overrides the `tls_verify` setting for connections to this database.
+    /// Useful when servers in the same cluster require different TLS settings,
+    /// e.g., managed databases where each instance has its own CA.
+    pub tls_verify: Option<TlsVerifyMode>,
```
rlittlefield, line 18: The optional here makes it harder to flatten onto then`general`, because general doesn't have this as optional, but it should be optional here.

levkk: Ah well, shucks. My dream of flattened configs will have to live to fight another day. I wonder why it's not optional in general...probably because we default to false with serde to make it easier to work with, but we can just as well make it optional and do `general.tls_verify.unwrap_or_default()` in our code dealing with it.

rlittlefield:
Added some integration tests!

integration/tls/run.sh
```diff
@@ -37,4 +60,9 @@ echo "PgDog is ready"
 
 bash ${SCRIPT_DIR}/dev.sh
 
+# Backend mTLS tests need the Postgres client-cert setup above (CI only).
+if [ "${GITHUB_ACTIONS:-}" = "true" ]; then
```
levkk, line 64: CI only test coverage is test coverage that doesn't get run locally :) We will, sooner rather than later, need the test suite to run locally so our team can test this stuff for regressions before pushing to the borg kube which has 0 9's of uptime :)

levkk: No need to address this in this PR, just FYI for next time.

pgdog/src/net/tls.rs
```diff
@@ -217,21 +274,35 @@ pub(crate) fn reload() -> Result<(), Error> {
...
+    let general_settings = UpstreamTlsSettings::resolve(general, &no_overrides);
+    let database_settings = config
+        .config
+        .databases
+        .iter()
+        .map(|database| UpstreamTlsSettings::resolve(general, &database.tls));
+
+    for settings in std::iter::once(general_settings).chain(database_settings) {
+        if let Entry::Vacant(entry) = connectors.entry(settings.cache_key()) {
+            let client_config = build_connector(entry.key())?;
```
levkk, line 293: Note for future self: I wonder how expensive this is and should we just read the contents from disk and compare byte-for-byte before rebuilding.

levkk (approved):
Let's rock and roll.

PR #1487, opened 2026-09-03
fix: check TTL expiration for Prepare

### LINKED ISSUE #1382: PREPARE doesn't work with prepared statements TTL
**PgDog version**
`main`

**Description**
#1368 added partial support for `PREPARE` and `EXECUTE`. It currently doesn't work with TTL, since we don't check the statement TTL when executing queries. #1351 

**Configuration**

```toml
[general]
prepared_statements_ttl = 5_000
```

jkaczman:
Enforces TTL expiration on simple `Prepare` / `Execute`, where it'll now re-`Prepare` internally if expired.

The unit test I added checks that it does a `Close` -> `Prepare` on an expired statement, and the integration test I added checks that everything works as it should.

I should note that `check_prepared` does have pre-existing expiration check logic (for `Parse`), and I considered modifying it to work for both cases, however it would over-complicate the function by accounting for two different cases (e.g. different returns based on `Prepare` vs `Parse`), so I think the simple check I added instead makes a little more sense.

Fixes https://github.com/pgdogdev/pgdog/issues/1382.


integration/pgdog.toml
```diff
@@ -11,7 +11,7 @@ read_write_strategy = "aggressive"
 openmetrics_port = 9090
 openmetrics_namespace = "pgdog_"
 prepared_statements_limit = 500
-prepared_statements = "extended"
+prepared_statements = "full"
```
levkk, line 14: Nice! That one will break like half the test suite, so don't worry about the red that's about to flow your way

integration/rust/tests/integration/simple_prepared.rs
```diff
@@ -0,0 +1,33 @@
...
+
+    sqlx::raw_sql("PREPARE __pgdog_test (int) AS SELECT $1")
+        .execute(&mut *transaction)
+        .await
+        .unwrap();
+
+    integration_tests_rust::utils::assert_setting_str("prepared_statements_ttl", "5000").await;
+    integration_tests_rust::utils::assert_setting_str("prepared_statements_ttl_jitter", "1000")
+        .await;
+    sleep(Duration::from_secs(8)).await;
```
levkk: I would probably set the TTL to like 100ms so we don't make the test suite too too slow, but yeah, lgtm otherwise. Also, curious how it interacts with the transaction, and if that matters at all (I don't think it does, just a thought that popped into my head).

levkk (approved):
Awesome!

integration/pgdog.toml
levkk: I was wrong! That's so cool. We used to set this one to full and it would just blow up, but apparently not anymore. 🚀

integration/rust/tests/integration/simple_prepared.rs
jkaczman: Will change to 100ms!

I think the transaction does matter (-ish). The expiration relies on `local_cache` in `PreparedStatements` which "has the names currently prepared on the server connection." If I don't use the transaction, it doesn't reach the expiration check I have here; I noticed that when I was testing early on. I think the `EXECUTE` without using a transaction may check out a different server connection (with a `local_cache` that doesn't have the statement). I opted for the transaction here for consistency in ensuring it guarantees it's testing what we want it to.

levkk: Sounds good!

PR #1492, opened 2026-09-03
Release LISTEN subscriptions and advisory locks on DISCARD ALL

murex971:
- Stop all LISTEN subscriptions on DISCARD ALL.
- Release advisory locks on PostgreSQL and clear PgDog’s lock tracking.
- Unpin the backend after advisory locks are released.


levkk (approved):
Awesome!!

PR #1493, opened 2026-09-03
feat(config): expand environment variables in pgdog.toml and users.toml

### LINKED ISSUE #1479: Expand environment variables in configuration files
I ended up implementing a feature to expand environment variables in pgdog.toml/users.toml to make it easier to work with how we are introducing sealed secrets into our Kubernetes tooling that deploys instances of PgDog:  https://github.com/pgdogdev/pgdog/compare/main...censys-oss:pgdog:env-toml?expand=1. We could have gone the route of using an initContainer to glue everything together but this ended up being pretty easy and much easier to work with.

I wanted to ask here if this was something that ya'll would be interested in contributing back to PgDog? I'd be very happy to contribute this, but wanted to make an issue before firing off a PR per the contributing guidelines.

Thanks for the consideration and for this awesome project!

ChrisRx:
Configuration files can now reference the process environment and the contents of files on disk, so secrets and per-environment values no longer have to be baked into `pgdog.toml` and `users.toml`:

```toml
[admin]
password = "${file./run/secrets/admin_password}"

[general]
shutdown_timeout = ${env.PGDOG_SHUTDOWN_TIMEOUT:-60000}
```

Two reference kinds, both requiring the braced, prefixed form:

- `${env.VAR}` — the environment variable `VAR` from the process environment.
- `${file.PATH}` — the contents of the file at `PATH`, with trailing newlines trimmed. A relative path resolves against the current working directory.

Both accept a fallback (`${env.VAR:-value}`, `${file.PATH:-value}`), and `$${env.VAR}` is a literal `${env.VAR}`.

Everything else is literal text: a bare `$VAR`, an unprefixed `${VAR}`, a malformed or unterminated `${`, and an `env.` reference to a variable that isn't set. That leniency is what keeps existing configs working — `users.toml` is the file most likely to contain a stray `$`, so passwords like `sup$rsecret` or `p$$w0rd` pass through untouched instead of turning into a startup failure or a silently truncated credential.

`file.` references are strict where `env.` references are lenient. A `file.` reference names a path the operator wrote down on purpose, so a file that can't be read is an error — `Error::FileReference`, carrying the path and the underlying `io::Error` — unless a fallback is given. A silently empty password is the failure mode worth avoiding there. Trailing newlines are trimmed because secret managers and `echo` alike conventionally leave one behind.

Expansion runs on the document source before it is parsed, so a reference is interpolated as TOML rather than as a string. `${env.PASSWORD}` in value position still needs its surrounding quotes, and a value containing `"` or a newline will change how the rest of the document parses. This is also what allows bare `shutdown_timeout = ${env.VAR}` to work, and it is documented on `expand`.

Implementation notes:

- New `pgdog-config::expand` module. `expand()` returns `Cow::Borrowed` when the source contains no `${`, so the common case costs no allocation. A `Reference` enum carries the parsed target, so recognition stays where the `}` is found and substitution just matches on what was recognised.
- An unrecognised `${` emits through the `${` and rescans immediately after it, so a stray `${` in one value can't swallow a real reference later in the document.
- `is_name` is the shell rule (letters, digits, underscores, not leading with a digit). `is_path` is deliberately loose — non-empty, no whitespace, no `$` or `{`. It exists to decide reference-or-literal, not to validate a path; anything stricter would reject paths the filesystem accepts, and `read_to_string` reports the real answer anyway.
- Only trailing `\r` and `\n` are trimmed, not interior newlines or trailing spaces, both of which can be part of a secret.
- `FromToml::from_toml` replaces bare `toml::from_str` at the three sites that parse config text read from disk: both branches of `ConfigAndUsers::load` and `bootstrap_logger`. Every other `toml::from_str` in the tree parses a test literal, where expansion is unwanted, and is untouched.
- The trait carries a blanket impl over `DeserializeOwned`, so no per-type boilerplate is needed. `from_toml`, not `from_str`, to avoid colliding with the crate's many `std::str::FromStr` impls.
- `Error::config` now receives the expanded text, so the line numbers it reports stay correct when a reference's value contains a newline.
- `ConfigAndUsers` keeps `config_text`/`users_text` as the raw, unexpanded source. Resolved secrets must not be written back to disk when the config is reloaded or backed up.
- Tests write real files with `tempfile` rather than mocking the read, which keeps the newline-trimming and missing-file cases honest.

The branch's first commit introduced the unprefixed `${VAR}` form and the third moved it under `env.`; that syntax was never released, so nothing in the wild depends on it. The namespace is what makes a second reference kind possible without guessing at the target from its shape, and it widens the set of values that pass through untouched: a `${...}` that isn't one of the two known prefixes is left alone rather than being treated as a lookup that happened to miss.

Fixes #1479


pgdog-config/src/expand.rs
```diff
@@ -0,0 +1,209 @@
...
+    let mut rest = source;
+
+    while let Some(start) = rest.find(OPEN) {
+        let body = &rest[start + OPEN.len()..];
+
+        // A reference is `${`, a valid name, an optional `:-fallback`, and `}`.
+        // Anything else is literal text: emit through the `${` and rescan right
+        // after it, so a stray `${` in one value can't swallow a real reference
+        // later in the document.
+        let reference = body.find('}').and_then(|end| {
```
levkk, line 41: I think all characters are allowed in a Postgres password, e.g., $, { and }, so this is a valid password which will be expanded to an empty string:

```
${hello}
```

Curious if you have any thoughts. Maybe we should only expand settings that are entirely covered by an env var, e.g.:

```toml
password = "${PASSWORD}" # setting value starts with `${` and ends with `}`
```

That would require us to perform shellexpand on each value after deserialization (or write a custom serializer).

Just thinking out loud, let me know what you think.

ChrisRx: In the case with `${hello}` it would need to be set as the password value and also be set in the environment, so unless it has an environment variable for `hello=` it will keep it as the original string, ultimately leaving as `${hello}`.

I was definitely concerned with using `shellexpand`, I could imagine a situation where generated passwords or especially some longer tokens could easily contain something where it would match shorter commonly set environment variables, like `CC`, where a randomly generated value like `....$CC.....` would get expanded to the value of `CC` with something like `gcc`. But I feel a lot better with the new non-shellexpand approach being that it requires:

1. The environment variable must still be set in the process environment
2. It must be a sequence of `${` followed by a `}`
3. The variable name itself can only contain contain characters `[a-zA-Z0-9_]` and cannot start with `_` or a digit (I based it off of POSIX standard, but including lowercase letters)

I don't know how to go about calculating a probability on it myself, but it seems like it would be practically impossible given the confluence of things that would need to happen coupled with password generators usually don't create passwords that include `{` or `}` and tokens  like JWTs are usually base64 encoded which would exclude that as well.

I really appreciate the discussion with this btw, I think this kind of thing IME is not something you can think too much about for sure!

ChrisRx: I had another thought that might be a better developer experience and would be even more impossible to match on sequences unintentionally. I have used this other project previously that had a similar capability, but it has a much more specific opening sequence since it can handle both environment variables and reading from files:  https://www.apollographql.com/docs/graphos/routing/configuration/yaml#variable-expansion. It uses `${}` but needs either `env.` or `file.` to specify the environment variable name or file name, respectively.

With an opener of `${env.` it means that only sequences matching roughly `${env.([a-zA-Z0-9_])+}` would even perform an environment variable lookup. I'm reasonably confident the previous approach would be practically impossible, but this would be without any doubt impossible to do unintentionally. And IMO is a better developer/user experience since it is very obvious reading the configuration file what is happening since it says "env" with each variable name.

Thoughts on this approach? It is trivial to make this change since it is just adjusting the opener.

PR #1496, opened 2026-09-03
test: run backend mTLS suite locally via docker compose

rlittlefield:
Follow-up to #1486: the backend mTLS tests were gated on GitHub CI because they reconfigured the host's Postgres cluster (ALTER SYSTEM ssl, pg_hba edits, pg_ctlcluster restart, sudo). This makes them run anywhere by following the pattern the other docker-based suites (load_balancer, prefer_primary, copy_data, resharding, vault) already use: the suite brings its own Postgres via docker compose.

- integration/tls/backend/docker-compose.yml: postgres:17 on port 15436, project-scoped as tls_backend.
- integration/tls/backend/docker/setup.sh (modeled on load_balancer/docker/primary.sh): installs the suite's server cert and CA, prepends "hostssl pgdog_mtls all all scram-sha-256 clientcert=verify-ca" to pg_hba.conf, restarts, creates the pgdog role and pgdog_mtls database.
- integration/tls/backend/run.sh: compose up, same five assertions as before (per-database override accepted, certless [general] fallback rejected, untrusted cert rejected, in-place rotation picked up by RELOAD, [general] still certless after reload), compose down in an EXIT trap. Exports PGDOG_STOP_TRAP=1 so run_pgdog does not replace the cleanup trap when the suite is run standalone.
- PgDog listens on 16432 for this suite: 6432 is commonly taken on dev machines by a running pooler or tunnel.
- integration/tls/run.sh: the GITHUB_ACTIONS gate around the backend tests and the host-Postgres mTLS setup are removed; the pre-existing CI block for the client-side tests is back to its original form.

Verified locally on macOS against a colima docker daemon: 5 passed, 0 failed, clean teardown. The trap fix and the 16432 port choice both came out of local runs that CI could not have caught.


levkk (approved):
Thank you!

PR #1499, opened 2026-09-04
refactor(resharding): extract tables sync to separate function

meskill:
I started with extracting the logic related to table loading from orchestrator/publisher and ended up with OmniOwnership refactoring and new edge cases testing.

This PR moves the sync_table logic to separate module with omni-sharded tables distribution between streams embedded. This allows for:
- distribute omni for both copy_data and replication. That allows to drop OmniOwnership since it's basically the same approach from another perspective - let the single stream to handle updates for omni tables to avoid multiple parallel updates to the same rows
- add tests and handle the cases when the table changes from sharded to omni and backward. 
- even with all of this etc. there are some cases when we can deadlock (the duplicates on multiple shards that we don't expect to duplicate etc.) actually. I don't want to lose the parallelization for this cases, but also don't even want to leave things to chances, that's why I added the lock_timeout enforcement during the resharding


pgdog/src/backend/replication/logical/subscriber/tests.rs
```diff
@@ -2211,203 +2150,3 @@ async fn full_identity_delete_matches_null_column() {
...
-                columns: vec![text_column(id), text_column(name), text_column(value)],
-            },
-        }
-        .to_bytes(),
-    )
-}
-
-/// Two subscribers race on the same omni-table rows, reproducing the cross-destination deadlock.
-#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
-async fn cross_subscriber_omni_deadlock_two_databases() {
```
meskill, line 2305: this test constructs very specific environment and flow, that is not exactly how the whole flow runs during the replication.

This was helpful for debugging initially, but now it's a burden. I'm checking the full flow with more integration test `integration/resharding/repro_deadlock.sh`

levkk:
Any chance you could add #1473 into this as well or would that be out of scope?

meskill:
> Any chance you could add #1473 into this as well or would that be out of scope?

Looks like a different scope yes. But this should resolve the comment related to OmniOwnership, since there is no more

levkk (approved):
In Kiryl I trust

PR #1501, opened 2026-09-05
feat: handle DISCARD TEMP and release pinned connections

### LINKED ISSUE #1491: Handle DISCARD TEMP correctly
PgDog currently intercepts DISCARD TEMP without dropping temporary tables.

It should:

- Send DISCARD TEMP to the PostgreSQL connection.
- Clear PgDog’s temporary-table tracking after PostgreSQL succeeds.
- Unpin the client when no temporary tables remain.

murex971:
- Forward DISCARD TEMP to PostgreSQL and safely unpin the client.
- Correctly mark autocommit temporary tables as committed.
- closes #1491


pgdog/src/frontend/client/query_engine/mod.rs
```diff
@@ -103,6 +106,7 @@ impl QueryEngine {
             advisory_locks: AdvisoryLocks::default(),
             manual_lock: false,
             temp_tables: Default::default(),
+            discarded_temp_tables: None,
```
levkk: I think we should probably abstract that inside `temp_tables` which should now be its own struct. Just like we do with `Parameters`.

pgdog/src/frontend/client/query_engine/discard.rs
```diff
@@ -3,27 +3,28 @@ use crate::net::{CommandComplete, Protocol, ReadyForQuery};
...
-                self.backend
-                    .execute("SELECT pg_advisory_unlock_all()")
-                    .await?;
+
+        if target == DiscardTarget::Temp && self.backend.connected() {
+            self.execute(context, None).await?;
+
+            if !context.in_error() {
+                self.discard_temp_tables(context.in_transaction());
+                self.cleanup_backend(context)?;
```
levkk: I don't fully understand why you're calling `cleanup_backend` here.

pgdog/src/frontend/client/query_engine/discard.rs
```diff
@@ -3,27 +3,28 @@ use crate::net::{CommandComplete, Protocol, ReadyForQuery};
...
         extended: bool,
     ) -> Result<(), Error> {
         let _extended = extended;
-        if target == DiscardTarget::All {
-            if self.backend.connected() {
-                self.backend
-                    .execute("SELECT pg_advisory_unlock_all()")
-                    .await?;
+
+        if target == DiscardTarget::Temp && self.backend.connected() {
```
levkk: nit: a match on `target` might make this easier to read.

pgdog/src/frontend/client/query_engine/end_transaction.rs
```diff
@@ -77,14 +78,7 @@ impl QueryEngine {
...
 
-        if rollback {
-            self.temp_tables.retain(|_, state| state.committed);
-        } else {
-            self.temp_tables.retain(|_, state| {
-                state.committed = true;
-                !state.drop_on_commit
-            });
-        }
+        self.finish_temp_table_transaction(rollback);
```
levkk: This can be abstracted on `temp_tables` if it used its own struct with transaction state tracking I suppose. As we add more primitives here, we should encapsulate their handling inside their own entities. `QueryEngineContext` is becoming quite overloaded otherwise.

pgdog/src/frontend/client/query_engine/discard.rs
murex971: `execute()` does cleanup before PgDog’s temporary-table is updated, so backend is still considered pinned. After `DISCARD TEMP` succeeds, we clear the tracking and recalculate the lock. Calling `cleanup_backend()` again releases the now-unpinned connection immediately.

levkk (approved):
Nice!

PR #1502, opened 2026-09-05
feat: add admin FORCE_RELOAD command

### LINKED ISSUE #1472: [Admin] Add FORCE_RELOAD command
This reloads the config _and_ terminates any in-flight transactions. This guarantees that all transactions 
following the completion of this command are sent to the new connection pools.

jkaczman:
Adds an admin `FORCE_RELOAD` command, which acts as a normal `RELOAD` plus guarantees all in-flight transactions are terminated prior to the command's conclusion (and subsequent transactions are sent to the new `Pool`s).

TODO:
- Benchmark

PR for documentation of this command: https://github.com/pgdogdev/docs/pull/111
Fixes #1472.

### DIFF
--- a/pgdog/src/backend/pool/cluster.rs
+++ b/pgdog/src/backend/pool/cluster.rs
@@ -8,6 +8,7 @@ use pgdog_config::{
 };
 use std::{sync::Arc, time::Duration};
 
+use crate::backend;
 use crate::backend::schema::SchemaCache;
 use crate::backend::server::ServerRequest;
 use crate::frontend::router::sharding::ShardedTable;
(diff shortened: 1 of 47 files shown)


pgdog/src/backend/pool/cluster.rs
```diff
@@ -719,6 +720,42 @@ impl Cluster {
...
+                    //
+                    // TODO: If we error below on standalone or execute,
+                    // it'll leave the `Pool` in a pause state incorrectly.
+                    pool.pause(true);
+
+                    // Connect outside of `Pool` idle connections to prevent waiting for an available connection.
+                    // This also bypasses [`Pool.pause`]
+                    //
+                    // TODO: Say that this fails for some reason; it already internally re-tries multiple times.
+                    //       should we Error because not all transactions are terminated? Ignore it?
```
jkaczman: I wanted to get some input on this before implementing a solution for handling errors. I think that if we're guaranteeing all transactions are terminated, then we probably want to exit early with an error to the client and not reload. 

However, what if there were some circumstance (e.g. no free connection slots on the server), which prevents us from terminating forever? To handle cases like that, should we have two different variants of this command (or perhaps a parameter) to "override" a potential failure in terminating a transaction?

levkk:
I'm curious what you think about an alternative approach.

Instead of making Postgres cancel queries, do it from our side. You can add a `select!` here, for example: https://github.com/pgdogdev/pgdog/blob/75942792af5db4db258067ed030ae7cc8a055a5a/pgdog/src/frontend/client/query_engine/query.rs#L69-L73

The other branch can be a `CancellationToken` you get from `Cluster` (should be in-scope already) that `FORCE_RELOAD` cancels; that way, you can make this command pool-specific if you wanted to. This will stop waiting on a in-flight query. Once cancelled, execute a function on the server connection to force-close it.

For clients that are idle-in-transaction, you could do the same here: https://github.com/pgdogdev/pgdog/blob/75942792af5db4db258067ed030ae7cc8a055a5a/pgdog/src/frontend/client/mod.rs#L658 except you just send the client an error and exit, no need to do anything to the server connections I don't think; the `Guard` will clean up the abandoned transaction.

Do all of this after setting the pool to offline, so no new queries are accepted.

I'm thinking this should be faster than opening a new conn to Postgres for each connection pool, and also guarantees that this command is processed by us. `pg_terminate_backend` may not terminate the connection immediately (although it usually does).

jkaczman:
I definitely agree; I wasn't a fan of it either. Thanks!

I pushed a commit with some revisions representing a draft of your suggestions. Also, I added another test for cancelling during a query using `pg_sleep` (in addition to an idle transaction).

> that way, you can make this command pool-specific if you wanted to

I added internal support for per-pool cancellation, but it definitely adds some extra complexity (the Box::pin, futures::select_all, a Binding method, ..). Is this something we want to support in the command now, or is that structure we want for potentially doing something in the future?

pgdog/src/frontend/client/query_engine/mod.rs
```diff
@@ -125,6 +126,11 @@ impl QueryEngine {
         self.stats.state
     }
 
+    /// Fetch all `CancellationToken`s for the backend.
+    pub(crate) fn cancellation_tokens(&mut self) -> Vec<CancellationToken> {
```
levkk: I think you should make these `Cluster`-scoped. Each client will be connected to at most one cluster, so you won't need a `Vec` anymore (1 less allocation in the hot path).

Also, you're cancelling _all_ pools every time, so having it live on the pools isn't necessary, just the cluster will do.

levkk (approved):
Good stuff!!

PR #1504, opened 2026-09-06
feat: multi-shard LIMIT+OFFSET in-flight re-write in PREPARE+EXECUTE

### LINKED ISSUE #1383: PREPARE and OFFSET doesn't work for cross-shard queries
**PgDog version**
`main`

**Description**
#1368 introduced support for `PREPARE` and `EXECUTE`. It currently doesn't work for cross-shard queries that use `OFFSET` because it requires a in-flight `LIMIT` clause rewrite.

**Example**

```sql
PREPARE stmt1 AS SELECT * FROM sharded LIMIT 5 OFFSET $1;
EXECUTE stmt1(5);

PREPARE stmt1 AS SELECT * FROM sharded LIMIT 5 OFFSET 50;
EXECUTE stmt1;
-- rewrite: SELECT * FROM sharded LIMIT 55 OFFSET 0
-- and update `Route` with limit/offset settings.
```

jkaczman:
Implements functionality that handles multi-shard queries that use `LIMIT` + `OFFSET` in `PREPARE` + `EXECUTE` to re-write the `LIMIT` and `OFFSET` values (`LIMIT = LIMIT + OFFSET`, `OFFSET = 0`) and update the `Route` to expect them later to handle the appropriate multi-shard response for the client.

Specifically,
- When we first receive a `PreparedStatement` with a `SELECT` query that has a `LIMIT` + `OFFSET`, it re-writes any `A_Const` nodes (constant numbers) to be `ParamRef` nodes. This allows us the ability to always be able to modify the limit/offset parameters later on, in a helper method stemming off of `OffsetPlan::apply_after_parser`. This later-on modification (during execution) only happens if it's multi-shard.
- When we get the matching `ExecuteStatement`, if the Client previously used `A_Const` nodes, which means we had to store them in cache, we append the relevant cached numbers to the parameters list for the `ExecuteStatement` (as the Client does not know we internally re-wrote the `PreparedStatement` in our cache).  If they previously used `ParamRef` nodes, the (cached, cloned) `OffsetPlan` is updated to the newly passed in values to complete it, so that we can use it in the `OffsetPlan::apply_after_parser` helper method.
- Added `Option<OffsetPlan>` to prepared statement. Did not modify `CacheKey::Simple`; I modified it to use the original query (before rewrite) to store in the cache, and store the rewritten query (if any) in the cache. This allows us to later (1) which parameter is offset / which is limit, as well as (2) if the client pre-specified `A_Const` nodes, we must save them in our cache, as they will not be given to us in the `ExecuteStatement`.

Added some integration tests to test various cases: both limit/offset being `ParamRef`s, one being a `ParamRef`, both being `A_Const`, switching up the ordering (LIMIT $2 OFFSET $1), adding in an unrelated `ParamRef` (WHERE id < $2)... used ORDER BY for determinism in ensuring that the response we got is accurate. 

Integration test and unit test also checks to make sure that two Queries that are the same due to a re-write won't conflict in `GlobalCache`.

Fixes https://github.com/pgdogdev/pgdog/issues/1383.

### DIFF
--- a/pgdog/src/frontend/router/parser/rewrite/statement/offset.rs
+++ b/pgdog/src/frontend/router/parser/rewrite/statement/offset.rs
@@ -1,4 +1,6 @@
-use pg_raw_parse::{ConstValue, Node, Owned, StmtList, nodes};
+use std::ops::Deref;
+
+use pg_raw_parse::{ConstValue, Node, Owned, StmtList, deparse, nodes};
 
 use crate::frontend::ClientRequest;
 use crate::frontend::router::parser::Limit;
@@ -7,11 +9,12 @@ use crate::net::messages::bind::{Format, Parameter};
 
 use super::*;
 
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub(crate) struct OffsetPlan {
     pub(crate) limit: Limit,
     pub(crate) limit_param: usize,
     pub(crate) offset_param: usize,
+    pub(crate) prepare_execute: bool,
 }
 
 impl OffsetPlan {
@@ -25,6 +28,9 @@ impl OffsetPlan {
             return Ok(());
         }
 
+        if self.prepare_execute {
+            return self.handle_prepare_execute(request);
+        }
         // Resolve actual values: use literal if known, otherwise read from Bind.
         let mut limit_val = self.limit.limit;
         let mut offset_val = self.limit.offset;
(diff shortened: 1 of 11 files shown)


pgdog/src/frontend/router/parser/rewrite/statement/offset.rs
```diff
@@ -106,8 +112,80 @@ impl OffsetPlan {
 
         Ok(())
     }
+
+    /// `apply_after_parser` helper method for handling Prepare + Execute cases, where
+    /// we need to re-write limit / offset for multi-shard queries upon execution.
+    fn handle_prepare_execute(&self, request: &mut ClientRequest) -> Result<(), Error> {
```
levkk, line 119: What about this:

1. If you see `PREPARE ... SELECT ... LIMIT`, rewrite it immediately to `... LIMIT $x OFFSET $y` and store that in the cache
2. If you see `EXECUTE __stmt_above`, rewrite it to include the $x and $y parameters & set them accordingly

This way, you don't need to have the `OffsetPlan` as a separate entity and each execution of a statement that matches the existing cache key corresponds to an immutable query rewrite plan you can just execute. Wdyt?

pgdog/src/frontend/router/parser/rewrite/statement/simple_prepared.rs
```diff
@@ -122,6 +156,205 @@ fn rewrite_single_prepared<'a>(
...
+    if matches!(stmt_query.limit_count(), pg_raw_parse::Node::None)
+        || matches!(stmt_query.limit_offset(), pg_raw_parse::Node::None)
+    {
+        return None;
+    }
+
+    // Count the `ParamRef` nodes in the query, so that we know what number to start at
+    // if we need to add some more.
+    let mut param_refs_count: usize = 0;
+    pg_raw_parse::walk::walk(stmt_query.into(), |node| {
```
levkk, line 230: We do this already here: https://github.com/pgdogdev/pgdog/blob/af2edf54fc00aad788db5037e439ccff7caeb8e6/pgdog/src/frontend/router/parser/rewrite/statement/mod.rs#L147 and here https://github.com/pgdogdev/pgdog/blob/af2edf54fc00aad788db5037e439ccff7caeb8e6/pgdog/src/frontend/router/parser/rewrite/statement/mod.rs#L125-L131

Not a big deal, but at some point in the near/far future, we should consolidate all of our walks into one. It's cheap but it's not free.

levkk (approved):
Very nice!

PR #1506, opened 2026-09-07
feat: restore startup session parameters on DISCARD ALL

### LINKED ISSUE #1475: Complete session state cleanup for DISCARD ALL

murex971:
## Summary
- `DISCARD ALL` now restores session to the values from the client's startup message.
- Runtime `SET` values are dropped; parameters that were never in the startup packet stay unset and are reset on a held backend.
- `pgdog.pin` taken after connect is released, so the backend is no longer stuck after `DISCARD ALL`.

closes #1475

### DIFF
--- a/pgdog/src/frontend/client/query_engine/discard.rs
+++ b/pgdog/src/frontend/client/query_engine/discard.rs
@@ -1,3 +1,4 @@
+use crate::frontend::router::parameter_hints::PGDOG_PIN;
 use crate::net::{CommandComplete, Protocol, ReadyForQuery};
 
 use super::*;
@@ -23,6 +24,7 @@ impl QueryEngine {
             self.advisory_locks.clear();
             context.prepared_statements.close_all();
             self.backend.unlisten_all();
+            self.reset_session_params(context).await?;
             self.check_lock();
         }
         let bytes_sent = context
@@ -33,6 +35,32 @@ impl QueryEngine {
             ])
             .await?;
         self.stats.sent(bytes_sent);
+        Ok(())
+    }
+
+    async fn reset_session_params(
+        &mut self,
+        context: &mut QueryEngineContext<'_>,
+    ) -> Result<(), Error> {
+        context.params.restore_startup(context.startup_params);
+        self.manual_lock = context
+            .params
+            .get(PGDOG_PIN)
+            .and_then(|value| value.as_str())
+            .map(|value| matches!(value, "true" | "t"))
+            .unwrap_or_default();
+        self.comms.update_params(context.params);
+
+        // Parameters the client SET while holding this server were sent straight
+        // through, and `link_client` only replays what it knows diverged. Reset the
+        // server so the startup values are re-applied onto a clean session.
+        if self.backend.connected() {
+            self.backend.execute("RESET ALL").await?;
+            self.backend
+                .link_client(context.id, context.params, None)
+                .await?;
+        }
+
         Ok(())
     }
 }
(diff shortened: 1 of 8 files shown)


pgdog/src/frontend/client/query_engine/discard.rs
```diff
@@ -35,4 +37,30 @@ impl QueryEngine {
...
+            .get(PGDOG_PIN)
+            .and_then(|value| value.as_str())
+            .map(|value| matches!(value, "true" | "t"))
+            .unwrap_or_default();
+        self.comms.update_params(context.params);
+
+        // Parameters the client SET while holding this server were sent straight
+        // through, and `link_client` only replays what it knows diverged. Reset the
+        // server so the startup values are re-applied onto a clean session.
+        if self.backend.connected() {
```
levkk: According to https://www.postgresql.org/docs/current/sql-discard.html, `DISCARD ALL` cannot be sent inside a transaction, so this situation should not happen. We should return an error to the client (same as Postgres does) if the client sends `DISCARD ALL` while `context.transaction()` is true.

pgdog/src/net/parameter.rs
```diff
@@ -263,6 +263,16 @@ impl Parameters {
         self.transaction_local_params.remove(&name);
     }
 
+    /// Restore parameters to the values supplied in the startup message,
+    /// dropping everything changed since with `SET`.
+    pub(crate) fn restore_startup(&mut self, startup: &Parameters) {
+        self.params.clone_from(&startup.params);
+        self.transaction_params.clear();
```
levkk: Same note as above: `DISCARD ALL` cannot be sent inside transactions.

pgdog/src/net/parameter.rs
```diff
@@ -263,6 +263,16 @@ impl Parameters {
         self.transaction_local_params.remove(&name);
     }
 
+    /// Restore parameters to the values supplied in the startup message,
+    /// dropping everything changed since with `SET`.
+    pub(crate) fn restore_startup(&mut self, startup: &Parameters) {
+        self.params.clone_from(&startup.params);
```
levkk, line 269: Pretty odd way of writing `self.params = startup.clone()` :). Just curious, why `clone_from`?

murex971: Probably a tiny optimization, but I used `clone_from` because it can reuse the map’s existing allocation 😅

levkk: TIL!

pgdog/src/net/parameter.rs
murex971: handled

pgdog/src/frontend/client/query_engine/discard.rs
murex971: adressed

levkk: You shouldn't be sending `RESET ALL` to servers. Just update the client params back to what they were and let our parameter synchronization logic synchronize the server(s) on next checkout.

levkk (approved):
LGTM!

PR #1507, opened 2026-09-07
fix(types): hash 0.0 and -0.0 to the same value

MaxFreedomPollard:
The `Float` wrapper in pgdog-config and pgdog-vector and the `Double` wrapper in pgdog-postgres-types compare zeros numerically, so `0.0` and `-0.0` are equal, but their `Hash` impls hashed the raw bit pattern, which differs by the sign bit, so two equal values hash differently and a `HashMap` or `HashSet` keyed on them keeps them apart. `Datum` derives `Hash`, so this reaches the cross-shard GROUP BY buffer in `backend::pool::connection::aggregate`, where a `-0.0` returned by one shard forms its own group instead of merging with the `0.0` from another. Each of the three `Hash` impls now hashes the bits of positive zero when the value is zero, the same way Postgres normalizes zero in `hashfloat4` and `hashfloat8`. New tests cover it in all three crates (`test_negative_zero_hashes_like_zero` in pgdog-config and pgdog-vector, `test_negative_zero_datums_hash_like_zero` over `Datum::Float`, `Datum::Double` and `Datum::Vector`, `test_double_negative_zero_groups_with_zero`, and `test_distinct_values_still_hash_apart` so zero does not swallow other values), the two existing hash consistency tests that asserted the old behaviour now assert equality, and `cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings` and `cargo nextest run` are clean for the three crates.

### DIFF
--- a/pgdog-postgres-types/src/datum.rs
+++ b/pgdog-postgres-types/src/datum.rs
@@ -343,6 +343,15 @@ mod tests {
     use super::*;
     use bytes::{BufMut, BytesMut};
     use std::assert_matches;
+    use std::collections::HashSet;
+    use std::collections::hash_map::DefaultHasher;
+    use std::hash::{Hash, Hasher};
+
+    fn hash_of(datum: &Datum) -> u64 {
+        let mut hasher = DefaultHasher::new();
+        datum.hash(&mut hasher);
+        hasher.finish()
+    }
 
     #[test]
     fn test_multidimensional_text_array_falls_back_to_unknown() {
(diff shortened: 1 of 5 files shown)


pgdog-postgres-types/src/datum.rs
```diff
@@ -378,4 +387,30 @@ mod tests {
         assert_matches!(datum, Datum::Unknown(_));
         assert_eq!(datum.encode(Format::Binary).unwrap(), input);
     }
+
+    #[test]
+    fn test_negative_zero_datums_hash_like_zero() {
```
sgrif, line 392: Since we're testing the hash impl on `Float` explicitly, I don't think we need to test the derived impl here.

sgrif (approved):
CLA needs to be signed but this looks good

pgdog-postgres-types/src/datum.rs
MaxFreedomPollard: Dropped the `Datum` test in `ae24b74`.

levkk (approved):

MaxFreedomPollard:
CLA signed.

PR #1510, opened 2026-09-09
fix: allow omnisharded reads with a shard directive inside transactions

rlittlefield:
A plain SELECT from an omnisharded table inside a read/write transaction failed with cannot write to an omnisharded table with a shard directive whenever a directive was present, whether `/* pgdog_shard */`, `/* pgdog_sharding_key */`, or `SET pgdog.shard`. The guard used `Route::is_write()` to decide whether a statement mutates, but under the conservative read/write strategy every statement in a read/write transaction is marked a write so it runs on the primary. Any application that keys its transactions and reads a replicated table inside one hit this on a multi-shard cluster.

Route now carries a statement-intrinsic mutates flag: DML, DDL, data-modifying CTEs, locking clauses, write functions and advisory locks. The omnisharded coverage check uses it; read is unchanged and still drives primary/replica placement. The guard is per statement, so a later omni write in the same pinned transaction is still rejected.

This finishes the fix for #1374. #1384 and #1398 diagnosed the same false positive and relaxed the check when schema sharding is used exclusively, leaving the mixed `sharded_schemas` + `sharded_tables` case for later ("the error will still throw ... I also added a test for that for later"). Scoping the exemption by statement kind instead of by config covers that case: a plain read of a pinned shard is correct whether the table is omnisharded or schema-sharded. The "invalid cases" test in `integration/rust/tests/integration/cross_shard_omni_check.rs`, which asserted the pinned read fails, now checks the read succeeds and that an INSERT into an omnisharded table in the same pinned transaction is still rejected.

Covered by parser and route unit tests plus engine tests against Postgres for reads inside pinned transactions.

### DIFF
--- a/integration/rust/tests/integration/cross_shard_omni_check.rs
+++ b/integration/rust/tests/integration/cross_shard_omni_check.rs
@@ -140,7 +140,7 @@ async fn test_cross_shard_omni_check_valid_cases() {
     }
 }
 
-// All of these should throw errors.
+// Reads inside a pinned transaction succeed; writes to omnisharded tables throw.
 #[tokio::test]
 async fn test_cross_shard_omni_check_invalid_cases() {
     let admin = admin_sqlx().await;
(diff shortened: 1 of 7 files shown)


levkk (approved):
Got to love PRs where number of test lines exceeds the actual change by a factor of 10. Thank you sir!

PR #1511, opened 2026-09-09
fix: broadcast omnisharded writes inside CTEs and stop treating CTE names as tables

rlittlefield:
Two routing bugs for writes to omnisharded tables inside CTEs. A write hidden in a CTE, such as `WITH ins AS (INSERT INTO omni ... RETURNING id) SELECT id FROM ins`, was sent to a single round-robin shard even without a directive, silently diverging the table. And the statement walker counted CTE names as tables, so an unqualified reference to a CTE made the statement non-omnisharded and bypassed the omni-write-with-directive guard entirely.

Mutating omni selects now broadcast like an omni INSERT, and CTE names are excluded from table extraction. Schema-qualified names still refer to the real table.

Covered by parser unit tests plus an engine test against Postgres that verifies a CTE write lands on every shard and returns a single RETURNING row.


levkk (approved):
Image

PR #1517, opened 2026-09-10
fix cross-shard ORDER BY on non-projected columns

### LINKED ISSUE #1135: [Query Engine] ORDER BY on non-projected column can produce incorrect cross-shard ordering
### Description
When `ORDER BY` references a column not present in the projected output, cross-shard ordering can be incorrect.

### Example
`SELECT id FROM users ORDER BY name;`

In this case, PgDog resolves sort columns from `RowDescription`; if `name` is not present, the key may be skipped, causing incorrect global ordering.

murex971:
- Temporarily project missing ORDER BY columns during query rewriting.
- Use hidden columns for correct global cross-shard sorting.
- Remove hidden columns before sending rows and RowDescription to clients.
- Support direct-shard results.
- closes #1135


pgdog/src/frontend/client/query_engine/query.rs
```diff
@@ -146,6 +150,12 @@ impl QueryEngine {
         context: &mut QueryEngineContext<'_>,
         mut message: Message,
     ) -> Result<(), Error> {
+        if !self.backend.is_multishard() {
+            drop_projected_columns(
```
levkk, line 154: Aren't we calling `drop_columns` twice now? Once in multi shard binding, and once here, so twice for multi-shard queries and once for direct-to-shard queries.

pgdog/src/frontend/client/query_engine/query.rs
```diff
@@ -515,6 +525,30 @@ impl QueryEngine {
     }
 }
 
+fn drop_projected_columns(
```
levkk, line 528: I would probably push this concern into the connection binding (e.g.. `read` method). The architecture is as such that piece of code handles the cross-shard/direct-to-shard message manipulation. The query engine doesn't actually know whether it's talking to one or more servers, so this leaks it here.

pgdog/src/frontend/router/parser/query/select.rs
```diff
@@ -135,16 +135,33 @@ impl QueryParser {
...
+            *column = if column.asc() {
+                OrderBy::Asc(helper.projected_column + 1)
+            } else {
+                OrderBy::Desc(helper.projected_column + 1)
+            };
+        }
+        let order_by = order_by
+            .into_iter()
+            .map(|(_, order_by)| order_by)
+            .collect::<Vec<_>>();
```
sgrif, line 164:
```suggestion
        let helpers = cached_ast.rewrite_plan.projection.order_by_helpers();
        let order_by = Self::select_sort(stmt, router_context.bind)
            .into_iter()
            .map(|(position, column)| {
                if let Some(helper) = order_by.iter().find(|helper| position == helper.sort_position) {
                    if column.asc() {
                        OrderBy::Asc(helper.projected_column + 1)
                    } else {
                        OrderBy::Desc(helper.projected_column + 1)
                    }
                } else {
                    column
                }
            })
            .collect::<Vec<_>>();

```

pgdog/src/frontend/router/parser/query/select.rs
```diff
@@ -301,17 +315,18 @@ impl QueryParser {
     fn select_sort(
         stmt: &nodes::SelectStmt,
         params: Option<StatementParameters<'_>>,
-    ) -> Vec<OrderBy> {
+    ) -> Vec<(usize, OrderBy)> {
```
sgrif, line 318: While we're touching this signature, what do you think about changing this to return `impl Iterator` instead?

pgdog/src/frontend/router/parser/rewrite/statement/order_by.rs
```diff
@@ -0,0 +1,190 @@
...
+                    },
+                )
+            })
+            .collect::<Vec<_>>();
+
+        if helpers.is_empty() {
+            return;
+        }
+
+        let (targets, mappings): (Vec<_>, Vec<_>) = helpers.into_iter().unzip();
```
sgrif, line 56: We can skip an intermediate vec here.

```suggestion
            });
        let (targets, mappings): (Vec<_>, Vec<_>) = helpers.unzip();
        
        if targets.is_empty() {
            return;
        }
```

pgdog/src/frontend/router/parser/rewrite/statement/order_by.rs
```diff
@@ -0,0 +1,190 @@
...
+        .take(fields.len().saturating_sub(1))
+        .filter_map(Node::as_str)
+        .collect::<Vec<_>>();
+
+    match qualifiers.as_slice() {
+        [] => true,
+        [table] => order_by.table == Some(*table),
+        [schema, table] => order_by.schema == Some(*schema) && order_by.table == Some(*table),
+        _ => false,
+    }
```
sgrif, line 107: What do you think about this instead? The way `qualifiers` is constructed feels noisy to me, and we can skip the intermediate vec.

```suggestion
    let mut fields = column_ref.fields().iter();
    if !matches!(fields.next_back(), Some(Node::A_Star(_))) {
        return false;
    }

    let table = fields.next_back();
    let schema = fields.next_back();

    match (schema, table) {
        (None, None) => true,
        (Some(table), None) => order_by.table == Some(*table),
        (schema, table) => order_by.schema == schema && order_by.table == table,
    }
```

sgrif (commented):
It might be out of scope for this PR, but a refactor I've been wanting to make is to pass through the original target list of the select statement, and changing the code to keep those columns rather than keep track of which ones to drop. This could probably even just be a length instead of a list since we're always going to append any helpers to the end of the list

murex971:
@sgrif  I am closing this and handling it in https://github.com/pgdogdev/pgdog/pull/1545
will handle suggested fixes there

PR #1522, opened 2026-09-10
fix: move connections on reload even when a replica is removed

jkaczman:
Previously, if a reload occurred, and if a replica was removed, **all** connections for **all** `Pool`s (whether affiliated or not with the removed replica), would **not** be moved to the new `Pool`s.

This is because `Cluster::can_move_conns_to` asserted that **every** existing `Pool` (address) must be present in the new configuration before attempting to move any. 

To fix this,
- I removed all the upfront `has_compatible_address_with` checks, and since `LoadBalancer::move_conns_to` already uses `has_compatible_address_with` itself, no further changes were needed there.
- Additionally, I modified `Address::compatible` to account for `database_number` being different when a replica is removed.

Added a regression test for this. Removed some tests that asserted the old functionality.

Fixes https://github.com/pgdogdev/pgdog-enterprise/issues/615

### DIFF
--- a/pgdog/src/backend/databases.rs
+++ b/pgdog/src/backend/databases.rs
@@ -425,8 +425,8 @@ impl Databases {
 
             if let Some(dest) = dest
                 && cluster.can_move_conns_to(dest)
+                && cluster.move_conns_to(dest)?
             {
-                cluster.move_conns_to(dest)?;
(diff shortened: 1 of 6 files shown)


levkk (approved):
Nice!

PR #1525, opened 2026-09-11
feat: cross-account RDS IAM auth via server_iam_assume_role

lyupan-cb:
RDS IAM tokens are signed with PgDog's single ambient identity, and RDS IAM auth is evaluated in the database's own AWS account, so a backend in a different account can't be reached today (region can be overridden, but not account).

Add an optional per-user `server_iam_assume_role` (an IAM role ARN). When it is set and `server_auth = "rds_iam"`, PgDog assumes that role via STS and signs the auth token with the assumed credentials instead of the ambient identity.

- new `server_iam_assume_role` field on the user config
- rds_iam::token assumes the role (aws-config AssumeRoleProvider) when set; otherwise the path is unchanged (ambient identity)
- warn when the field is set without `server_auth = "rds_iam"`
- backward compatible: unset = current behavior; TokenCache/monitor refresh are unchanged
- unit tests for config parsing, Address plumbing, and the assume-role decision (pure, no AWS required)

### DIFF
--- a/pgdog-config/src/users.rs
+++ b/pgdog-config/src/users.rs
@@ -835,56 +835,6 @@ vault_refresh_percent = 60
     }
 
     #[test]
-    fn test_server_iam_assume_role_parses() {
-        let source = r#"
-[[users]]
-name = "app"
-database = "tenant"
-server_auth = "rds_iam"
-server_iam_region = "us-west-2"
-server_iam_assume_role = "arn:aws:iam::111122223333:role/pgdog-rds-connect"
-"#;
-        let users: Users = toml::from_str(source).unwrap();
-        let user = users.users.first().unwrap();
-        assert_eq!(user.server_auth, ServerAuth::RdsIam);
-        assert_eq!(
-            user.server_iam_assume_role.as_deref(),
-            Some("arn:aws:iam::111122223333:role/pgdog-rds-connect")
-        );
-    }
-
-    #[test]
-    fn test_server_iam_assume_role_defaults_none() {
-        let source = r#"
-[[users]]
-name = "app"
-database = "tenant"
-server_auth = "rds_iam"
-"#;
-        let users: Users = toml::from_str(source).unwrap();
-        let user = users.users.first().unwrap();
-        assert!(user.server_iam_assume_role.is_none());
-    }
-
-    #[test]
-    fn test_server_iam_assume_role_warns_without_rds_iam_but_parses() {
-        // A misconfiguration (assume-role set but server_auth != rds_iam) is a
-        // warning, not a hard error — check() must not reject it.
-        let mut users = Users {
-            users: vec![User {
-                name: "app".into(),
-                database: "tenant".into(),
-                server_auth: ServerAuth::Password,
-                password: Some("p".into()),
-                server_iam_assume_role: Some("arn:aws:iam::111122223333:role/x".into()),
-                ..Default::default()
-            }],
-            ..Default::default()
-        };
-        users.check(&Config::default());
-    }
-
-    #[test]
     fn test_vault_path_appears_in_passwords_as_vault_static_role() {
         let user = User {
             name: "alice".into(),
--- a/pgdog-config/src/users.rs
+++ b/pgdog-config/src/users.rs
@@ -126,6 +126,13 @@ impl Users {
                 }
             }
 
+            if user.server_iam_assume_role.is_some() && user.server_auth != ServerAuth::RdsIam {
+                warn!(
+                    r#"user "{}" (database "{}") sets "server_iam_assume_role" but "server_auth" is not "rds_iam"; the assume-role will be ignored"#,
+                    user.name, user.database
+                );
+            }
+
             if !user.database.is_empty() && !user.databases.is_empty() {
                 warn!(
                     r#"user "{}" is configured for both "{}" and "{:?}", defaulting to "{:?}""#,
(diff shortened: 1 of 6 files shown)


pgdog-config/src/users.rs
```diff
@@ -822,6 +834,56 @@ vault_refresh_percent = 60
...
+database = "tenant"
+server_auth = "rds_iam"
+"#;
+        let users: Users = toml::from_str(source).unwrap();
+        let user = users.users.first().unwrap();
+        assert!(user.server_iam_assume_role.is_none());
+    }
+
+    #[test]
+    fn test_server_iam_assume_role_warns_without_rds_iam_but_parses() {
```
levkk, line 870: This test doesn't assert and can't validate what it's testing (you're testing a log output). This is common to bridge the code coverage gap, which I guess is fine, but feels incomplete, especially in Rust, since we know the code compiles, but we don't know if it's doing what we want it to, so we might be better off leaving that code uncovered as an indicator of a testing gap.

pgdog-config/src/users.rs
```diff
@@ -822,6 +834,56 @@ vault_refresh_percent = 60
         assert_eq!(user.vault_refresh_percent, Some(60));
     }
 
+    #[test]
+    fn test_server_iam_assume_role_parses() {
```
levkk, line 838: This tests serde deserialization, which is redundant.

pgdog-config/src/users.rs
```diff
@@ -822,6 +834,56 @@ vault_refresh_percent = 60
...
+        let user = users.users.first().unwrap();
+        assert_eq!(user.server_auth, ServerAuth::RdsIam);
+        assert_eq!(
+            user.server_iam_assume_role.as_deref(),
+            Some("arn:aws:iam::111122223333:role/pgdog-rds-connect")
+        );
+    }
+
+    #[test]
+    fn test_server_iam_assume_role_defaults_none() {
```
levkk, line 857: Same comment as above.

pgdog/src/backend/auth/rds_iam.rs
```diff
@@ -45,14 +50,64 @@ fn resolve_region(addr: &Address) -> Result<String, Error> {
     })
 }
 
+/// Whether `addr` must sign its RDS IAM token with an assumed role (cross-account
+/// RDS IAM), and with what parameters. Returns `None` for the normal same-account
+/// path, where the token is signed with PgDog's ambient identity. Pure so the
+/// account-spanning decision is unit-testable without AWS.
+fn assume_role_plan(addr: &Address, region: &str) -> Option<AssumeRolePlan> {
```
levkk, line 57: Why not make it `AssumeRolePlan::new`?

pgdog/src/backend/auth/rds_iam.rs
```diff
@@ -45,14 +50,64 @@ fn resolve_region(addr: &Address) -> Result<String, Error> {
...
+    session_name: String,
+}
+
+/// Build the AWS config used to sign the RDS IAM token.
+///
+/// Same-account: PgDog's ambient identity (`load_defaults`). Cross-account: the
+/// ambient identity is used only as the *source* credentials to assume
+/// `server_iam_assume_role` in the database's account, and the token is signed
+/// with those assumed credentials.
+async fn sign_config(addr: &Address, region: &str) -> SdkConfig {
```
levkk, line 83: It would be good to name functions by what they do, e.g., `build_aws_sdk_config`, making its docs almost redundant.

pgdog/src/backend/auth/rds_iam.rs
```diff
@@ -1,10 +1,15 @@
...
-use aws_config::{BehaviorVersion, Region};
+use aws_config::sts::AssumeRoleProvider;
+use aws_config::{BehaviorVersion, Region, SdkConfig};
 use aws_sdk_rds::auth_token::{AuthTokenGenerator, Config as AuthTokenConfig};
 
 use crate::backend::{Error, pool::Address};
 
+/// STS session name used when PgDog assumes a role to mint a cross-account RDS
+/// IAM token.
+const IAM_ASSUME_ROLE_SESSION_NAME: &str = "pgdog-rds-connect";
```
levkk, line 11: Any chance you'd want this configurable? I know these are just used for debugging in CloudTrail...but still, might be something the user would want to set? I'm not too sure, not familiar with how you guys use this identifier and if it matters if it accidentally conflicts.

pgdog-config/src/users.rs
lyupan-cb: Thanks for the suggestion! Redundant tests are removed in the last commit

pgdog-config/src/users.rs
lyupan-cb: Removed

pgdog-config/src/users.rs
lyupan-cb: Removed

pgdog/src/backend/auth/rds_iam.rs
lyupan-cb: Done — moved it to AssumeRolePlan::new

pgdog/src/backend/auth/rds_iam.rs
lyupan-cb: Done, renamed `sign_config`  to `build_aws_sdk_config`

pgdog/src/backend/auth/rds_iam.rs
lyupan-cb: `RoleSessionName` doesn't have to be unique as it's only an audit label in CloudTrail. We don't really use this identifier for now but added for debugging in case needed. Happy to make it per-user configurable or change to some other value if you have a preference.

levkk: Ok sounds good to me!

PR #1526, opened 2026-09-11
refactor(resharding): migrate copy_data to the tasks api

meskill:
Refactor table copies logic to the api/tasks.

- move out logic for copying from orchestrator/publisher/table to separate data_sync entities
- extend the info types in pgdog-stats for data copy
- migrate to tasks context that is sent to ee
- add WorkerPool entity to distribute load between resharding replicas more evenly - use work stealing with cap configured with `resharding_parallel_copies` for single replica. Previously the distribution was not fair and some of the replicas could acquire significantly more work
- add more tests coverage for the process.

I verified the copy_data benchmarks and didn't notice any performance changes after the migration. (the benchmarks doesn't cover multiple replicas though, but I think that should improve performance)


levkk (approved):
GH can't even render the diff, so I'm gonna go with "yes" here. One thing that's worth trying is testing the multi-replica stuff. We have an AWS test bed for this, so could be fun to try it out.

PR #1528, opened 2026-09-11
feat: expose total_wait_time and avg_wait_time to OpenMetrics

### LINKED ISSUE #1456: Add total_wait_time counter metric in addition to maxwait
PgDog currently exposes the maxwait metric, which reports the maximum amount of time a client has waited for a connection from the pool.
While maxwait is useful for identifying individual periods of high connection wait time, it does not provide enough information to measure the overall impact of pool exhaustion over time.

Proposal

Add a new total_wait_time counter metric that tracks the cumulative connection wait time that could be used with rate() to calculate the amount of connection wait time accumulated over a given interval.
There are already similar metrics like total_query_time, total_connect_time

rimrakhimov:
Add `total_wait_time` and `avg_wait_time` to OpenMetrics. 

- Simply exposes already existing `wait_time` stat to OpenMetrics
- Closes https://github.com/pgdogdev/pgdog/issues/1456

### DIFF
--- a/pgdog/src/stats/pools.rs
+++ b/pgdog/src/stats/pools.rs
@@ -69,6 +69,8 @@ impl Pools {
         let mut avg_idle_xact_time = vec![];
         let mut total_query_time = vec![];
         let mut avg_query_time = vec![];
+        let mut total_wait_time = vec![];
+        let mut avg_wait_time = vec![];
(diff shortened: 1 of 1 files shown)


levkk (approved):
Good find!

PR #1534, opened 2026-09-13
perf(copy_data): extend benchmark to verify replica load distribution

meskill:
Update the copy_data benchmark to make sure the https://github.com/pgdogdev/pgdog/pull/1526 is better in some conditions during the copy.

Default comparison with old main:
Screenshot 2026-09-12 163318

Comparison with the old main and toxi enabled - multiple replicas where single replica is artificially slowed down with higher latency.
Screenshot 2026-09-12 163329


benches/resharding/schema_sync/prepare.sh
```diff
@@ -0,0 +1,58 @@
...
+        --config "${PGDOG_CONFIG}" \
+        --users "${SETUP_DIR}/users.toml" \
+        schema-sync \
+        --from-database source \
+        --to-database destination \
+        --publication bench_schema \
+        "$@" >/dev/null
+}
+
+echo ">>>>> prepare ${STAGE}: dropping bench_schema on destinations"
```
levkk, line 38: I would use `=>` or something similar instead of the arrows. Those are too close to git conflict markers.

PR #1540, opened 2026-09-14
fix(resharding): use replica_session role and disable FK validation

meskill:
Handle the FK constraints during resharding process:

When we copy tables in parallel we make the snapshots of tables at the specific moment of time and because of this we can have tables on destination with different WALs (that's why we save lsn for tables after copy) and that's could make data inconsistent for example for FK and other cross-table constraints. 

The error could happen in both cases: when we do schema sync for the constraints (they may fail on apply), or during the replication (that may fail because part of the transaction could be already in some table snapshot, but not another table snapshot).

To resolve this:
1. Use `SET session_replication_role=replica` on destination in order to disable rules and triggers during the update. That requires additional specific GRANT or superuser, so an error handling for this with a little explanation was added.
2. Add `NOT VALID` for FK during replication, so the FK won't be triggered when added (the role doesn't disable it automatically).

This basically disables this validations to the sake of assumption that we have consistent data on source and we do the proper copy that maintains the consistency so during the replication we'll eventually reach the consistent state.


levkk (approved):
Perfect! Thank you 🙏

meskill:
Added check for destination pg versions to avoid possible issues with FK NOT VALID for partitioned tables since it's only supported from pg18. See https://www.postgresql.org/docs/17/sql-altertable.html#SQL-ALTERTABLE-DESC-ADD-TABLE-CONSTRAINT and https://www.postgresql.org/docs/18/release-18.html#RELEASE-18-CONSTRAINTS
for partitioned that will fallback to the old FK to not break the whole resharding

PR #1541, opened 2026-09-14
feat: re-write date/time functions (calls + DEFAULT) in omni inserts

### LINKED ISSUE #1521: [Rewrites] Omnisharded inserts with timestamps should use proxy time
For inserts targeting omnisharded columns, we should intercept and rewrite calls to `now()` (and other timestamp-related functions) and replace them with a value we generate. This makes sure the value is the _same_ on all shards.

**Example**

```sql
INSERT INTO omni (name, created_at) VALUES ('test', now());
INSERT INTO omni (name, created_at) VALUES ('test', '2026-09-10T10:00:01.342Z');
```

Make sure to handle this for `TIMESTAMPTZ` (UTC) _and_ `TIMESTAMP`; the latter needs to fetch the database timezone and use that.

Additionally, for queries that don't specify `now()` and rely on a database default, we should inject our timestamp, just like we do for primary keys (auto_id).

jkaczman:
When we perform an `INSERT` into an omnisharded table, right now, the functions within the statement (or `COLUMN_DEFAULT` in the table schema) are executed separately on each shard. In the case of time/date functions, this means that they will drift slightly apart, as their execution time will not be the exact same.

To fix this, this PR re-writes all those time/date functions ([Postgres ref](https://www.postgresql.org/docs/current/functions-datetime.html)), specifically: `current_date`, `current_time`, `current_timestamp`, `clock_timestamp`, `localtime`, `localtimestamp`, `now`, `statement_timestamp`, `timeofday`, and `transaction_timestamp`.

It handles both calls within a `VALUES` list, as well as is compatible with functions specified within `COLUMN_DEFAULT` in the table schema. It's also compatible with simple protocol, extended protocol, and prepare/execute.

Refactoring
- Pushed `TransactionType` into a `Transaction` type, so that the time the transaction started could be saved for use with now() /  other functions that rely on transaction start time.
- Created a `PreparedPlan` type to encapsulate everything the PreparedStatement cache stores for Prepare statements (instead of passing around separate fields in a tuple).
- Pushed `GeneratedId` and its corresponding parameter number into a `GeneratedParam` type.

Shortfalls:
- If the client specified a timezone, it only works with strings like "UTC" and "America/New_York"; Postgres supports other types such as `INTERVAL '-08:00' HOUR TO MINUTE`. Right now it errors to prevent incorrect behavior falling through. We previously didn't support any kind of timezone parsing whatsoever.

Docs: https://github.com/pgdogdev/docs/pull/114.
Helm: https://github.com/pgdogdev/helm/pull/140.
Fixes https://github.com/pgdogdev/pgdog/issues/1521.


integration/rust/tests/integration/omni_timestamps.rs
```diff
@@ -0,0 +1,681 @@
+use std::ops::Sub;
```
levkk, line 1: Just want to make sure (I didn't spot it in the test code) that we're testing implicit timestamp injection, i.e.:

```sql
CREATE TABLE users (email VARCHAR UNIQUE, created_at TIMESTAMPTZ);
INSERT INTO users (email) VALUES ('john@pgdog.dev') RETURNING *;
```

This should be rewritten to:

```sql
INSERT INTO users (email, created_at) VALUES ('john@pgdog.dev', <timestmap>);
```

Btw, to make this easier, we can (and currently do!) stack rewriters, e.g.,: you can have one rewriter inject `now()` into the above query and then have the subsequent one rewrite `now()` to an actual timestamp value. This is how `auto_id` works: it just injects `pgdog.unique_id()` and then `unique_id` rewriter changes that to a bigint. Anyway, I haven't read your implementation yet, so maybe this is exactly how you did it, just leaving this thought here before I forget :)

integration/pgdog.toml
```diff
@@ -44,6 +44,9 @@ enabled = false
 shard_key = "ignore"
 split_inserts = "error"
 # primary_key = "rewrite"
+#
+# TODO: This doesn't strictly affect DEFAULT; might want to re-name
```
levkk: Do you mean it doesn't handle this case?

```sql
INSERT INTO users (email, created_at) VALUES ('john@pgdog.dev', DEFAULT) RETURNING *
```

pgdog/src/backend/replication/logical/subscriber/context.rs
```diff
@@ -55,6 +55,7 @@ impl StreamContext {
             &BufferedQuery::Prepared(parse),
             &ast_context,
             &mut PreparedStatements::default(),
+            QueryTimestamps::now(),
```
levkk: This reminds me: we should make sure to disable these rewrites during logical replication. We have a constructor for the `Cluster`  somewhere that we use to disable all rewrite rules, we should add database defaults = ignore in there too.

pgdog/src/backend/prepared_statements.rs
```diff
@@ -655,17 +665,33 @@ impl PreparedStatements {
...
             return Ok(());
         }
 
         let mut parameter_description = ParameterDescription::from_bytes(message.payload())?;
-        parameter_description.rewrite_data_types(mappings);
+        if let Some(mappings) = mappings {
+            parameter_description.rewrite_data_types(mappings);
+        }
+        if let Some(client_params) = client_params {
+            parameter_description.truncate(client_params as usize);
```
levkk, line 705: This relies on the invariant that all injected parameters are appended.  It would be good to note that somewhere.

pgdog/src/frontend/client/query_engine/rewrite.rs
```diff
@@ -38,9 +38,14 @@ impl QueryEngine {
         if let Some(query) = query {
             let cluster = self.backend.cluster()?;
             let ast_ctx = AstContext::from_cluster(cluster, context.params);
-            let ast = Cache::get().query(&query, &ast_ctx, context.prepared_statements)?;
+            let timestamps = context.timestamps();
```
levkk: You could pass it down inside `AstContext`. That way, you don't have to change the method signature below. Just a nit.

pgdog/src/frontend/client/mod.rs
```diff
@@ -697,6 +703,7 @@ impl Client {
 
             if timer.is_none() {
                 timer = Some(Instant::now());
+                self.statement_start = Utc::now();
```
levkk, line 704: Maybe we use just one timer, e.g., `Utc::now()`? Getting the current time is cheap, but not free.

pgdog/src/frontend/client/transaction_type.rs
```diff
@@ -1,3 +1,69 @@
...
+pub(crate) struct Transaction {
+    transaction_type: TransactionType,
+    start_time: DateTime<Utc>,
+}
+
+impl Transaction {
+    pub(crate) fn new(transaction_type: TransactionType) -> Self {
+        Self {
+            transaction_type,
+            start_time: Utc::now(),
```
levkk, line 16: You already have the request start time in the client you could pass down here. May not be worth the code spaghetti, since getting the current time is cheap, but it would be consistent.

pgdog/src/frontend/prepared_statements/global_cache.rs
```diff
@@ -94,6 +95,7 @@ impl GlobalCache {
         //       to use `RewritePlan` for `offset_plan` too (which isn't possible; see comment below)
         rewrite_plan: &RewritePlan,
         offset_plan: Option<OffsetPlan>,
+        generated_ids: Vec<(u16, GeneratedId)>,
```
levkk: nit: Could be good to have these params be a typed struct so it's a bit more self-documenting.

pgdog/src/frontend/prepared_statements/global_cache.rs
```diff
@@ -131,12 +134,18 @@ impl GlobalCache {
     }
 
     /// Rewrite prepared statement in the global cache.
-    pub(crate) fn rewrite(&mut self, parse: &Parse) {
+    pub(crate) fn rewrite(&mut self, parse: &Parse, client_params: u16) {
```
levkk, line 138: Don't forget to document the new param (or rename it).

pgdog/src/frontend/router/parser/cache/ast.rs
```diff
@@ -68,13 +67,11 @@ impl Deref for Ast {
 
 impl Ast {
     /// Parse statement and run the rewrite engine, if necessary.
-    pub(super) fn new(
+    pub(super) fn with_context(
```
levkk: nit: or just `new` :)

pgdog/src/frontend/router/parser/cache/context.rs
```diff
@@ -23,6 +23,8 @@ pub(crate) struct AstContext<'a> {
     pub(crate) user: &'a str,
     /// Search path for table lookups.
     pub(crate) search_path: Option<&'a ParameterValue>,
+    /// Allows `timestamp` types to use the Client's local time when excecuting a `TimeFunction`
+    pub(crate) timezone: Option<&'a ParameterValue>,
```
levkk, line 28: I think we should probably use the server timezone (it will be available as part of https://github.com/pgdogdev/pgdog/blob/87bca544c5cab1ea4c3d62e0270ef32b97025e28/pgdog/src/frontend/client/mod.rs#L371 so you can easily grab it and inject into the client state.

It's a safe assumption that all shards have the same timezone (today is sept 15, 2026, the day that shall live in infamy, when we discover that this is wrong 😆 )

pgdog/src/frontend/router/parser/rewrite/statement/auto_id.rs
```diff
@@ -551,11 +551,16 @@ mod tests {
...
+            timezone: None,
         });
         let mut plan = Default::default();
         let ast = make::try_owned(|mem| {
             let mut copy = mem.make_unique(&*ast.into_inner());
-            plan = rewriter.maybe_rewrite(copy.as_mut().into_iter().next().unwrap(), mem)?;
+            plan = rewriter.maybe_rewrite(
+                copy.as_mut().into_iter().next().unwrap(),
+                mem,
+                crate::frontend::client::QueryTimestamps::now(),
```
levkk: Changing function interfaces across 20 different files is so annoying haha. We should make it a rule to use structs as arguments (whenever the argument count exceeds 1? 2?)? That way, we can easily inject a new arg.

levkk: Or maybe we don't, and this way we know when context is leaking between modules.

levkk: Talking to myself here...

pgdog/src/frontend/router/parser/rewrite/statement/mod.rs
```diff
@@ -185,8 +198,45 @@ impl<'a> StatementRewrite<'a> {
             self.limit_offset(&select, &mut plan);
         }
 
+        let timestamp_rewrite = matches!(
+            config().config.rewrite.omni_database_defaults,
+            RewriteMode::Rewrite | RewriteMode::RewriteOmni | RewriteMode::RewriteOmniGlobal
```
levkk: Could do a negative instead, e.g.,:

```rust
!matches(RewriteMode::Ignore)
```

pgdog/src/frontend/router/parser/rewrite/statement/plan.rs
```diff
@@ -1,20 +1,23 @@
...
+use crate::net::messages::bind::{Format, Parameter};
+use crate::net::{Bind, Parameters, Parse, ProtocolMessage, Query};
+use crate::unique_id::UniqueId;
 
+/// TODO: Document that this is also stored in PreparedStatement cache.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub(crate) enum GeneratedId {
     UniqueId,
     Sequence(SequenceCall),
+    ProxyTime(TimeFunction),
```
levkk, line 27: Nice! That fits really nicely in here. We could rename this to `GeneratedValues` or something else...eventually.

pgdog/src/frontend/router/parser/rewrite/statement/plan.rs
```diff
@@ -94,29 +98,59 @@ impl RewritePlan {
...
-            let param = match format {
-                Format::Binary => Parameter::new(&id.to_be_bytes()),
-                Format::Text => Parameter::new(itoa::Buffer::new().format(id).as_bytes()),
+
+            let param = match id {
+                MyResponse::Int(id) => match format {
+                    Format::Binary => Parameter::new(&id.to_be_bytes()),
+                    Format::Text => Parameter::new(itoa::Buffer::new().format(id).as_bytes()),
+                },
+                // TODO: This could use pgdog-postgres-types/src/timestamp.rs
```
levkk: Should imo, so we keep data type manipulation consistent.

pgdog/src/frontend/router/parser/rewrite/statement/timestamp.rs
```diff
@@ -0,0 +1,616 @@
...
+
+                    Some(node)
+                }
+                _ => Some(node),
+            }),
+        );
+    }
+
+    /// Iterates through Schema to find DEFAULT columns
+    /// Adds the column to target list & all the values lists (ParamRef or String)
```
levkk, line 557: There it is! Found it. Ignore my previous previous previous previous comment.

levkk (approved):
Image

PR #1545, opened 2026-09-14
refactor: apply route-dependent rewrites after routing

### LINKED ISSUE #1135: [Query Engine] ORDER BY on non-projected column can produce incorrect cross-shard ordering

### LINKED ISSUE #1425: [Cross shard] temporary columns for `stddev()` visible in client response

murex971:
Aggregate helpers, ORDER BY helpers, and LIMIT/OFFSET pushdown now run only after the route is known, so direct shard execution keeps original SQL and does not leak helper columns.

closes #1425  closes #1135


levkk:
👀

pgdog/src/backend/prepared_statements.rs
```diff
@@ -192,12 +205,25 @@ impl PreparedStatements {
...
+                            let mut describe = self.internal_describe(bind.statement());
                             self.state.add('2');
                             if self.config.level.rewrite_anonymous() {
                                 let mut bind = bind.clone();
                                 bind.anonymize();
+                                if let Some(describe) = &mut describe {
+                                    describe.anonymize();
+                                }
+                                if let Some(describe) = describe {
+                                    return Ok(HandleResult::PrependProtocolMessageRewrite {
```
levkk, line 228: Q: why do we need this? I'm not too familiar with this piece of code, would be good to have a small example explaining.

levkk (commented):
This looks good to me I think. It would be good to get feedback from @jkaczman after we merge #1541 (both touch the rewriter heavily). Should be easy to resolve conflicts, but just want to double check one doesn't break the other and vice versa.

pgdog/src/backend/prepared_statements.rs
```diff
@@ -518,9 +544,27 @@ impl PreparedStatements {
             // it still holds.
             close: expired.then(|| ProtocolMessage::Close(Close::named(name))),
             parse: ProtocolMessage::Parse(parse),
+            describe: None,
         }))
     }
 
+    fn internal_describe(&mut self, name: &str) -> Option<ProtocolMessage> {
```
levkk, line 571: Oh I see. The cross-shard `Describe` has different columns that direct-to-shard, so we need to rewrite it. I'm remembering now.

pgdog/src/frontend/prepared_statements/global_cache.rs
```diff
@@ -39,12 +40,20 @@ impl MemoryUsage for GlobalCache {
...
             + self.names.memory_usage()
+            + self.cross_shard_variants.memory_usage()
             + self.counter.memory_usage()
             + self.unused.capacity() * 1usize.memory_usage()
     }
 }
 
 impl GlobalCache {
+    pub(crate) fn cross_shard_variant_name(&self, name: &str) -> Option<String> {
+        let variant_name = format!("{name}_cross_shard");
```
levkk: This should be a helper. I saw it copy/pasted somewhere else.

pgdog/src/frontend/prepared_statements/global_cache.rs
```diff
@@ -137,10 +146,41 @@ impl GlobalCache {
         }
     }
 
+    pub(crate) fn cross_shard_variant(&mut self, name: &str, query: &str) -> Option<String> {
+        if let Some(variant_name) = self.cross_shard_variant_name(name) {
+            return Some(variant_name);
+        }
+        let variant_name = format!("{name}_cross_shard");
```
levkk: There it is

jkaczman (commented):
LGTM! 

It would be great if you could document certain decisions e.g. why you chose to move from an explicit `A_Const` node with the limit + offset to an `A_Expr` executing on Postgres. I think I understand the tradeoffs, but it would be great to see the author's POV when reading the code.

I would also be curious to know if the benchmark from before vs after changed at all.

murex971:
Benchmarked release builds of main and this PR using pgbench prepared protocol against the same three shards. Each case used a 3s warmup followed by three 15s runs at 10 and 50 clients.

Results were effectively neutral: direct and cross-shard AVG ranged from −0.6% to +1.0%; comparable ORDER BY was ~1% slower (expected from post-route bookkeeping). CPU was unchanged (+0.13%), with ~476 KiB higher peak RSS.

pgdog/src/frontend/router/parser/route.rs
```diff
@@ -402,12 +404,12 @@ impl Route {
         self.is_cross_shard() && self.is_write()
     }
 
-    pub(crate) fn aggregate_rewrite_plan(&self) -> &AggregateRewritePlan {
-        &self.rewrite_plan
+    pub(crate) fn projection_rewrite_plan(&self) -> &ProjectionRewritePlan {
```
jkaczman: Could replace this field with pub(crate) and get rid of the getter/setter

pgdog/src/frontend/router/parser/query/select.rs
```diff
@@ -16,7 +16,7 @@ impl QueryParser {
     ///
     pub(super) fn select(
         &mut self,
-        cached_ast: &Ast,
+        _cached_ast: &Ast,
```
jkaczman, line 19: I think we're fine to remove this as it's no longer needed in this func

pgdog/src/frontend/router/parser/rewrite/statement/aggregate/mod.rs
```diff
@@ -1,39 +1,22 @@
...
-            return Ok(());
-        }
-
-        let aggregate = Aggregate::parse(select, schema);
-        if aggregate.is_empty() {
-            return Ok(());
-        }
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum HelperKind {
+    Count,
```
jkaczman, line 9: Why remove the comments that were originally in aggregate/plan.rs?

pgdog/src/frontend/router/parser/rewrite/statement/order_by.rs
```diff
@@ -0,0 +1,199 @@
+use pg_raw_parse::{ConstValue, Node, make, nodes};
+
```
jkaczman, line 2: It would be great to have some comments explaining some of the logic in this file.

pgdog/src/frontend/router/parser/rewrite/statement/projection.rs
```diff
@@ -0,0 +1,255 @@
...
+        self.aggregate_helpers.push(helper);
+    }
+
+    pub(crate) fn add_order_by_helper(&mut self, helper: OrderByHelper) {
+        self.order_by_helpers.push(helper);
+    }
+}
+
+#[derive(Debug, Default, Clone)]
+pub(crate) struct RewriteOutput {
```
jkaczman, line 93: It looks like this struct is just a basic wrapper around `ProjectionRewritePlan`. Why not just use `ProjectionRewritePlan` by itself?

pgdog/src/frontend/router/parser/rewrite/statement/projection.rs
```diff
@@ -0,0 +1,255 @@
...
+            .iter()
+            .map(|helper| helper.projected_column)
+            .chain(
+                self.order_by_helpers
+                    .iter()
+                    .map(|helper| helper.projected_column),
+            )
+    }
+
+    pub(crate) fn aggregate_helpers(&self) -> &[AggregateHelper] {
```
jkaczman, line 75: These fields can be exposed pub(crate) and used directly if we're providing basic getters/setters without any special invariants enforced or anything.

pgdog/src/frontend/router/parser/rewrite/statement/order_by.rs
```diff
@@ -0,0 +1,199 @@
...
+    select.target_list().iter().any(|target| {
+        (unqualified && target.name() == name)
+            || matches!(target.val(), Node::ColumnRef(projected)
+                if star_covers(projected, column)
+                    || same_column(projected, column)
+                    || (unqualified
+                        && projected.fields().into_iter().next_back().and_then(Node::as_str)
+                            == name))
+    })
+}
```
jkaczman, line 47: I think this will fail with something like `SELECT p.* FROM products p ORDER BY price`. Looks like it would work fine if it was `SELECT * FROM products ORDER BY price` or `SELECT p.* FROM products p ORDER BY p.price`, but not if we have a mixed case of one being qualified and the other not.

pgdog/src/frontend/router/parser/route.rs
```diff
@@ -402,12 +404,12 @@ impl Route {
...
+        &self.projection_rewrite_plan
     }
 
-    pub(crate) fn set_rewrite_plan(&mut self, plan: AggregateRewritePlan) {
-        self.rewrite_plan = plan;
+    pub(crate) fn set_projection_rewrite_plan(&mut self, plan: ProjectionRewritePlan) {
+        self.projection_rewrite_plan = plan;
```
jkaczman, line 412: This is another case we could probably get rid of the getter/setter

pgdog/src/frontend/prepared_statements/global_cache.rs
```diff
@@ -13,6 +13,10 @@ use fnv::FnvHashSet as HashSet;
 
 use super::*;
 
+fn cross_shard_variant_name(name: &str) -> String {
+    format!("{name}_cross_shard")
+}
```
jkaczman, line 18: I wonder if we could re-use the previous allocation somehow, here's a super simple example (probably not optimal):

buf = "x__pgdog_1"
cross_shard_variant = buf
base = buf[1..]

pgdog/src/frontend/router/parser/rewrite/statement/order_by.rs
```diff
@@ -0,0 +1,199 @@
...
+        let needs_helper = match node {
+            Node::ColumnRef(column) => !projects_column(select, column),
+            Node::A_Expr(_) => source.is_some(),
+            _ => false,
+        };
+        if !needs_helper {
+            continue;
+        }
+
+        let projected_column = select.target_list().len() + helpers.len();
```
jkaczman, line 94: I think we shouldn't rely on this column index; it gives us the parse tree entries, but not the actual output columns (considering when * is used). I wonder if it would be better to use what you do in aggregate.rs again (`decoder.row_description().field_index(&helper.alias) else {`). Right now, I think there are edge cases where this can accidentally delete user columns / sort on the wrong column.

pgdog/src/frontend/router/parser/rewrite/statement/order_by.rs
```diff
@@ -0,0 +1,199 @@
...
+    fn skips_qualified_star_select() {
+        let (sql, plan) = rewrite("SELECT products.* FROM products ORDER BY products.id");
+
+        assert!(!sql.contains("__pgdog_order_col"));
+        assert!(plan.is_noop());
+    }
+
+    #[test]
+    fn projects_column_not_covered_by_qualified_star() {
+        let (sql, plan) = rewrite("SELECT a.* FROM a JOIN b ON a.id = b.a_id ORDER BY b.score");
```
jkaczman, line 175: This should also test the index (going off of my comment above)

pgdog/src/frontend/router/parser/rewrite/statement/aggregate/mod.rs
murex971: thought too trivial, will add back :sweat_smile:

pgdog/src/frontend/router/parser/rewrite/statement/order_by.rs
murex971: ah! yes, need to handle this

pgdog/src/frontend/router/parser/rewrite/statement/order_by.rs
murex971: yes, I thought this for the above case handling

pgdog/src/frontend/prepared_statements/global_cache.rs
murex971: I don’t think the buffer-sharing works here. The base name is created by next_name() at Parse time and already lives in self.names long before we know a cross-shard variant is needed, so there’s nothing left to slice from. Recovering the base is already allocation free via strip_suffix("_cross_shard").

pgdog/src/frontend/router/parser/rewrite/statement/projection.rs
murex971: it was introduced when this struct return more than one field, but later only 1 field remained, so removing now.

murex971:
@jkaczman ready for re-review

PR #1552, opened 2026-09-15
fix: tables should be considered omnisharded if not explicity listed

### LINKED ISSUE #1551: [Query router] Omnisharded check doesn't default to omni if explicitly configured
**PgDog version**
`main`

**Description**
All tables are omnisharded by default unless configured in `[[sharded_tables]]`. However, if we configure `[omnisharded_tables]`, the query router will only treat those tables as omnisharded and won't default to omnisharded anymore.

This is not correct. It should check the config in any scenario, and if the table isn't specified, default to omnisharded.

**Configuration**

```toml
[general]
omnisharded_sticky = true

[omnisharded_tables]
database = "prod"
tables = ["table_a"]

[[sharded_tables]]
database = "prod"
column = "user_id"
```

In this scenario, all following queries  need to be omnisharded:

```sql
SELECT * FROM tables_a LIMIT 1;
SELECT * FROM table_a INNER JOIN table_b [...]; -- table_a is explicitly listed, table_b is not => default omni
```

Additionally, tables in `[omnisharded_tables]` should take priority over `[[sharded_tables]]`. I.e., if a table is listed in the former, and contains the sharding key, it is still treated as omnisharded.

### Reproduction

https://github.com/mscrivo/pgdog-omni-join-repro

ygxio:
Fixes: #1551


levkk (approved):
Yup! Looks right! Would be really great to add a couple tests, specifically to cover those ambiguous cases where the column is specified in `[[sharded_tables]]` but the table is listed in `[omnisharded_tables]`.

ygxio:
i pushed some test, missing tests for column only sharding rule

ygxio:
Done

levkk (approved):
Perfect, tyvm!

PR #1553, opened 2026-09-15
Lock backend to frontend when in progress cancel requests are in flight to prevent cancel race between two frontends

### LINKED ISSUE #1505: Transaction pooling can route a stale `CancelRequest` to the next frontend
**PgDog version**

`v0.1.57` (main, commit `7594279`, transaction pool mode)

**Description**

In transaction pooling, a `CancelRequest` captured for frontend A can cancel a later query from frontend B after A's physical backend is returned to the pool and assigned to B. Direct PostgreSQL does not reproduce this cross-frontend cancellation.

PgDog snapshots A's backend key, releases the pool lock, and only then opens a new TCP connection to send the cancel [`pool_impl.rs`](https://github.com/pgdogdev/pgdog/blob/7594279/pgdog/src/backend/pool/pool_impl.rs#L286-L292) [`server.rs`](https://github.com/pgdogdev/pgdog/blob/7594279/pgdog/src/backend/server.rs#L450-L454). During that window, transaction pooling can give the returned backend to a waiting frontend [`inner.rs`](https://github.com/pgdogdev/pgdog/blob/7594279/pgdog/src/backend/pool/inner.rs#L289-L305), so the in-flight key can cancel B instead of A [`taken.rs`](https://github.com/pgdogdev/pgdog/blob/7594279/pgdog/src/backend/pool/taken.rs#L39-L69).

**Reproduce**

1. Start PostgreSQL 16 directly and PgDog in transaction pool mode with pool size 8. Set `DIRECT_DSN` and `PGDOG_DSN` to the two endpoints.
2. Install `psycopg[binary]` and run:

```python
#!/usr/bin/env python3
from contextlib import suppress
import threading
import time

import psycopg


RUNS = 5


def suppress_query(conn, sql: str) -> None:
    with suppress(BaseException):
        conn.execute(sql)


def one(dsn: str) -> bool:
    a, *holders, b = [psycopg.connect(dsn, autocommit=True) for _ in range(9)]
    workers = []
    for conn in holders:
        thread = threading.Thread(target=suppress_query, args=(conn, "select pg_sleep(1)"), daemon=True)
        thread.start()
        workers.append(thread)
    thread = threading.Thread(target=suppress_query, args=(a, "select pg_sleep(.5)"), daemon=True)
    thread.start()
    workers.append(thread)
    time.sleep(.03)

    error = []

    def wait() -> None:
        try:
            b.execute("select pg_sleep(.5)")
        except BaseException as exc:
            error.append(exc)

    thread = threading.Thread(target=wait, daemon=True)
    thread.start()
    workers.append(thread)
    cancels = [threading.Thread(target=a.cancel_safe, kwargs={"timeout": 1}, daemon=True) for _ in range(2)]
    for thread in cancels:
        thread.start()
    for thread in cancels:
        thread.join(1)
    for thread in workers:
        thread.join(2)
    for conn in [a, b, *holders]:
        conn.close()
    return bool(error and getattr(error[0], "sqlstate", None) == "57014")


for label, dsn in (("direct", DIRECT_DSN), ("pgdog", PGDOG_DSN)):
    print(f"{label}: cross_client_57014={sum(one(dsn) for _ in range(RUNS))}/{RUNS}")
```

**Expected behavior**

The query submitted by B must not be canceled by A's `CancelRequest`. Both endpoints should report zero cross-client `57014` results.

**Actual behavior**

The second cancel races backend reassignment, so the count is timing-dependent. A complete run on `7594279` produced:

```text
direct: cross_client_57014=0/5
pgdog: cross_client_57014=5/5
```

KennanHunter:
# The issue

Fixes #1505, a cancel request can be made by the frontend, then before pgdog's cancel packet actually lands at Postgres, the physical backend that frontend was using can be returned to the pool and handed to a different frontend. The cancel then lands on the new frontend's query instead of the original one.

Concretely, in `Pool::cancel`:

```rust
let key = self.lock().cancel_key(id).cloned();   // phase 1: snapshot under lock
if let Some(key) = key {
    Server::cancel(self.addr(), key).await?;      // phase 2: TCP send, no lock
}
```

# The fix

Taken picks up a `backend_cancels_in_flight` map (backend pid → count). Pool::cancel takes a `CancelLease`, a small RAII guard in pool/cancel.rs, that bumps the count going in and drops it on the way out whether the cancel succeeded or not. `Pool::checkin` looks at the map: if there's a pending cancel for that backend, the check-in waits until the count hits zero, otherwise it's the exact same sync path as today. I went with a count instead of a bool so a couple of cancels landing on the same backend at once just work. `Pool::move_conns_to` bails with a retryable `Error::CancelInFlight` if anything's on the wire.

`Server::cancel` now hangs on until Postgres sends EOF. Postgres only does that after it's validated the request and poked the target backend, so that's our signal the cancel is really done. There's a 5s cap so a wedged server can't pin backends forever.

Cost-wise, only waiters on that one backend notice if a cancel is in flight, the rest of the pool's connections are unblocked.

Tests: `test_cancel_request_cross_frontend_race` (A starts a pg_sleep, cancel fires, B grabs the backend and runs its own pg_sleep, make sure B doesn't get 57014), cancel_leases_stack_on_same_backend, and the repro script from #1505 as `integration/python/test_cancel_race.py`.

# Next steps

## Style

Theres some duplication in logic between locking due to cancel requests, and locking due to advisory locks. I think for right now it's fine, but if we see more locking behaviors it might be a good idea to merge those in some way.

Could look something like this, with a single queue for locked connections on a pool side.

```rust
#[derive(Default)]
struct Pin {
    client_locked: bool,
    cancels_in_flight: u32,
}

pub(super) struct Taken {
    frontend_to_cancel: HashMap<FrontendPid, Checkout>,
    backend_to_frontend: HashMap<BackendPid, FrontendPid>,
    pins: HashMap<BackendPid, Pin>,
}
```

## Scoping

An interesting idea longer-term is that the whole lease mechanism could disappear if pgdog tagged every forwarded statement with a `/* pgdog:<FrontendPid> */` comment and routed cancels through a helper connection running `SELECT pg_cancel_backend(pid) FROM pg_stat_activity WHERE pid = $1 AND state = 'active' AND query LIKE '/* pgdog:F<n> */%'`

The problem is it gets pretty complicated injecting the tag cleanly across every query, especially when we consider mechanisms like `PREPARE`, `EXECUTE`, and `COPY`.

## Benchmarking

I think we should probably benchmark this on a cancel heavy workflow with actual network delay between frontend/backend, to verify that nothing goes haywire. Lmk if there's any changes you'd like to see.


KennanHunter:
Might want CANCEL_ACK_TIMEOUT to be configurable, not sure.

PR #1555, opened 2026-09-16
fix(auth): preserve user config on passthrough password change

mehcode:
When `passthrough_auth` is set to `enabled_allow_change` or `enabled_plain_allow_change`, a client connecting with a rotated password replaces that user's entry with the one built from the startup message, resetting settings like `pool_size`, `pooler_mode`, etc.


levkk (approved):
Good catch!

PR #1556, opened 2026-09-16
feat: re-write uuid functions (calls + DEFAULT) in omni insert

jkaczman:
Adds support to rewrite UUID functions (`uuidv4`, `uuidv7`, `gen_random_uuid`) for omni inserts on top of the support for date-time functions (from #1541). Some refactoring to push those two classes of functions into separate files and generalize handling a bit more.

It does **not** support specifying an interval argument within uuidv7() yet.


levkk (approved):
Image

PR #1557, opened 2026-09-16
fix premature ReadyForQuery in pub/sub pipelines

### LINKED ISSUE #1512: Transaction-pool pipeline `LISTEN` sends `ReadyForQuery` before `Sync`
**Version**

`v0.1.57` (main `7594279`, transaction pool mode with pub/sub enabled; `pub_sub_channel_size = 100`)

**Description**

A `LISTEN` followed by another command in one psycopg pipeline succeeds against PostgreSQL, but the transaction-pool endpoint reports an unexpected pipeline synchronization result instead of completing the pipeline. In the extended query protocol, `ReadyForQuery` terminates a `Sync` response, not an individual command ([PostgreSQL protocol flow](https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-EXT-QUERY)).

Transaction mode routes `Command::Listen` to the local [`listen()`](https://github.com/pgdogdev/pgdog/blob/7594279/pgdog/src/frontend/client/query_engine/mod.rs#L249-L257) path. That path calls [`command_complete()`](https://github.com/pgdogdev/pgdog/blob/7594279/pgdog/src/frontend/client/query_engine/pub_sub.rs#L6-L15), which sends both `CommandComplete` and `ReadyForQuery` before the client `Sync` ([`pub_sub.rs`](https://github.com/pgdogdev/pgdog/blob/7594279/pgdog/src/frontend/client/query_engine/pub_sub.rs#L56-L70)). Session mode instead forwards `LISTEN` through `execute()`.

**Reproduce**

Run with psycopg 3 against a direct PostgreSQL endpoint and a PgDog transaction-pool endpoint. `DIRECT_DSN` and `PGDOG_DSN` refer to those endpoints.

```python
import logging

import psycopg


logging.getLogger("psycopg").setLevel(logging.ERROR)


def run(label, dsn):
    conn = psycopg.connect(dsn, autocommit=True)
    try:
        with conn.pipeline() as pipeline:
            with conn.cursor() as cur:
                cur.execute("LISTEN pgdog_pipeline_listen")
                cur.execute("SELECT 1")
                pipeline.sync()
        print(f"{label}: {conn.execute('SELECT 42').fetchone()}")
    except Exception as exc:
    finally:
        conn.close()


run("direct", DIRECT_DSN)
run("pgdog", PGDOG_DSN)
```

**Expected behavior**

Both endpoints complete the pipeline and print:

```text
direct: (42,)
pgdog: (42,)
```

**Actual behavior**

`7594279` prints:

```text
direct: (42,)
pgdog: InternalError: unexpected result status from query: PIPELINE_SYNC
```

murex971:
Use existing command-response path so responses follow the original frontend messages:
- Parse → ParseComplete
- Bind → BindComplete
- Execute → CommandComplete
- Flush → no response
- Sync → ReadyForQuery

Session-mode behavior is unchanged.

closes #1512


levkk (approved):
Nice!

PR #1558, opened 2026-09-16
test: stabilize asynchronous database fixtures

dipeshbabu:
Several CI fixtures depend on scheduling speed or wait for only part of their asynchronous work. This causes timeouts and premature row-count checks on otherwise valid runs.

- Wait for the mirrored table to appear, with a bounded timeout and immediate failure for unexpected database errors.
- Create the load-balancer stress test's table once before starting its 100 concurrent clients, then drop it after the clients finish.
- Wait for initial COPY to finish, drain pgbench's server connections, and observe a replication sentinel on each destination shard before stopping data sync and checking row counts.
- Set `jit=off` when opening asyncpg test connections so type metadata lookup fits within the existing query timeout.
- Use bulk string operations to generate large pgbench response payloads.

Validation: Rust/Go formatting and shell/Python syntax checks pass. PostgreSQL verified the generated payload lengths at 2, 61, 62, 63, and 1,000,000 bytes. All 27 jobs passed in [pre-PR validation](https://github.com/dipeshbabu/pgdog/actions/runs/35107083777), covering builds, strict linting, unit/doc tests, and the 21-suite integration matrix.


levkk (approved):
Thank you!

PR #1559, opened 2026-09-16
fix(types): compare and hash intervals by PostgreSQL duration

dipeshbabu:
Cross-shard interval comparisons use individual fields, so `1 mon` sorts after `31 days` and equivalent values such as `1 day` and `24 hours` form separate groups. This also produces incorrect `MIN`/`MAX` results.

Use one allocation-free `i128` duration for ordering, equality, and hashing, matching PostgreSQL's 30-day months and 24-hour days without overflowing the interval range.

Regression tests cover equivalent values, mixed signs, extreme ranges, and multi-shard ordering/grouping/min/max in text and binary formats.

Validation: all 27 jobs pass in [CI](https://github.com/dipeshbabu/pgdog/actions/runs/35408696969), including strict formatting/Clippy, unit/doc tests, Linux/macOS builds, and all 21 integration suites on PostgreSQL 18.

PR #1560, opened 2026-09-16
fix(aggregate): handle empty and singleton variance inputs

dipeshbabu:
Cross-shard variance/stddev queries fail when every shard returns NULL partial sums. Population variants also return NULL for a single finite value, while sample variants incorrectly return NaN for a singleton NaN.

Check the input count before inspecting partial sums, return NULL for insufficient sample inputs, and return zero for a finite singleton population before performing arithmetic.

Regression tests compare empty, all-NULL, singleton, and mixed-NULL inputs with PostgreSQL across integer, numeric, real, and double precision types.

Validation: all 27 jobs pass in [CI](https://github.com/dipeshbabu/pgdog/actions/runs/35408697000), including strict formatting/Clippy, unit/doc tests, Linux/macOS builds, and all 21 integration suites on PostgreSQL 18.


levkk:
Same ask as #1559, let's stack these and merge together!

dipeshbabu:
Rebased onto main after #1558 merged. This PR now contains only the variance fix. All 27 validation jobs pass.

PR #1565, opened 2026-09-16
Bump CI to Postgres 18

jkaczman:


levkk:
I'm obviously paranoid, but make sure to pin the action to a specific hash; this is a common way for people to take over other people's repos and inject stuff

PR #1567, opened 2026-09-16
Serve buffered client requests before sending the shutdown notice

scottjacobsen:
## What

On graceful shutdown (SIGINT or admin `SHUTDOWN`) `Client::run` raced the shutdown token against the client socket:

- with tokio's default random `select!` order, a request whose bytes had already been read could lose to the shutdown branch and be dropped, and the socket was closed with unread data, so drivers saw a connection reset instead of the 57P01 notice;
- a partially received extended-protocol request (Parse+Bind without Sync) survived the cancelled `buffer()` poll only to be wiped by `client_request.clear()` on re-entry, so the client got a backend error for its Execute+Sync before the FATAL.

## Changes

- `select!` is `biased` with the shutdown wake-up polled last, so a request that is ready when the loop polls is served before the notice.
- The notice is deferred while `stream_buffer` holds unconsumed bytes or `client_request.is_partial()`.
- `buffer()` only clears a completed request. This also fixes the same message loss when an async backend message (for example a NotificationResponse) arrived between a client's Parse and Sync.
- Logging: each notice is logged at debug, and the shutdown summary reports idle clients notified versus clients waited on, from one `comms().clients()` snapshot (idle and not holding a lock).

## What this does not cover

A request whose bytes arrive after the loop has decided to send the notice is still cut off; that is inherent to a server-initiated close. A client that sends part of a request and then goes quiet is now waited on until `shutdown_timeout`, like an idle-in-transaction client, instead of being disconnected immediately.

## Tests

Two new tests in `graceful_shutdown.rs`: a Query already on the socket when shutdown starts is answered before 57P01, and Parse+Bind, shutdown, Execute+Sync completes before 57P01 (fails without the `buffer()` change). `cargo fmt` and `cargo clippy --all-targets` are clean; the four graceful shutdown tests pass.

Context: found while teaching Rails' ActiveRecord to handle the shutdown notice. libpq treats an ErrorResponse received while idle as a notice, so the next statement hits a closed socket.


levkk:
👁️

pgdog/src/frontend/listener.rs
```diff
@@ -146,14 +147,20 @@ impl Listener {
     async fn execute_shutdown(&self) {
         let shutdown_timeout = config().config.general.shutdown_timeout();
 
+        let comms = comms();
+        let clients = comms.clients();
+        let idle_clients = clients
+            .values()
+            .filter(|client| client.stats.state == State::Idle && !client.stats.locked)
```
levkk, line 154: Why the `!locked` check?

PR #1568, opened 2026-09-16
feat: (always) re-write funcs that rely on transaction start time

jkaczman:
Previously, when a `now()` date-time function (or other function that relies on the Transaction start time) was ran on multiple Shards in a singular Transaction, the `now()` output values would differ across connections within the Transaction.

This behavior is now changed to re-write this class of functions in all contexts.


levkk (approved):
Awesome!

PR #1571, opened 2026-09-17
fix: verify passthrough credentials before storing them

ziomarco:
On 2026-08-31 one of our production databases became unusable for a user because of how passthrough auth handles a wrong password. When a client logged in with a mistyped password, PgDog stored it as the pool's server credential. The `min_pool_size` maintenance loop then retried that password roughly every 333 ms per pool, which showed up as a login storm on the Postgres side, and since we do not allow password changes through passthrough, the correct password was rejected with `NoPassthroughPasswordChange` until we restarted PgDog.

This PR fixes that in three pieces. None of them depend on the auth plugin work.

First, new or changed passthrough credentials are verified with a one-shot server connection before they are stored (`databases::add_passthrough`, `ConnectReason::PassthroughVerify`). A bad password gets the normal authentication error and never reaches the pool. If the server cannot be reached the client gets a distinct `PassthroughVerificationFailed` error. Every shard of the database is checked, concurrently, because a stored credential poisons all of them; the results are combined asymmetrically on purpose, so a rejection from any shard rejects the login while a shard that is simply unreachable does not, otherwise one down shard would block every new login.

Because the client decides when a verification happens, the cost is bounded: `backend::passthrough` caps how many verifications run at once and remembers a credential the server rejected for a few seconds, so an application reconnecting in a loop with a wrong password costs one server connection instead of one per attempt. The memory is keyed by a hash of (user, database, credential) and capped, so a client that corrects its password is never held back by its earlier mistake and a client that varies the user name cannot grow the map.

Second, credentials learned through passthrough are tagged (`PasswordSource::Passthrough`). When the server later rejects one with 28P01, for example after a password rotation, PgDog evicts it: a discovered user is removed and a learned password is cleared, so the next client login stores the current one instead of the pool staying poisoned. Eviction rebuilds every pool and re-runs the configuration checks, so it is handed to a task of its own rather than done inside `Server::connect`; it re-reads the configuration and does nothing unless that exact credential is still stored, so nothing waits on it.

Third, a pool that just had its credentials rejected stops replenishing `min_pool_size` connections for 30 seconds. Connections for waiting clients still go through and a success clears the backoff.

While porting this onto main I made two related changes. Verification now runs against the entry the pool would actually store, so a configured user with `server_password` or an external identity provider is not verified with a client password the pool never uses. And a permitted password change no longer replaces the configured entry with a bare one, which used to drop settings like `pool_size`.

The live test that checks a wrong password against Postgres needs password authentication on the test server, same as the existing pool auth tests.


levkk:
You should be able to add a flag in the connection pool to check for server errors, and if it's the authentication kind, you can set a flag on the pool. Much simpler architecture.

> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

I don't think that's worth submitting most of the time; just send us the prompt, and we can do it :)

PR #1579, opened 2026-09-17
fix: avoid pinning failed advisory try locks

### LINKED ISSUE #1578: Failed pg_try_advisory_lock pins the client in transaction pooling
### Summary

In transaction pooling mode, `pg_try_advisory_lock(...)` returning `false` still marks the client as holding an advisory lock. PgDog then pins that client to its backend even though PostgreSQL granted no lock.

This is distinct from #1263 (avoiding pinning for advisory locks generally) and #1407 (cleanup after a client disconnects while holding a real lock). This report is about false bookkeeping after an unsuccessful try-lock.

### Version / configuration

- Reproduced with the `v0.1.56` source commit (`5d8152273c066f7cd36003c79477fa58c7c350d0`)
- Current `main` at `6b227afc6eeebfedcaa30329e6ff33b9e0164e55` is also affected
- `pooler_mode = "transaction"`
- `query_parser = "session_control_and_locks"`
- Client: pgx using cached statements / extended protocol

### Reproduction

1. On a direct PostgreSQL connection, acquire a session advisory lock for a chosen key.
2. Through PgDog, execute a parameterized `SELECT pg_try_advisory_lock($1::bigint)` for the same key.
3. The query correctly returns `false`.
4. On PostgreSQL, verify that the PgDog contender backend has zero granted advisory locks in `pg_catalog.pg_locks`.
5. Run `SHOW CLIENTS` in PgDog. The contender client nevertheless reports `locked = true` and remains pinned to that backend.

The same behavior occurs for the shared try-lock variant. It also reproduces when the result has additional columns, so consuming a single expected result shape would not be sufficient.

### Expected behavior

- A failed try-lock must not add advisory-lock ownership or pin the client.
- A successful try-lock should continue to pin the client.
- Unlocking the acquired lock should clear the pin.

### Impact

Applications commonly poll with `pg_try_advisory_lock`. Every failed attempt can leave a client unnecessarily pinned, preventing normal backend reuse and lifetime-based retirement and gradually reducing transaction-pool capacity.

### Root cause

The parser records advisory-lock intent from the SQL statement. At `ReadyForQuery`, the query engine merges that intent into the client's advisory-lock state without checking whether a `pg_try_advisory_lock` call actually returned `false`. The pool then sees non-empty advisory-lock state and keeps the backend pinned.

### Validation / possible fix direction

I validated a local fix that reconciles the tracked state with PostgreSQL's actual granted advisory locks after a statement that may change session advisory-lock state. The focused regression coverage uses the real PgDog binary and pgx extended protocol and checks:

- failed exclusive and shared try-locks do not pin;
- error responses do not falsely pin;
- successful try-locks still pin;
- unlocks unpin; and
- a backend is eligible for lifetime retirement after a failed try-lock.

The focused Rust tests, `cargo check -p pgdog`, `cargo build -p pgdog`, and the live pgx integration test pass. PgDog was exercised on Darwin ARM64 against PostgreSQL 16.15 on Linux ARM64; I have not yet run the PgDog binary itself on Linux ARM64.

I have a tested local patch and regression test available, but have not opened a PR.

ewhauser:
## Summary

- inspect `pg_try_advisory_lock*` result rows before merging optimistic lock bookkeeping
- avoid pinning a transaction-pooled client when a directly returned try-lock result is `false`
- preserve pinning for acquired, shared, reentrant, and transaction-spanning session locks
- add focused Rust unit and integration regression coverage for simple and extended protocol paths

PgDog previously merged advisory-lock intent from the parsed statement at `ReadyForQuery`, regardless of whether a try-lock succeeded. The parser now records the result columns for directly returned session try-lock calls, and the query engine inspects only those `DataRow` fields. It accounts for text and binary results, multiple result columns, and repeated rows without adding another backend query. Missing or undecodable results remain conservatively pinned.

Fixes #1578.

## Validation

- reproduced the failure with the real v0.1.56 PgDog binary and cached extended protocol
- `cargo test -p pgdog frontend::client::query_engine::test::advisory_lock -- --test-threads=1` (15 passed)
- focused parser advisory-lock tests (22 passed)
- Rust integration test target compiled with `cargo test -p integration_tests_rust --test mod integration::advisory_lock::test_failed_try_advisory_lock_does_not_pin_client --no-run`
- `cargo clippy -p integration_tests_rust --test mod --no-deps -- -D warnings`
- `cargo check -p pgdog`
- `cargo build -p pgdog`
- `git diff --check`

## Notes

- Result inspection is enabled only for parsed session try-lock calls whose boolean is returned directly (optionally through casts). More complex expressions remain conservatively optimistic.
- The live PgDog binary validation was on Darwin ARM64 against PostgreSQL 16.15 on Linux ARM64. The PgDog binary itself has not yet been exercised on Linux ARM64.


levkk (changes requested):
The advisory lock query will return an error if the lock failed to get acquired. It's much easier for us (and we do this already) to inspect the error we receive and remove the lock from our optimistic data structure.

ewhauser:
Thanks. I may be misunderstanding the suggested path, but PostgreSQL does not return an error when pg_try_advisory_lock* cannot acquire a contended lock - it returns a normal boolean false result.

Inspecting errors would cover actual execution failures, but not lock contention. Would you prefer that PgDog inspect the returned false value instead of reconciling through pg_locks? I used reconciliation because it remains correct for binary results, multiple result columns, nested lock calls, and reentrant locks, but I’m happy to align with the preferred result-inspection design.

levkk:
Nope, it's me who completely misunderstood 🤣  Sorry, getting late in the week (albeit it's only Thursday..). Yes, I think you're right, we need to inspect the result (we support parsing `DataRow` messages already at least), and make sure the boolean is set.

Running an extra query is too expensive, inspecting the result is cheap. Do make sure to do that if and only if we are running an advisory lock query (maybe flag it in the context) so we don't have to inspect _all_ results we receive.

ewhauser:
Thanks — updated in ec10acb7.

PgDog now records the directly returned `pg_try_advisory_lock*` result columns in the parsed route and inspects only those `DataRow` fields. A `false` result is excluded from the optimistic lock set; `true` remains pinned. The result decoder follows the requested text/binary format, including when Bind and Execute are separated by Flush, and maps multiple result columns and repeated rows. Missing or undecodable results remain conservative.

The `pg_locks` reconciliation query has been removed, so this adds no backend round trip and ordinary result rows are not inspected.

pgdog/src/frontend/client/query_engine/mod.rs
```diff
@@ -70,6 +71,10 @@ pub(crate) struct QueryEngine {
     pending_explain: Option<ExplainResponseState>,
     hooks: QueryEngineHooks,
     advisory_locks: AdvisoryLocks,
+    successful_try_advisory_locks: FnvHashSet<i64>,
```
levkk: I'm assuming I'm talking to your agent here, so I'm gonna be pretty demanding :) Can you move this to `AdvisoryLocks` structure just above this?

pgdog/src/frontend/client/query_engine/query.rs
```diff
@@ -174,12 +181,47 @@ impl QueryEngine {
...
+                    .advisory_locks()
+                    .try_lock_ids(),
+            );
             if let Some(state) = self.pending_explain.as_mut() {
                 state.annotated = true;
             }
             self.pending_explain = None;
         }
 
+        if code == 'D' {
```
levkk, line 190: Could you move this to its own method?

pgdog/src/frontend/client/query_engine/mod.rs
ewhauser: Half and half but be as demanding as you want :)

pgdog/src/frontend/client/query_engine/query.rs
ewhauser: Done in ec10acb7 — DataRow inspection is now encapsulated in AdvisoryLocks::process_data_row.

pgdog/src/frontend/client/query_engine/mod.rs
ewhauser: Done in ec10acb7 — all transient try-lock result state now lives in AdvisoryLocks.

PR #1587, opened 2026-09-18
reset session identity during pool cleanup

### LINKED ISSUE #1341: Transaction pooling leaks `SET ROLE` between connections

murex971:
`RESET ALL` does not reset `role` or `session_authorization`.
Run `SET SESSION AUTHORIZATION DEFAULT` before the existing dirty cleanup. This
restores the authenticated user and clears any active role, covering both
`SET ROLE` and `SET SESSION AUTHORIZATION`.

closes: #1341


pgdog/src/backend/pool/guard.rs
```diff
@@ -177,6 +195,49 @@ mod test {
...
+            .execute_checked(format!("CREATE ROLE {role}"))
+            .await
+            .unwrap();
+        guard
+            .execute_checked(format!("SET ROLE {role}"))
+            .await
+            .unwrap();
+        assert_eq!(identity(&mut guard).await, ("pgdog".into(), role.clone()));
+
+        guard.mark_dirty(true);
```
levkk, line 216: I think to resolve the original issue you want to make sure we detect the role change automatically and reset it back. This test artificially creates that condition.

I think we need to adjust the query parser set handler to detect SET ROLE and mark the server as dirty for cleanup.

murex971: Automatic dirty detection and test added to `query_engine`.
This can stay as just cleanup

pgdog/src/frontend/client/query_engine/set.rs
```diff
@@ -58,6 +59,13 @@ impl QueryEngine {
         }
 
         if self.backend.connected() {
+            if params.iter().any(|param| {
```
levkk, line 62: I suspect you want to mark the backend for cleanup before we connect. For example, consider this flow:

```sql
SET ROLE TO other_user;
SELECT 1;
```

Currently I believe we intercept `SET ROLE` and set it as a parameter (probably incorrectly).

The tests included in this PR test the connected case, but don't test the above case as well.

PR #1590, opened 2026-09-18
fix: set session_replication_role via SET instead of startup parameters

levkk:
### Problem

Managed Postgres is micromanaging our lives:

```
failed: Resharding connection denied for user "y" on database "x": FATAL: 42501 permission denied to set parameter "session_replication_role".
```

### Solution

Apparently this can be set via `SET` if the user has `REPLICATION` permissions.


pgdog/src/backend/server_options.rs
```diff
@@ -7,11 +7,13 @@ use crate::net::{Parameter, parameter::ParameterValue};
 #[derive(Debug, Clone)]
 pub(crate) struct ServerOptions {
     pub(crate) params: Vec<Parameter>,
+    pub(crate) session_replication_role: bool,
```
meskill, line 10: maybe will be simpler just add the execution here https://github.com/pgdogdev/pgdog/blob/87bca544c5cab1ea4c3d62e0270ef32b97025e28/pgdog/src/backend/replication/logical/subscriber/mod.rs#L24

levkk: This is done on the server/connection pool level, so no shard context! We could pass it the server but that's probably the same thing really.

levkk: I'll send it over to the guys and clean up my mess next week if that's cool. Late Friday, brain power is rapidly diminishing.

meskill: yes, thank you, I hope that'll work. Seems like there is really the difference for GCP and Azure when setting some params like this

PR #1596, opened 2026-09-20
fix(aggregate): preserve types of cast COUNT results

dipeshbabu:
Cross-shard `COUNT(*)::integer` returns 0 for six rows in binary format: PgDog sends bigint bytes under an int4 field description. Casts to smallint, real, and double precision also return incorrect values.

Keep the total as an i64, then produce the requested result type with checked integer conversions. Add overflow tests and text/binary PostgreSQL comparisons for ordinary, grouped, empty, and all-NULL inputs.

Pre-PR CI: [all 27 jobs passed](https://github.com/dipeshbabu/pgdog/actions/runs/35521347936) before opening this PR.


levkk (approved):
🚀

PR #1597, opened 2026-09-20
fix(build): upgrade jemallocator for ARM64 page support

### LINKED ISSUE #1591: Crash on aarch64 hosts with 64K pages: "<jemalloc>: Unsupported system page size"
**PgDog version**                                                                                                                                                                                                                 
   `v0.1.59` (also reproduces on `v0.1.58`)                                                                                                                                                                                          
                                                                                                                                                                                                                                     
   **Description**                                                                                                                                                                                                                   
   PgDog aborts immediately on startup on aarch64 Linux hosts that use 64K memory pages, before it reads its configuration. It should start and listen on port 6432.                                                                 
                                                                                                                                                                                                                                     
   The image runs fine on x86_64. On the affected arm64 hosts the container exits with code 133 with the output below.                                                                                                               
                                                                                                                                                                                                                                     
   Environment:                                                                                                                                                                                                                      
   - Image: `ghcr.io/pgdogdev/pgdog:v0.1.59` (arm64)                                                                                                                                                                                 
   - Node arch: `aarch64`                                                                                                                                                                                                            
   - OS/kernel: Oracle Linux Server 7.9, `5.4.17-2136.338.4.2.el7uek.aarch64` (OCI Ampere A1)                                                                                                                                        
   - Runtime page size: `getconf PAGE_SIZE` → `65536`                                                                                                                                                                                
                                                                                                                                                                                                                                     
   Root cause: PgDog statically links jemalloc as the Rust global allocator (`pgdog/src/main.rs`, `#[global_allocator] static GLOBAL: tikv_jemallocator::Jemalloc`), with no feature flag to opt out. jemalloc's `LG_PAGE` is fixed  
 at build time and defaults to the build host's page size. Upstream arm64 images are built on 4K-page runners (`PAGE = 4096`), and jemalloc's `pages_boot()` aborts when the runtime page size is larger than that (`os_page = 65536 > 4096`). Since jemalloc is statically linked and is the global allocator, no runtime option (`MALLOC_CONF`, `LD_PRELOAD`, etc.) can work around it.                                                                                
                                                                                                                                                                                                                                     
   Suggested fix: build the aarch64 artifacts with `JEMALLOC_SYS_WITH_LG_PAGE=16` (supported by `tikv-jemalloc-sys`). jemalloc only rejects `os_page > PAGE`, so `LG_PAGE=16` works on both 4K- and 64K-page arm64 hosts.            
                                                                                                                                                                                                                                     
   **Logs**                                                                                                                                                                                                                          
   It never reaches the logger, so `RUST_LOG=trace` produces nothing. The crash happens before configuration is parsed, on the first allocation:                                                                                     
                                                                                                                                                                                                                                     
   ```                                                                                                                                                                                                                               
   <jemalloc>: Unsupported system page size                                                                                                                                                                                          
   <jemalloc>: Unsupported system page size                                                                                                                                                                                          
   memory allocation of 4 bytes failed                                                                                                                                                                                               
   note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace                                                                                                                                                     
   ```                                                                                                                                                                                                                               
                                                                                                                                                                                                                                     
   Exit code: `133`. Kubernetes reports `CrashLoopBackOff`.                                                                                                                                                                          
                                                                                                                                                                                                                                     
   **Configuration**                                                                                                                                                                                                                 
   Irrelevant to this bug (the process dies before reading it), but for completeness (`pgdog.toml` equivalent, secrets redacted):                                                                                                    
                                                                                                                                                                                                                                     
   ```toml                                                                                                                                                                                                                           
   workers = 2                                                                                                                                                                                                                       
   default_pool_size = 15                                                                                                                                                                                                            
   checkout_timeout = 5000                                                                                                                                                                                                           
                                                                                                                                                                                                                                     
   [[databases]]                                                                                                                                                                                                                     
   name = "infra_primary"                                                                                                                                                                                                            
   host = "infra-rw.db.svc.cluster.local"                                                                                                                                                                                            
   port = 5432                                                                                                                                                                                                                       
   role = "primary"                                                                                                                                                                                                                  
                                                                                                                                                                                                                                     
   [[databases]]                                                                                                                                                                                                                     
   name = "infra_replica"                                                                                                                                                                                                            
   host = "infra-ro.db.svc.cluster.local"                                                                                                                                                                                            
   port = 5432                                                                                                                                                                                                                       
   role = "replica"                                                                                                                                                                                                                  
                                                                                                                                                                                                                                     
   [[users]]                                                                                                                                                                                                                         
   name = "<redacted>"                                                                                                                                                                                                               
   database = "app"                                                                                                                                                                                                                  
   password = "<redacted>"                                                                                                                                                                                                           
   ```

dipeshbabu:
Fixes #1591.

Upgrade `tikv-jemallocator` and `tikv-jemalloc-ctl` to 0.7. The bundled jemalloc 5.3.1 defaults to 64 KiB pages on ARM64 Linux, so Docker images and release binaries can start on 64 KiB kernels without build-specific overrides.

Validation: [all 29 jobs passed](https://github.com/dipeshbabu/pgdog/actions/runs/35564000336), including the actual Docker image on a native 4 KiB host and a 64 KiB ARM64 kernel in QEMU. The Go download failed once with a connection reset; the existing Rust partial-disconnect test also failed once. Both jobs passed unchanged on retry.

Paired native ARM64 benchmarks used the same code, Rust 1.96.1, PostgreSQL 16, and configuration. Each case ran four alternating pairs, with 3s warmup and 10s measurement per run.

| Workload | Clients | 0.6 median TPS | 0.7 median TPS | Change | 0.6 / 0.7 peak RSS (KiB, median) |
| --- | ---: | ---: | ---: | ---: | ---: |
| SELECT 1 | 1 | 7,254 | 7,477 | +3.1% | 19,220 / 19,554 |
| SELECT 1 | 16 | 28,755 | 28,300 | -1.6% | 20,430 / 20,684 |
| 8 KiB result | 1 | 6,370 | 6,303 | -1.1% | 19,456 / 19,750 |
| 8 KiB result | 16 | 15,725 | 15,792 | +0.4% | 21,372 / 22,344 |

These short CI measurements show small mixed throughput changes and a modest RSS increase; they do not establish production-wide performance. Raw samples and the comparison script are in the validation run's `allocator-benchmark` artifact and CI branch.


.github/workflows/release-binaries.yml
```diff
@@ -45,6 +45,12 @@ jobs:
         with:
           targets: ${{ matrix.target }}
 
+      # ARM64 Linux kernels use 4K, 16K, or 64K pages. jemalloc's compile-time
+      # page size must cover the largest runtime page size.
```
levkk: I don't entirely follow the root cause and resolution. If the compile-time page must cover the largest runtime size, then why 16k page size works for 64k pages?

dipeshbabu: `16` is log2(bytes), so it means `2^16 = 65,536` bytes (64 KiB), not 16 KiB. The default ARM64 build uses 4 KiB, which jemalloc rejects on a 64 KiB kernel. The [CI test](https://github.com/dipeshbabu/pgdog/actions/runs/35529901487/job/106128540872) checks the image on a native 4 KiB host and a 64 KiB ARM64 kernel in QEMU.

jkaczman (changes requested):
Some more information on this:
- This is the original jemalloc issue tracking this: https://github.com/jemalloc/jemalloc/issues/2639
- `JEMALLOC_SYS_WITH_LG_PAGE=16` was made standard in the latest jemalloc (see Facebook's [merge of their fork ](https://github.com/jemalloc/jemalloc/pull/2864) including this [specific PR for it](https://github.com/facebook/jemalloc/pull/34)).
- The reason this error occurs is because we use [jemallocator](https://github.com/tikv/jemallocator/tree/0e9129157d261cf6bda86d5dcd9eba94f3cd0f96). They updated their jemalloc in latest (0.7.0) to 5.3.1, which has this change.

We started using jemalloc in #124 via jemallocator with version 0.6.0. I wonder if we can fix this by just bumping to 0.7; seems like it has significant improvements, and besides just fixing this bug, could help us. 
> This release includes over 390 commits spanning bug fixes, new features,
 performance optimizations, and portability improvements. Multiple percent
of system-level metric improvements were measured in tested production
workloads. The release has gone through large-scale production testing
at Meta.

We'd probably want to benchmark it, if we were to go this way, to be sure. But it looks promising.

dipeshbabu:
Updated both crates to 0.7 and removed the build overrides. The 4 KiB/64 KiB checks pass. Added paired ARM64 benchmarks to the PR body: median throughput changes range from -1.6% to +3.1%, with a small RSS increase.

jkaczman (approved):
Thanks!

meskill (approved):

PR #1598, opened 2026-09-20
fix(resharding): clean replication slots after forced cancellation

### LINKED ISSUE #1592: [Resharding] Cancelling `COPY_DATA` does not remove replication slot
**PgDog version**
`main`

**Description**
Explicitly calling the cancellation API for the `COPY_DATA` task does not remove the replication slot. This causes WAL build-up and eventually the server runs out of available replication slots:

```
backend: ERROR: 53400 all replication slots are in use
```

dipeshbabu:
When a blocked COPY exceeds the cancellation grace period, Tokio drops the migration future before it reaches slot cleanup. The task becomes cancelled while its permanent replication slot continues retaining WAL.

Make the publication guard schedule cleanup on abort and disarm it on successful completion. Keep unfinished slots registered during cleanup so an interrupted cleanup can be retried.

Both normal and forced-abort cancellation regressions passed locally against PostgreSQL. [All 27 pre-PR CI jobs passed](https://github.com/dipeshbabu/pgdog/actions/runs/35530381522) before opening this PR.

The first run hit a dependency download reset and a one-row mismatch in the existing live-write copy-data check. Both jobs passed their isolated reruns without changing the tests.

Fixes #1592.


meskill (approved):

meskill:
Thanks!
Although, during replication we still can leave slots running since we extract them from publisher, but I'm fixing this already in the replication refactoring that I will push soon, so don't worry about this.

PR #1599, opened 2026-09-20
fix(tls): preserve configured certificate chains

### LINKED ISSUE #834: TLS Certificate Chain support

dipeshbabu:
PgDog loads only the first certificate in a configured PEM chain. Clients trusting the root CA cannot verify a server certificate signed by an intermediate. The upstream mTLS client configuration has the same problem.

Load and send the full chain in both roles, reject empty bundles, and keep channel binding tied to the leaf certificate. Add strict root-CA handshake tests for server TLS and mutual TLS.

Reproduced the missing-intermediate error against the unchanged binary. [All 27 pre-PR CI jobs passed](https://github.com/dipeshbabu/pgdog/actions/runs/35532979838) before opening this PR.

Fixes #834.


pgdog/src/net/tls.rs
jkaczman: It looks like the load_certificate_chain method duplicates this code.

pgdog/src/net/tls.rs
```diff
@@ -330,9 +330,23 @@ pub(crate) fn reload() -> Result<(), Error> {
...
+}
+
 fn build_acceptor(cert: &Path, key: &Path, client_ca: Option<&Path>) -> Result<TlsListener, Error> {
-    let pem = CertificateDer::from_pem_file(cert)?;
-    let server_end_point = tls_server_end_point(&pem);
+    let certificates = load_certificate_chain(cert)?;
+    let leaf = certificates
+        .first()
+        .expect("certificate chain is non-empty");
+    let server_end_point = tls_server_end_point(leaf);
```
jkaczman: This could be combined into an and_then() since `tls_server_end_point` returns an Option. There's no need for the expect since we Error on that in `load_certificate_chain`.

pgdog/src/net/tls.rs
```diff
@@ -330,9 +330,23 @@ pub(crate) fn reload() -> Result<(), Error> {
     Ok(())
 }
 
+fn load_certificate_chain(path: &Path) -> Result<Vec<CertificateDer<'static>>, Error> {
+    let certificates = CertificateDer::pem_file_iter(path)?.collect::<Result<Vec<_>, _>>()?;
```
jkaczman: I'm not sure if this Error propagated with the `?` includes the file name. If not, it could be helpful to show to the user if they have multiple configured.

pgdog/src/net/tls.rs
```diff
@@ -706,7 +720,83 @@ impl ServerCertVerifier for NoHostnameVerifier {
...
+    }
+
+    #[test]
+    fn empty_certificate_chains_are_rejected() {
+        crate::logger();
+        let directory = tempfile::tempdir().expect("temporary certificate directory");
+        let empty = directory.path().join("empty.pem");
+        std::fs::write(&empty, "").expect("empty certificate bundle");
+        let key = PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("tests/tls/chain/leaf-key.pem");
+        assert!(build_acceptor(&empty, &key, None).is_err());
```
jkaczman: Would be great if we could specify what error exactly this should throw (for all .is_err() checks). There could be a case where we expect an error from our own check, but rustls (or some other library) gives an unrelated error.

jkaczman (changes requested):
Some initial thoughts. Thanks!

pgdog/src/net/tls.rs
dipeshbabu: Shared the PEM loader between certificate chains and CA bundles.

pgdog/src/net/tls.rs
dipeshbabu: Changed to `first().and_then(tls_server_end_point)`.

pgdog/src/net/tls.rs
dipeshbabu: Read and parse errors now include the certificate path. Added regression checks for both.

pgdog/src/net/tls.rs
dipeshbabu: The tests now assert `InvalidData` and the expected message. All 27 validation jobs pass.

jkaczman (approved):
LGTM!

levkk (approved):

PR #1600, opened 2026-09-20
fix(protocol): handle SQL PREPARE/EXECUTE over extended protocol

### LINKED ISSUE #1403: PARSE and EXECUTE sent via Parse (extended protocol) isn't handled correctly
**PgDog version**
`main`

**Description**
We rewrite all `PARSE` and `EXECUTE` statements to simple queries when executing them on the
servers. This doesn't work if they are sent via extended protocol because client expects an extended protocol reply.

**Configuration**
```toml
[general]
prepared_statements = "full"
```

Pseudocode:

```rust
let query = sqlx::prepare("PREPARE stmt1 AS SELECT $1").await;
sqlx::execute(query).await;
let query = sqlx::prepare("EXECUTE stmt1($1)").await;
sqlx::execute(query).bind(5);
```

Honestly, if you're doing this, you have bigger issues, but this exchange is technically legal, so...

dipeshbabu:
Fixes #1403.

SQL `PREPARE` sent through the extended protocol was replaced with a simple query, leaving clients waiting for ParseComplete. Preserve the original message sequence and track the inner statement when Execute completes.

SQL `EXECUTE` now reparses its rewritten SQL before binding, so cached outer statements use the current LIMIT/OFFSET and generated values. The internal Parse follows any required SQL PREPARE; its ParseComplete is consumed. The original Bind name remains available for decoding results. Client-local SQL preparation plans are not shared through the AST cache.

Keep the client's original unnamed Parse across executions so Postgrex can reuse it after Sync; request rewrites use a separate copy.

Regression coverage includes cross-shard and direct-shard pagination, mixed parameter positions, fresh generated IDs, separate client name registration, unnamed reuse, and error recovery. Added named/unnamed Ruby cases across full/extended/disabled modes and Postgrex protocol coverage. The pagination, generated-ID, and unnamed-statement failures were reproduced before their fixes. The new Ruby and Postgrex cases also passed in isolated local runs.

[All 27 validation jobs passed](https://github.com/dipeshbabu/pgdog/actions/runs/35591387812) before pushing this update.


jkaczman (changes requested):
Going off memory, I think LIMIT / OFFSET won't work with this. If it's a simple-ish fix to cover that, I think it'd be great to support that here as well. If it's difficult, I suppose we could file another issue, and leave a comment in the code explaining that it's broken. I'm also curious if anything else being re-written for Prepare/Execute is affected by this.

Ref to my previous issue implementing that and where I test that general code:
https://github.com/pgdogdev/pgdog/blob/9d83dea3c94deb5435643815a1cb19217edf74bf/integration/rust/tests/integration/simple_prepared.rs#L36-L39

Could you also write a basic test in our Ruby / Elixir suites to ensure there's no other issues? Ruby has suites covering full/extended/disabled prepared_statements, and Postgrex is pretty strict about ensuring protocol conformity through its state machine.

dipeshbabu:
Fixed LIMIT/OFFSET and generated-ID rewrites, including repeated executions. The Postgrex test also caught unnamed-statement reuse; that's fixed too. Added Ruby/Postgrex coverage; all 27 validation jobs pass.

pgdog/src/frontend/router/parser/rewrite/statement/plan.rs
```diff
@@ -249,6 +270,34 @@ impl RewritePlan {
...
+            // dependent LIMIT/OFFSET. Parse it again without adding cache entries.
+            // Keep the original Bind name for the cross-shard result decoder.
+            parse.anonymize();
+            request.anonymous_client_params = self.apply_parse(&mut parse);
+            for message in &mut request.messages {
+                if let ProtocolMessage::Bind(bind) = message {
+                    *message = ProtocolMessage::BindAnonymous(bind.clone());
+                }
+            }
+            request.last_parse = None;
```
jkaczman: Curious if we can move the request messages `Bind` search and the message replacement into the same loop, so that we don't have to iterate over the messages 2x.

pgdog/src/frontend/router/parser/rewrite/statement/offset.rs
```diff
@@ -168,12 +168,15 @@ impl OffsetPlan {
         let new_execute = new_execute.deref();
         let new_execute_sql = deparse(new_execute)?;
 
-        // `ExecuteStmt` will be a Query, because this is simple-protocol.
-        // Replace with our re-written `ExecuteStmt`
-        // (replacing limit/offset with proper multi-shard vlaues)
+        // SQL EXECUTE can also arrive through the extended protocol. Update its
+        // internal Parse so cached outer statements receive the current limits.
         for message in request.messages.iter_mut() {
```
jkaczman: This original comment definitely should be changed to be updated for the new functionality; however, you deleted the general premise of what this loop does. and replaced it with what happens for one of the two cases.

pgdog/src/frontend/client_request.rs
```diff
@@ -82,6 +82,8 @@ impl ClientRequest {
 
     /// Remove any saved state from the request.
     pub(crate) fn clear(&mut self) {
+        // Client keeps the original unnamed Parse for later requests; this
+        // request's copy may have been rewritten for a particular backend.
```
jkaczman: I don't know if this belongs here. Someone reading it without the context of this PR may be confused.

pgdog/src/frontend/router/parser/rewrite/statement/plan.rs
```diff
@@ -235,7 +252,11 @@ impl RewritePlan {
                     anonymous_client_params = self.apply_parse(parse);
                 }
                 ProtocolMessage::Query(query) => self.apply_query(query).await?,
-                ProtocolMessage::Bind(bind) => self.apply_bind(bind, timezone, timestamps).await?,
+                ProtocolMessage::Bind(bind) if self.prepare_rewrites.is_empty() => {
+                    // SQL PREPARE's placeholders belong to the inner statement.
+                    // SQL EXECUTE already has its generated values in the SQL text.
+                    self.apply_bind(bind, timezone, timestamps).await?
```
jkaczman, line 261: I understand why this is the way it is, but the comment is confusing on its own. Someone would have to understand that 
- (1) prepare_rewrites is non-empty when we're dealing with a "SQL" Prepare/Execute
- (2) `apply_bind` injects parameters for Bind for a re-written statement, 
- (3) this is wrong for Prepare/Execute, however, this doesn't apply to other cases, e.g., a regular extended protocol case  (which this match arms represents). 

More specifically, this comment is written for the check for when we shouldn't be entering this match arm, **but it's inside the body;** so, it looks like it's talking about the `apply_bind` specifically, and, it appears to be talking about why we're calling `apply_bind`, when this covers when we shouldn't be.

It's also good to sometimes show examples for clarity 
```
 - Parse, name="", query="PREPARE foo AS INSERT INTO t VALUES ($1, $2)" 
 - Bind, statement="", params=[]  
```

Vs

```
- Parse, query="INSERT INTO t VALUES ($1, $2)"  
     - rewritten to ($1, $2, $3)
- Bind, params=[1, 2] where apply_bind appends the 3rd
```

When we read the comment, we have no concept of what the 'inner' statement is.

I don't think this example should be provided in this place exactly, but it does help paint a picture that helps understand intent better.

jkaczman (commented):
Thanks for fixing that, here's some more thoughts! There's also a couple other places I see the same kind of issues with.

jkaczman (commented):
Oh, and, could you please delete the TODO I mentioned in my earlier comment?

pgdog/src/frontend/router/parser/rewrite/statement/plan.rs
dipeshbabu: Combined them: the replacement loop now records the first Bind index.

pgdog/src/frontend/router/parser/rewrite/statement/offset.rs
dipeshbabu: Updated it to explain the LIMIT/OFFSET rewrite for both protocols. Removed the obsolete TODO from the Rust test too.

pgdog/src/frontend/client_request.rs
dipeshbabu: Removed it here; the explanation is already next to Client::unnamed_parse.

pgdog/src/frontend/router/parser/rewrite/statement/plan.rs
dipeshbabu: Moved the explanation above the condition and added a PREPARE example showing which statement owns $1.

pgdog/src/backend/prepared_statements.rs
```diff
@@ -278,6 +294,22 @@ impl PreparedStatements {
                 self.state.add(ExecutionCode::ExecutionCompleted);
             }
 
+            ProtocolMessage::ExecutePrepare { prepare, .. } => {
+                if self.contains(prepare.name()) {
+                    // SQL statements use global names too, so another client may
+                    // already have prepared this query on the pooled connection.
```
jkaczman:
```suggestion
                    // We re-write `PrepareStmt`s to global names (e.g., __pgdog_1).
                    // No matter what client we receive from, if they use a query we previously received
                    // we resolve it to the same global name (even if a diff local name)
                    // If that already happened on this pooled connection, we must stop from preparing 
                    // again to prevent an error from re-preparing with this global name, 
                    // and return a fake response to the client.
```

jkaczman (commented):
I've read the changes over, I think it generally looks good other than some minor nits.

jkaczman (approved):

sgrif:
We don't need this many separate integration tests. Can you pare this down to a single one in the rust suite?

pgdog/src/net/protocol_message.rs
```diff
@@ -11,10 +11,19 @@ use super::{
...
+    /// Internal anonymous Parse whose ParseComplete is consumed by PgDog.
+    EnsureParsed(Parse),
     Describe(Describe),
     EnsurePrepared(Prepare),
     PrepareFromClient(Prepare),
+    /// Execute a SQL PREPARE sent through the extended protocol.
+    ExecutePrepare {
+        execute: Execute,
+        prepare: Prepare,
+    },
```
sgrif: This enum specifically represents PG protocol messages, we should not be adding to it with variants that do not exist in the protocol

dipeshbabu: Removed the new variants. The SQL rewrite state now lives on ClientRequest.

pgdog/src/frontend/router/parser/rewrite/statement/plan.rs
```diff
@@ -249,6 +266,23 @@ impl RewritePlan {
 
         request.anonymous_client_params = anonymous_client_params;
 
+        if self
+            .prepare_rewrites
+            .iter()
+            .any(|rewrite| matches!(rewrite, PrepareExecute::Execute(_)))
+            && let Some(BufferedQuery::Prepared(mut parse)) = request.query()?
+            && request.parameters()?.is_some()
+        {
```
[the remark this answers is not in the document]
dipeshbabu: PostgreSQL rejects `EXECUTE stmt($1)` too. [Checked both inferred and explicit parameter types with libpq](https://github.com/dipeshbabu/pgdog/actions/runs/36055641808).

PR #1602, opened 2026-09-21
refactor(resharding): migrate replication to api/task

meskill:
Migrate the replication process to task:
- separate the logic into nested tasks: ReplicationTask -> ReplicationClusterTask -> ReplicationShardTask with separate state and tracking
- extend the definitions and statuses used to display progress and send to EE
- rip apart most of the orchestrator and publisher


levkk (approved):
cool cool cool cool

Image

PR #1607, opened 2026-09-21
fix(types): preserve numeric infinities across shards

dipeshbabu:
Cross-shard queries over valid `numeric` infinities fail with `net error: unexpected payload`. Reproduced on current main with `GROUP BY value`.

Represent both infinities explicitly and preserve them through text/binary encoding, comparisons, hashing, arithmetic, and float conversions. Added protocol and arithmetic unit tests, plus PostgreSQL comparisons for sorting, arrays, grouping, and aggregates in both result formats.

The Rust integration regression reads `shard_0` and `shard_1` directly to verify each shard's rows before checking merged sorting, grouping, MIN/MAX/SUM/AVG results. It fails on main and passes with the fix.

Validation: 120 type tests and the live SQL and Rust regressions passed. [All 27 validation jobs passed](https://github.com/dipeshbabu/pgdog/actions/runs/35686076453) for this update.


jkaczman (commented):
The PR description states, "Cross-shard queries over valid numeric infinities..", however, the tests don't assert they're going to different shards. Could you add a rust test too, please?

dipeshbabu:
Added the Rust test. It checks both shards directly before asserting the merged results. All 27 validation checks pass.

dipeshbabu:
Fixed the dependency setup: downloads now retry and finish before archive extraction. All 27 validation jobs passed; upstream CI is running again.

integration/ci/download.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+# Download to a file so curl can discard partial data before retrying.
+# Piping retries directly into tar can concatenate separate download attempts.
+set -euo pipefail
+
+curl --fail --silent --show-error --location \
+    --connect-timeout 15 --max-time 120 \
+    --retry 3 --retry-all-errors --retry-max-time 300 \
+    --output "$2" "$1"
```

dipeshbabu: Removed the CI changes. This PR now contains only the numeric fix and its tests.

pgdog-postgres-types/src/numeric.rs
```diff
@@ -129,12 +159,23 @@ impl FromDataType for Numeric {
                     }),
                     Err(e) => {
                         // Check for special PostgreSQL values
-                        match s.to_uppercase().as_str() {
-                            "NAN" => Ok(Self {
-                                value: NumericValue::NaN,
-                            }),
-                            "INFINITY" | "+INFINITY" | "-INFINITY" => Err(Error::UnexpectedPayload),
-                            _ => Err(Error::NotFloat(e.to_string().parse::<f64>().unwrap_err())),
```
jkaczman, line 137: Looks like we weren't handling this before, I'm guessing because support for numeric Infinity wasn't introduced until Postgres 14.

jkaczman (approved):
Thanks! LGTM.

PR #1608, opened 2026-09-21
fix(pool): track pg_catalog-qualified advisory locks

dipeshbabu:
`pg_catalog`-qualified advisory calls bypass lock tracking. In transaction pooling, a client can lose its backend while holding a lock: another client can then acquire it reentrantly, or an unlock reaches the wrong connection. Reproduced both failures on current main.

Recognize the explicit system-catalog qualifier before classifying the lock call. Name matching borrows the parser's strings. Added parser/engine regressions and a two-client test covering simple and prepared protocols.

Validation: 38 advisory tests passed; both live protocol regressions fail before the fix and pass after it. [All 27 pre-PR CI jobs passed](https://github.com/dipeshbabu/pgdog/actions/runs/35649300727) before opening this PR.


jkaczman:
There's some functionality in the tests that will have to change after https://github.com/pgdogdev/pgdog/pull/1611 is merged.  That should happen today. I'll follow up after that's done suggesting some changes.

dipeshbabu:
Got it. I'll wait for #1611 and your follow-up before changing the tests.

jkaczman (approved):

PR #1609, opened 2026-09-21
fix(aggregation): check integer SUM overflow across shards

dipeshbabu:
Cross-shard integer SUM reduction can overflow even when each shard's result fits. For example, partial `SUM(value)::smallint` results of 32767 and 1 panic the worker in a debug build instead of reporting a range error.

Accumulate integer partials in checked `i128` values and validate the result width after all shards contribute. This also preserves valid cancellation such as MAX + 1 - 1. Preserve the typed error through multi-shard aggregation so out-of-range results return SQLSTATE `22003` through the existing fatal-error path.

Validation: 34 aggregate unit tests, two error-mapping tests, and 12 live PostgreSQL comparisons covering all three integer widths, both overflow directions, boundaries, cancellation, and text/binary results. [All 27 validation jobs passed](https://github.com/dipeshbabu/pgdog/actions/runs/35686086745).


dipeshbabu:
Most failures are dependency-download errors. Resharding also reported `tenants: source=4577 destination=4576`; that suite passed in pre-PR CI. Could you rerun the failed jobs? GitHub requires repository admin rights.

dipeshbabu:
Fixed the dependency setup: downloads now retry and finish before archive extraction. All 27 validation jobs passed; upstream CI is running again. Resharding passed in validation too.

integration/ci/download.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+# Download to a file so curl can discard partial data before retrying.
+# Piping retries directly into tar can concatenate separate download attempts.
+set -euo pipefail
+
+curl --fail --silent --show-error --location \
+    --connect-timeout 15 --max-time 120 \
+    --retry 3 --retry-all-errors --retry-max-time 300 \
```

dipeshbabu: Removed the CI changes. The diff is back to the SUM fix and its tests.

PR #1610, opened 2026-09-21
fix(pool): distinguish unresolved unlocks from unlock-all

dipeshbabu:
`pg_advisory_unlock(NULL)` was treated as `pg_advisory_unlock_all()` by the session lock tracker. With another session lock still held, PgDog cleared the tracker and released the backend pin. An individual unlock whose key cannot be resolved during parsing had the same problem.

Track lock, individual unlock, and unlock-all as separate actions. Only explicit unlock-all clears every tracked lock; an unresolved individual unlock preserves existing locks and the pin, without extra backend queries.

Validation: parser and query-engine regressions plus eight live cases covering NULL literals, NULL parameters, subqueries, and VALUES expressions in text and binary modes. Verified this merges cleanly with #1608. [All 27 validation jobs passed](https://github.com/dipeshbabu/pgdog/actions/runs/35689285753).

The checkpoint regression now rotates its final WAL segment before expecting full recycling. A live phase record can keep older identity segments after a transaction completes (#1535). Rotation hooks are test-only. The focused test passed five consecutive runs, alongside all six checkpoint safety tests.


dipeshbabu:
CI stopped during dependency installation (`apt.postgresql.org` timeout / curl 56). All 27 pre-PR checks passed. Could you rerun the failed jobs? I don't have the required repo permissions.

dipeshbabu:
Fixed the dependency setup: downloads now retry and finish before archive extraction. All 27 validation jobs passed; upstream CI is running again.

dipeshbabu:
Fixed the checkpoint fixture too: it now closes the live WAL segment before checking cleanup. Five repeat runs and all 27 validation jobs passed.

integration/ci/download.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+# Download to a file so curl can discard partial data before retrying.
+# Piping retries directly into tar can concatenate separate download attempts.
+set -euo pipefail
+
+curl --fail --silent --show-error --location \
+    --connect-timeout 15 --max-time 120 \
+    --retry 3 --retry-all-errors --retry-max-time 300 \
+    --output "$2" "$1"
```

jkaczman (changes requested):

integration/ci/download.sh
dipeshbabu: Removed the CI changes and the unrelated checkpoint test change. Only the unlock fix and its tests remain.

jkaczman:
There's some functionality in the tests that will have to change after https://github.com/pgdogdev/pgdog/pull/1611 is merged.  That should happen today. I'll follow up after that's done suggesting some changes.

dipeshbabu:
Got it. I'll wait for #1611 and your follow-up before changing the tests.

jkaczman:
@dipeshbabu alright, ready to start looking at this now! Please resolve the conflicts and address any failing tests.

dipeshbabu:
Resolved the conflicts and updated the tests for #1611. All checks pass.

PR #1611, opened 2026-09-22
fix: route advisory locks through id hash

### LINKED ISSUE #1583: [Advisory locks] Locks use round robin routing in sharded DBs, which breaks them
**PgDog version**
`main`

**Description**
Advisory lock routing uses round robin, which in sharded deployments, will lock multiple shards without accounting for routing.

E.g.,:

```sql
SELECT pg_advisory_lock(1234);
```

executed twice in a row will go to two different shards, defeating the purpose.

We should hash the ID and route it to a deterministic shard. Same for unlock. For unlock all, we should broadcast to all shards.

jkaczman:
When receiving a `pg_advisory_lock` function, or related methods (`pg_advisory_lock_shared`, `pg_try_advisory_lock_shared`, `pg_advisory_xact_lock`, `pg_advisory_xact_lock_shared`, `pg_try_advisory_xact_lock`,
 `pg_try_advisory_xact_lock_shared`, `pg_advisory_unlock`), this now deterministically hashes the ID provided in the function to a specific Shard. Previously, routing was based on round-robin, which defeated the purpose of locks, and locks with the same ID were sent to different shards on subsequent requests.

`pg_advisory_unlock_all` now broadcassts to all Shards, where it previously would only be sent to a singular one.

Going to reply to this PR with a couple of comments for some things I'm not 100% sure on.

Fixes #1583


pgdog/src/frontend/router/parser/error.rs
```diff
@@ -10,6 +10,18 @@ pub(crate) enum Error {
...
+    //
+    // A better message could advise them to use a singular advisory lock
+    // per query. However, that's not necessarily the limitation at hand...
+    //
+    // We could also tell them the algorithm used internally in the docs.
+    // Or, perhaps, make it configurable to an extent, e.g., with what we
+    // do with sharding keys (mapping specific lock IDs to specific shards)
+    #[error("the advisory locks in this query resolve to different shards")]
+    CrossShardAdvisoryLockAttempt,
+
```
jkaczman, line 24: Curious what you all think about this

pgdog/src/frontend/router/parser/query/select.rs
```diff
@@ -60,6 +60,59 @@ impl QueryParser {
...
+                //
+                // Since advisory locks are stored in memory, we don't have to worry
+                // about accounting for resharding / changing shard_count later in time
+                //
+                // Advisory lock can be None when an inner function is used, e.g., hashtext
+                // Since we don't parse that explicitly, they're deterministically always hashed
+                // to the first shard. TODO: if an app exclusively uses this, could cause issues
+                let hashed_shard: usize = match advisory_locks
+                    .iter()
+                    .map(|lock| lock.id.unwrap_or(0).unsigned_abs() as usize % context.shards)
```
jkaczman, line 88: and this

pgdog/src/frontend/router/parser/error.rs
levkk: 👍

pgdog/src/frontend/router/parser/query/select.rs
jkaczman: Going to look into adding support for a couple common functions

